PR #25, opened 2021-05-03
Update: Change version signature

Ibrahimmaga:
version signature 1.0.0


awentzel (changes requested):
I don't think we should bump to 1.0.0 until we have publish working.  Otherwise, we could accidentally publish a major version and then not be able to revert or recover that version in the future when we need it.

awentzel:
We are not ready to bump to 1.0 version.

Ibrahimmaga:
I already switched it back to 0.0.2

Issue #48, opened 2021-05-21
Add fluent-number-field wrapper to fast-blazor


vnbaaij:
# 🙋 Feature Request

In the set of Fluent Web Components there is a fluent implementation available for the Number field component. A wrapper for this fluent-number-field is missing from this fast-blazor library

## 🤔 Expected Behavior

There should be a `<FluentNumberField>` wrapper for the `<fluent-number-field>` component

Issue #39, opened 2021-05-06
Web component check boxes, switches, and list boxes don't bind in sample app


danroth27:
Repro steps:

- Run the FluentUIServerSample
- Browse the the Web components bindings
- Try out the check boxes, switches, and list boxes

Expected results:
- New value is correct bound

Actual result:
- Bound value stays unchanged

EisenbergEffect:
@nicholasrice Is this bug a result of Blazor trying to update through the attribute and not the property (and special casing for built-ins around this scenario)? This has come up in the community again and we need to pick back up that discussion thread and drive for a resolution. We may need @chrisdholt to help with that as it may involve changes in our components to help work around the missing Blazor feature (binding to properties).

vnbaaij:
This is also being discussed in https://github.com/dotnet/aspnetcore/pull/27030. There is a solution in the PR which might be able to be baked into .NET 6

nicholasrice:
Yea that's my understanding of the issue. I think making components work with attributes or properties is the most pragmatic approach at this point, though I do wish Blazor (and other lacking UI systems) provided explicit property and attribute bindings.

If we do end up doing the work to make components work with properties or attributes, we'll need to make sure that form-associated controls still behave correctly in some of the edge-case scenarios like form resets and [form-state restoration ](https://web.dev/more-capable-form-controls/#restoring-form-state.)

danroth27:
>  I do wish Blazor provided explicit property and attribute bindings.

@stevesandersonms @javiercn Is this something we should consider in Blazor?

javiercn:
@danroth27 Today Blazor only works with attributes.

We've discussed this in the past, however its not something to be done lightly since it has an impact on the language and the runtime. If we think this is important we can put it on the list for 7.0, since this needs design and time to bake as the consequences of getting it wrong are bad.

vnbaaij:
> @danroth27 Today Blazor only works with attributes.
> 
> We've discussed this in the past, however its not something to be done lightly since it has an impact on the language and the runtime. If we think this is important we can put it on the list for 7.0, since this needs design and time to bake as the consequences of getting it wrong are bad.

While I understand that you want this to be done right, waiting till 7.0 means there won't be a compelling reason to release this library until then. Ability to bind checkboxes etc is **crucial**. 
Can an interim solution like mentioned in dotnet/aspnetcore#27030 be used to make binding possible sooner (and maybe even backported to 5.x)?

javiercn:
[...]

PR #52, opened 2021-05-25
Add <FluentNumberField>

vnbaaij:
# Pull Request

## 📖 Description
This PR adds a `<FluentNumberField>` component to wrap the <fluent-number-field> component 
### 🎫 Issues
#48,
#39 
Bug in FAST (https://github.com/microsoft/fast/issues/4776) affects this component as well.

## 👩‍💻 Reviewer Notes
Besides the component itself (`FluentNumberField.razor`), pages have been added and modified to test the component. These are `FluentNumberFieldTest`, `ComponentBindings` and `RawBindings`

### General

### Component-specific

## ⏭ Next Steps
Work needs to be done on the binding of values related to the up/down buttons. Seems `onchange` events are not triggered by those.
The precision in the Javascript is limited to 12 digits. When using a .NET `long` the MinValue and MaxValue are longer than 12 digits. Is it possible to change this in the script?


EisenbergEffect:
Thanks @vnbaaij ! This will be a great contribution. I noticed there was an issue with the implicit bind. I think you need to add some configuration here: https://github.com/microsoft/fast-blazor/blob/main/src/Microsoft.Fast.Components.FluentUI/BindAttributes.cs Can you give that a try and see if that fixes the issue?

vnbaaij:
I added the configuration. This did indeed fix the implicit bind. It doesn't fix that up/down do not fire `onchange` though.

EisenbergEffect:
The buttons fire an `oninput` event. `onchange` is fired on blur only I believe while `oninput` is fired on all input (key, button press, etc.). Here's the commit that fixed the underlying issue which did exists: https://github.com/microsoft/fast/commit/66bcd958ae59e56618631af50020d7234bb7a2e2

vnbaaij:
So how would we be able to connect these specific `oninput` events to the value? Do we need to add more configuration to the BindAttributes file?

EisenbergEffect:
Let's ask @SteveSandersonMS for help on this one. There may be a way to configure things better. Perhaps this is just a case where the event override need to be used when binding? Or should input be the default event?

SteveSandersonMS:
You can configure `@bind` to use whatever default event name you want, but it does have to be one specific event. It would be most normal to use `change` so that an update doesn't fire on each keystroke unless the developer specifically opts into that by putting `@bind:event="oninput"` on the element.

For `<input type="number">` we use the `change` event, and the standard DOM element does fire a `change` when you click its up/down buttons. I would have expected a `<fluent-number-field>` to behave like `<input type="number">` in terms of which events it raises.

EisenbergEffect:
Thanks Steve!

Pinging @robarbms who worked on this component, as there may be a bug as well if we're not matching the native component's event patterns.

EisenbergEffect:
Pinging @robarbms to check on the event patterns.

EisenbergEffect:
@vnbaaij Can you update this PR to use the `change` event and then we'll get it merged in. Thanks!

vnbaaij:
@EisenbergEffect done!

EisenbergEffect:
Awesome. Adding to my queue for final review. Thanks!

robarbms:
Weird, I did not get this thread when you added me. I have a fix for this in the FAST library that's waiting on reviews. [PR 4881](https://github.com/microsoft/fast/pull/4881/files)

This fix will fire input and change when changing the value through the text box or using the increment or decrement buttons.

EisenbergEffect (approved):

EisenbergEffect:
Thanks @robarbms ! I reviewed and approved the other PR, so hopefully we can get that in shortly.

@vnbaaij This looks good. Thanks also for updating the examples. @nicholasrice Can you do a second review of this? I'd like to get this in and then do the 2.0 fixup work.

nicholasrice (approved):

EisenbergEffect (approved):

PR #53, opened 2021-05-25
FluentTextField changes and additions

vnbaaij:
# Pull Request

## 📖 Description
Additions and changes to the FluentTextField component. 
- Rename parameter `Resize`  to `Size`. As per https://github.com/microsoft/fast/blob/master/packages/web-components/fast-foundation/src/text-field/text-field.spec.md a `FluentTextField` does not contain a `Resize` parameter (of type Resize?) but does contain a `Size` parameter (of type int?)

Add new parameters:
- MinLength (int?); the minimum number of characters long the input can be and still be considered valid
- MaxLength (int?);  the maximum number of characters the input should accept
- TextFieldType (TextFieldType? enum (text, email, password, tel, url))
- Spellcheck (bool?);controls whether or not to enable spell checking for the input field, or if the default spell checking configuration should be used 

## 👩‍💻 Reviewer Notes
Look at `FluentTextField.razor` for implementation of parameters, `TextFieldType.cs` for enum and `FluentTextFieldTest.razor` for testing the new parameters

### General

### Component-specific

### DIFF
--- a/src/Microsoft.Fast.Components.FluentUI/Components/FluentTextField.razor
+++ b/src/Microsoft.Fast.Components.FluentUI/Components/FluentTextField.razor
@@ -1,5 +1,20 @@
 @inherits FluentInputBase<string>
-<fluent-text-field autofocus="@Autofocus" appearance="@Appearance.ToAttributeValue()" value="@BindConverter.FormatValue(CurrentValue)" @onchange="@(EventCallback.Factory.CreateBinder<string>(this, __value => CurrentValueAsString = __value, CurrentValueAsString))" readonly="@Readonly" disabled="@Disabled" required="@Required" placeholder="@Placeholder" @attributes="AdditionalAttributes">@ChildContent</fluent-text-field>
+<fluent-text-field autofocus="@Autofocus"
+                   appearance="@Appearance.ToAttributeValue()"
+                   value="@BindConverter.FormatValue(CurrentValue)"
+                   @onchange="@(EventCallback.Factory.CreateBinder<string>(this, __value => CurrentValueAsString = __value, CurrentValueAsString))"
+                   type="@TextFieldType.ToAttributeValue()"
+                   size="@Size"
+                   minlength="@MinLength"
+                   maxlength="@MaxLength"
+                   spellcheck="@Spellcheck"
+                   readonly="@Readonly"
+                   disabled="@Disabled"
+                   required="@Required"
+                   placeholder="@Placeholder"
+                   @attributes="AdditionalAttributes">
+    @ChildContent
+</fluent-text-field>
 @code{
     [Parameter] public bool? Disabled { get; set; }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Fast.Components.FluentUI/TextFieldType.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Fast.Components.FluentUI
+{
+    public enum TextFieldType
+    {
+        Text,
+        Email,
+        Password,
+        Tel,
```
awentzel, line 14: Do you think it would be more intuitive and consistent to spell out 'Tel' to 'Telephone'?

Ibrahimmaga: The most used is Phone rather than Tel

src/Microsoft.Fast.Components.FluentUI/Components/FluentTextField.razor
```diff
@@ -1,5 +1,20 @@
 @inherits FluentInputBase<string>
-<fluent-text-field autofocus="@Autofocus" appearance="@Appearance.ToAttributeValue()" value="@BindConverter.FormatValue(CurrentValue)" @onchange="@(EventCallback.Factory.CreateBinder<string>(this, __value => CurrentValueAsString = __value, CurrentValueAsString))" readonly="@Readonly" disabled="@Disabled" required="@Required" placeholder="@Placeholder" @attributes="AdditionalAttributes">@ChildContent</fluent-text-field>
+<fluent-text-field autofocus="@Autofocus"
+                   appearance="@Appearance.ToAttributeValue()"
+                   value="@BindConverter.FormatValue(CurrentValue)"
+                   @onchange="@(EventCallback.Factory.CreateBinder<string>(this, __value => CurrentValueAsString = __value, CurrentValueAsString))"
+                   type="@TextFieldType.ToAttributeValue()"
+                   size="@Size"
+                   minlength="@MinLength"
```
awentzel, line 8: As a matter of consistency, is it strange that `@MinLength` is PascalCased whereas most of the others are just lowercase?  Would it be an improvement to bring consistency to all?

src/Microsoft.Fast.Components.FluentUI/TextFieldType.cs
vnbaaij: The way it works now with all the `enum`s in the library is there is a helper function that  'translates' the value to lowercase and inserts that string in the component variable. For an input to be 'phone enhanced' the value of the type needs to be 'tel'. That is why I used 'Tel' in the `enum`

src/Microsoft.Fast.Components.FluentUI/Components/FluentTextField.razor
vnbaaij: Think you got a point there. Did not touch on the ones that were already there, but would be more consistent to rename `@Readonly` to `@ReadOnly` throughout the library and check potential others

awentzel (approved):

Ibrahimmaga (approved):

src/Microsoft.Fast.Components.FluentUI/Components/FluentTextField.razor
```diff
@@ -9,12 +24,23 @@
 
     [Parameter] public bool? Autofocus { get; set; }
 
-    [Parameter] public Resize? Resize { get; set; }
+    [Parameter] public int? Size { get; set; }
 
     [Parameter] public Appearance? Appearance { get; set; }
 
+    [Parameter] public TextFieldType? TextFieldType { get; set; }
```
EisenbergEffect, line 31: I wonder if we should rename this to "Type" so that it matches the attribute names better. This one seems like it doesn't match the pattern of everything else right now.

EisenbergEffect:
@vnbaaij Could re rename the property `TextFieldType` to `Type` so that it's more consistent with the underlying web component?

vnbaaij:
@EisenbergEffect 'Type' is a reserved word in C# so we cannot use it in tihs context

EisenbergEffect:
Ok, so it's not so much a conflict but would likely create confusion. Understood. We'll leave it as you have it then.

EisenbergEffect (approved):

Issue #62, opened 2021-06-16
feat: expose .show() and .hide() functionality. in FluentDialog


markdav-is:
# 🙋 Feature Request
expose .show() and .hide() functionality. in FluentDialog

## 🤔 Expected Behavior

being able to call  .show() and .hide() methods in FluentDialog blazor control
## 😯 Current Behavior
today you have to play around with the hidden attribute to get the dialog to pop. That's OK, but it would be handy to call the underlying web component.

## 💁 Possible Solution

use JS INterop to call the underlying  .show() and .hide() functionality. in FluentDialog

## 🔦 Context

I have a load dialog I'm trying to pop when a button is pressed.

## 💻 Examples

https://gist.github.com/EddyHaigh/639a2b28d542a7e416bd484e7187867a

EisenbergEffect:
We're definitely open to adding JS Interop where it makes sense. This seems like a good case for that. @markdav-is Are you interested in putting together a pull request for this? We'd be happy to work with you to get it in. My team hasn't done any JS Interop work with Blazor yet, so this could also help us learn a bit from the community.

markdav-is:
I can do that.  I'm on deadline for the next few days, but I'm excited to contribute next week.  Should be a very minor PR.

EisenbergEffect:
No rush! We'd love to have your contribution whenever you can get around to it 😄

PR #63, opened 2021-06-17
feat: add Show/Hide methods to FluentDialog

emimontesdeoca:
# Pull Request

## 📖 Description

This is an implementation of `show` and `hide` from `fluent-dialog` component that fixes the issue #62.

My intention was to use the Javascript methods using `JSInterop`, but since you can't directly work in the DOM with Blazor, we would have to create a new Javascript file to hand the `show`/`hide` methods.

This implementation is pretty basic and it only adds a new property called `Hidden`, which will be used to update the `fluent-design` component.

There are two new methods added to the `FluentDialog` component called `Show` and `Hide`. 

```razor
<fluent-dialog Modal="@Modal" @attributes="AdditionalAttributes" hidden="@(Hidden)">@ChildContent</fluent-dialog>
@code
{
    [Parameter] public bool? Modal { get; set; }

    [Parameter] public RenderFragment ChildContent { get; set; }

    [Parameter(CaptureUnmatchedValues = true)] public IDictionary<string, object> AdditionalAttributes { get; set; }

    public bool Hidden { get; set; } = true;

    public void Show() => Hidden = false;

    public void Hide() => Hidden = true;
}
```

In order to either show or hide the dialog, you would need to reference the component by using `@ref` and then you are able to call the methods.

```razor
<FluentDialog @ref="MyFluentDialog" id="foo" aria-label="Simple dialog" Modal="true" >
    <h2>Dialog with text and button. The button should recieve focus</h2>
    <button>Button A</button>
    <button id="element" autofocus>Should autofocus</button>
    <button @onclick="OnCloseModalButtonClick">Close dialog</button>
</FluentDialog>

<button @onclick="OnOpenModalButtonClick">Open dialog</button>

@code{
    private FluentDialog MyFluentDialog;
    public bool IsModalOpen;

    private void OnOpenModalButtonClick() => MyFluentDialog.Show();
    private void OnCloseModalButtonClick() => MyFluentDialog.Hide();
}
```

### 🎫 Issues

Not really an issue, but maybe an explanation on why I decided to not go the JS Interop route.

If we wanted to use JS Interop, we would need to do something like following code, that would have to be included a new Javascript file linked to the `FluentDialog` component, and that calls the function to either `open` or `close`.

```js
var showModal = function(id)
{
    document.getElementById(id).show();
}

var hideModal = function(id)
{
    document.getElementById(id).hide()
}
```

Then, from the `FluentDialog` component we would have to change more stuff, like having the component `id` as random so it doesnt collide with any other `FluentDialog` that we might have.

```razor
<fluent-dialog id="@Id" Modal="@Modal" @attributes="AdditionalAttributes" hidden="@(Hidden)">@ChildContent</fluent-dialog>
@code
{
    [Parameter] public bool? Modal { get; set; }

    [Parameter] public RenderFragment ChildContent { get; set; }

    [Parameter(CaptureUnmatchedValues = true)] public IDictionary<string, object> AdditionalAttributes { get; set; }

    public bool Hidden { get; set; } = true;

    public Guid Id { get; set; } = Guid.NewGuid();

    [Inject] public IJSRuntime JSRuntime { get; set; }

    public async Task Show()
    {
        await JSRuntime.InvokeVoidAsync("showModal", Id.ToString());
    }

    public async Task Hide()
    {
        await JSRuntime.InvokeVoidAsync("hideModal", Id.ToString());
    }
}
```

Probably the code could be refactored to be smaller, but we still would have to create a new Javascript file, and then we would have to do for each component  in order to keep consistency.

## 👩‍💻 Reviewer Notes

Look at `FluentDialog.razor` for the new changes and `FluentDialogTest.razor` for testing the component.

### General

### Component-specific

## ⏭ Next Steps

Maybe the product manager should decide if the components should have Javascript files that call the front-end methods.

### DIFF
--- a/src/Microsoft.Fast.Components.FluentUI/Components/FluentDialog.razor
+++ b/src/Microsoft.Fast.Components.FluentUI/Components/FluentDialog.razor
@@ -1,9 +1,17 @@
-<fluent-dialog Modal="@Modal" @attributes="AdditionalAttributes">@ChildContent</fluent-dialog>
+<fluent-dialog id="@Id" Modal="@Modal" @attributes="AdditionalAttributes" hidden="@(Hidden)">@ChildContent</fluent-dialog>
 @code
 {
+    [Parameter] public string Id { get; set; }
+
     [Parameter] public bool? Modal { get; set; }
 
     [Parameter] public RenderFragment ChildContent { get; set; }
 
     [Parameter(CaptureUnmatchedValues = true)] public IDictionary<string, object> AdditionalAttributes { get; set; }
+
+    public bool Hidden { get; set; } = true;
+
+    public void Show() => Hidden = false;
+
+    public void Hide() => Hidden = true;
 }
--- a/src/Microsoft.Fast.Components.FluentUI/Components/FluentDialog.razor
+++ b/src/Microsoft.Fast.Components.FluentUI/Components/FluentDialog.razor
@@ -1,8 +1,6 @@
-<fluent-dialog id="@Id" Modal="@Modal" @attributes="AdditionalAttributes" hidden="@(Hidden)">@ChildContent</fluent-dialog>
+<fluent-dialog Modal="@Modal" @attributes="AdditionalAttributes" hidden="@(Hidden)">@ChildContent</fluent-dialog>
 @code
 {
-    [Parameter] public string Id { get; set; }
-
     [Parameter] public bool? Modal { get; set; }
 
     [Parameter] public RenderFragment ChildContent { get; set; }
--- a/src/Microsoft.Fast.Components.FluentUI/Components/FluentDialog.razor
+++ b/src/Microsoft.Fast.Components.FluentUI/Components/FluentDialog.razor
@@ -7,7 +7,7 @@
 
     [Parameter(CaptureUnmatchedValues = true)] public IDictionary<string, object> AdditionalAttributes { get; set; }
 
-    public bool Hidden { get; set; } = true;
+    [Parameter] public bool Hidden { get; set; } = false;
 
     public void Show() => Hidden = false;
 
--- a/examples/FluentUIServerSample/Components/FluentDialogTest.razor
+++ b/examples/FluentUIServerSample/Components/FluentDialogTest.razor
@@ -16,7 +16,6 @@
 
 @code{
     private FluentDialog MyFluentDialog;
-    public bool IsModalOpen { get; set; }
     public bool ModalHidden { get; set; } = true;
 
 
--- a/examples/FluentUIServerSample/Components/FluentDialogTest.razor
+++ b/examples/FluentUIServerSample/Components/FluentDialogTest.razor
@@ -3,8 +3,28 @@
       <h2 id="heading">Dialog with aria-labelledby support</h2>
       <button>Button (should autofocus)</button>
   </fluent-dialog> -->
-<FluentDialog id="foo" aria-label="Simple dialog" Modal="true">
+<FluentDialog @ref="MyFluentDialog" Id="foo" aria-label="Simple dialog" Modal="true" >
     <h2>Dialog with text and button. The button should recieve focus</h2>
     <button>Button A</button>
     <button id="element" autofocus>Should autofocus</button>
+    <button @onclick="OnCloseModalButtonClick">Close dialog</button>
 </FluentDialog>
+
+@if (MyFluentDialog == null)
+{
+<p>MyFluentDialog is null</p>
+}
+else
+{
+<p>Modal hidden value: @MyFluentDialog.Hidden</p>
+}
+
+<button @onclick="OnOpenModalButtonClick">Open dialog</button>
+
+@code{
+    private FluentDialog MyFluentDialog;
+    public bool IsModalOpen;
+
+    private void OnOpenModalButtonClick() => MyFluentDialog.Show();
+    private void OnCloseModalButtonClick() => MyFluentDialog.Hide();
+}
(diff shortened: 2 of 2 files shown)


Ibrahimmaga (approved):

EisenbergEffect (approved):

EisenbergEffect:
@nicholasrice Can you give this a review?

src/Microsoft.Fast.Components.FluentUI/Components/FluentDialog.razor
```diff
@@ -1,9 +1,15 @@
-<fluent-dialog Modal="@Modal" @attributes="AdditionalAttributes">@ChildContent</fluent-dialog>
+<fluent-dialog Modal="@Modal" @attributes="AdditionalAttributes" hidden="@(Hidden)">@ChildContent</fluent-dialog>
```
nicholasrice, line 1: Is it possible here for an author to apply the `hidden` attribute via AdditionalAttributes? If so, I'm assuming that it won't take priority, is that correct?

emimontesdeoca: Probably you could add them to the `AdditionalAttributes`, but then you would have to handle the `show` and `hide` from outside of the component? As updating the value inside the `AdditionalAttributes` parameter for the component to be hidden.

EisenbergEffect: Do we need to move `hide` to before `AdditionalAttributes` or does the order not matter here?

src/Microsoft.Fast.Components.FluentUI/Components/FluentDialog.razor
```diff
@@ -1,9 +1,15 @@
...
 @code
 {
     [Parameter] public bool? Modal { get; set; }
 
     [Parameter] public RenderFragment ChildContent { get; set; }
 
     [Parameter(CaptureUnmatchedValues = true)] public IDictionary<string, object> AdditionalAttributes { get; set; }
+
+    public bool Hidden { get; set; } = true;
```
nicholasrice, line 10: The component's default hidden value is `false`, I think this should probably align unless there is a good reason not to: https://github.com/microsoft/fast/blob/master/packages/web-components/fast-foundation/src/dialog/dialog.ts#L32

src/Microsoft.Fast.Components.FluentUI/Components/FluentDialog.razor
nicholasrice: Hmm actually assuming order *does* matter I think we want them ordered as you have them, otherwise the Blazor control could get out of sync with the DOM element in cases where the attribute is applied via `AdditionalAttributes`.

src/Microsoft.Fast.Components.FluentUI/Components/FluentDialog.razor
emimontesdeoca: Makes sense, but it's a bit weird, do we want the component by default to be shown? 

Having it to `false` would make that the developer have to reference the dialog and on the initial rendering call `Hide()` in order to have it hidden at the start. Making it `true` ensures that the component itself it's always hidden in the start until `Show()` is called.

Is there any reason why the original fast component default value is `false`?

EisenbergEffect: I believe this is because the native `dialog` element follows the same pattern and has a `hidden` attribute that is also `false` by default. The idea being that if you add a dialog to the page, it should display itself, unless otherwise configured. So, you would typically use this in a few ways:

* Bind the hidden property using any framework and control it that way.
* Use your frameworks conditional rendering mechanism to dynamically add/remove the dialog per display needs
* Use DOM APIs to manually create and append/remove a dialog per display needs

One of our goals with the web components is to stay as close to the native platform as possible, so I don't think we can change the default value of `hidden`. I also don't want to create confusion by having the Blazor wrappers have different defaults than the underlying web components. That also creates a maintenance and documentation burden for us that we can't take on.

So, I think we do need to switch the wrapper here to match the standard component behavior. I believe that the way this is setup, Blazor devs wouldn't need to reference the element and call show/hide. They should be able to just set the `Hidden` property to false themselves and even control that through binding.

emimontesdeoca: Okay, so in order to keep consistency, you would delete the existing `Hide` and `Show` methods and set the hidden as a optional parameter of the wrapper? So the parent component handles if it's shown or not.

```csharp
[Parameter] public bool Hidden { get; set; } = false;
```

I like your approach and it makes sense, I still think that eventualy developers will have to duplicate code for each dialog they want to use, like having multiples `boolean` objects to handle what's being shown or not. In any case, I think that keeping consistency with the main framework is the best idea.

How do you want the apporach to be? As I mention before, making the `Hidden` property as an optional parameter would be the best for it.

EisenbergEffect: I think we are agreed. We can make the parameter optional with the default being the same as the underlying web component *and* keep the hide/show methods, since they mirror the hide/show on the base component. It just turns out that it's easier to mirror those in Blazor than to do the JS interop to call the originals (which do the same thing in the end).

emimontesdeoca: Updated the code with the new approach and added both cases in the test page.

nicholasrice (approved):

PR #85, opened 2021-10-06
Selectable row FluentDataGrid demo

limefrogyank:
# Pull Request

## 📖 Description

This may not be needed, but if you're planning on recommending data-grid as an alternative to the react-based DetailsList, you may want to have demos that show how parity may be achieved.  This addition shows how to make a data-grid row selectable and continue to work with keyboard navigation.

image

### 🎫 Issues

https://github.com/microsoft/fast-blazor/issues/47

### General

- [X] I have read the [CONTRIBUTING](https://github.com/Microsoft/fast-blazor/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://www.fast.design/docs/community/code-of-conduct/#our-standards) for this project.

## ⏭ Next Steps

None


nicholasrice:
I'm leery of adding this as a component example because I don't think the implementation achieves row selection in an accessible way. I'd prefer to see this example when selected-row support is implemented into data-grid itself rather then set a precedent for inaccessible experiences.

nicholasrice:
`aria-selected` will need to be leveraged appropriately. It should also implement the keyboard model detailed in https://www.w3.org/TR/wai-aria-practices-1.1/#grid (it looks like shift + space selects rows), and ideally other forms of selection like column and cell would be possible. Lastly, we would need to confirm the mouse interaction; I'm uncertain that a mouse click should select a row rather than focus the row or cell so we need to do some research to validate the interaction model.

scomea:
Note that the current version of data grid does not yet have a built-in notion of selection, although the templates should enable authors to implement their own easily enough.  I'm not sure how much of the "selection" concept we should implement on the base data-grid class as there are a lot of different possible selection models for a grid - ie. select regions, select rows, select multiple cells that aren't otherwise connected, etc...  There's also the question of how we track selection since row/cell data objects are currently whatever the author chooses them to be.

awentzel:
@limefrogyank @nicholasrice @scomea what's the status on this PR? Is there more work to do?  It has proper approvals, I'm just not clear on the status and I'm preparing for a Nuget release for .NET 6.0 asap.  Thanks

scomea (changes requested):
On second thought I agree with Nick that we should be careful about creating samples that aren't as accessible as they should be.

awentzel:
@limefrogyank @scomea how would you two like to proceed with this PR? Thanks

limefrogyank:
I'm OK with not using this PR. It's just one way of doing things that may not be the "right way". It probably belongs in a blog post or a samples archive rather than the main demo page.

scomea:
Thinking about this - elements with a role of "row" support an "aria-selected" state - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/row_role

We have an issue on FAST to support row selection. https://github.com/microsoft/fast/issues/4658 We should consider how to best add that.

In the meantime, authors could set aria-selected themselves and trigger styling off of that, but since rows aren't focusable you'd probably need to do it from a cell to make it accessible.  Think "check the header to select the row"

EisenbergEffect:
Closing this for now in favor of improvements to the core data grid tracked by the above referenced issue.

PR #199, opened 2022-06-24
Rename enums

vnbaaij:
# Pull Request

## 📖 Description

- Adding correctly named enums and obsoleting old ones. Make enum extensions public. Make demo pages use new enums.
- Replace SVGs with `<FluentIcon>` components
- Fix `<FluentNumberField>` nullable types handling

### 🎫 Issues

#196 

## 👩‍💻 Reviewer Notes
This PR builds on the events (#198) PR so this must be merged after #198 has been merged

### General

- [X] I have read the [CONTRIBUTING](https://github.com/Microsoft/fast-blazor/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://www.fast.design/docs/community/code-of-conduct/#our-standards) for this project.

### Component-specific


EisenbergEffect (approved):
What was the reason for not using the minified versions of the script? Is it to make debugging the demo easier? I worry that people might copy that and not use the minified version, which is the one that should be used in production. Should we add a comment to that effect in each place where the script is being used?

vnbaaij:
> What was the reason for not using the minified versions of the script? Is it to make debugging the demo easier? I worry that people might copy that and not use the minified version, which is the one that should be used in production. Should we add a comment to that effect in each place where the script is being used?

Yes, not using minified version was done for debugging purposes. I'll set it back to use the minified version before merging.

PR #253, opened 2022-12-01
Fix ToString for other culture

ehsangfl:
on some cultures, default tostring method has an issue for numeric types. so I changed it to use Invariant culture

### DIFF
--- a/examples/FluentUI.Demo.Shared/Pages/Lab/PresenceBadge/PresenceBadge.razor.cs
+++ b/examples/FluentUI.Demo.Shared/Pages/Lab/PresenceBadge/PresenceBadge.razor.cs
@@ -34,9 +34,9 @@ public partial class PresenceBadge : FluentComponentBase, IDisposable
 
     /// <summary />
     protected string? StyleValue => new StyleBuilder()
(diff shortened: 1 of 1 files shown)


vnbaaij:
Can you elaborate on what kind of issues you are referring to?

ehsangfl:
it is UI issue. the bottom property of  PresenceBadge does not setted correctly on some cultures

vnbaaij:
Can you post a screenshot please?

ehsangfl:
after

vnbaaij:
If I go to this page: https://brave-cliff-0c0c93310-233.centralus.azurestaticapps.net/Lab/PresenceBadge and set the direction to 'rtl', the `PresenceBadge`s are displayed in the correct spot, correct?

From the image it seems like you are using the `PresenceBadge` in a way it is not supposed to be used. It is supposed to show if a user is online, away, etc. It seems you are trying to use it inside a button. Try it the other way around, i.e. use the button inside of the `PresenceBadge`:

```
<PresenceBadge>
    <FluentButton>...</FluentButton>
</PresenceBadge>
```

And see how that renders.

vnbaaij (approved):

Issue #263, opened 2022-12-12
feat: Extend Datagrid


ehsangfl:
I Extended DataGrid component with new features :

1- make it editable with some default columns like `TextboxColumn`, `CheckboxColumn `... with validation functionality
2- new embedded default `ColumnOptions` based on column binding data type, for example, if column data is string, then display a textbox to search, or if datatype is 'Date' then display two datepicker for "`from`" and "`to`" filter
3- extend sort capability, so we can sort by a column, then by another column, then by another ...
4- user can remove sort from a column
5- auto generate column header by `[Display]` attribute on model
6- another way to sort `ColumnOptions` dropdown header similar to `excel` sorting added
7- split sort,filter,display property of column. so a column can display a property, sort by another property, filter by another property, or use same property, if not setted
8- Impoved render performance by setting all parameters of Row and Cells by one setter

if you have a plan or agree to add all these features or some of them, I update my fork and create a PR 
BR

vnbaaij:
Discussion on this has moved to #283

PR #267, opened 2022-12-16
Extend Datagrid  #263

ehsangfl:
Extended the datagrid component as described in #263


vnbaaij:
There are a lot of great things in this! Like I said in #263, I'm not sure if we want to take in all of it. Would it make sense to isolate changes and create some smaller PRs?
Some remarks on the current state::
1) We do not work with `#regions` in source files in this repo. Please remove them
2) We do not use `Nullable<...>`, Just use `...?` please
3) Please pull in the latest changes from main
4) Can you look at the conflicts that are there now. I made some modifications to the grid already.
5) Like the idea of the column types. Think prefixing them with DataGrid is a bit overkill. Also, please use the names of the corresponding Fluent components (Textbox->TextField)
6) Is a Combobox really what you would use in a grid? I would think in most cases you want the user to pick a predefined value from a set list and not let them type one themselves (i.e. use a `FluentSelect`)(or offer both)
7) The commit history looks a bit like a mess. Any way you can clean that up a bit?

ehsangfl:
I will create 4 PR
1- generate headers, by `Display` attr
2- new sorting ability
3- embedded filters
4- Inline datagrid editor
is it ok?

vnbaaij:
Sounds good!

One other thing I thought of (sorry, don't want to sound too critical. Just want to make sure we do the 'right things'). For the enums I am using the `System.ComponentModel.Description` attribute. Can you use that as well for the Display? So we can keep it consistent.

ehsangfl:
np. feel free to give tips

PR #268, opened 2022-12-19
Generate column header from DisplayAttribute

ehsangfl:


vnbaaij:
Can you explain why you are using `DisplayAttribute` (from `System.ComponentModel.DataAnnotations`) and not `DisplayName` (from `System.ComponentModel`). I'm already using the later one and using the first one means an extra DLL needs to be downloaded in case of a WASM project. If there is a good reason to use `DisplayAttribute` we can use it of course but otherwise, please change to `DisplayName`

Also, please take a look at the compilation warnings that were generated while running the workflow. Thanks.

ehsangfl:
there is 2 reason for using display attribute
1- `Display` attribute can load data attribute from resource file and `DisplayName` cannot. this is used for some features, somthing like localization.
2- I want to add a feature to datagrid for Auto generating columns. `Display` attribute have the `Order`  and the 'AutoGenerated' properties  for this purpose

vnbaaij:
Cool. I was thinking that was your reasoning so good to verify that.

If you can just look at the warnings then I'll go over the code again and then we can merge, I think

vnbaaij:
Hi @ehsangfl,

Upon going through the code, I think we should maybe rename the interface to `IBindableColumn`. Think that conveys its purpose better. Thoughts?

src/Microsoft.Fast.Components.FluentUI/Components/DataGrid/Infrastructure/ColumnTypes/IBindColumn.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System.Linq.Expressions;
+using System.Reflection;
+
+namespace Microsoft.Fast.Components.FluentUI.DataGrid.Infrastructure;
+
+/// <summary>
+/// a column that can bind to a property of model
+/// </summary>
+public interface IBindColumn
```
vnbaaij, line 9: Rename to IBindableCoulmn ?

src/Microsoft.Fast.Components.FluentUI/Components/DataGrid/Infrastructure/DisplayAttributeExtesions.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Reflection;
+using System.ComponentModel.DataAnnotations;
+
+namespace Microsoft.Fast.Components.FluentUI.DataGrid.Infrastructure;
+internal static class DisplayAttributeExtesions
```
vnbaaij, line 10: Typo: DisplayAttributeExtensions (you forgot the 'n') *(also in the file name)

src/Microsoft.Fast.Components.FluentUI/Components/DataGrid/Infrastructure/ColumnTypes/IBindColumn.cs
vnbaaij, line 9: Rename to IBindableCoulmn ?

src/Microsoft.Fast.Components.FluentUI/Components/DataGrid/Infrastructure/DisplayAttributeExtesions.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+
+    public static string? GetDisplayAttributeString(this Type itemType, string propertyName)
+    {
+        PropertyInfo? propertyInfo = itemType.GetProperty(propertyName);
```
vnbaaij, line 15: We need to get rid of these warnings below

src/Microsoft.Fast.Components.FluentUI/Components/DataGrid/Infrastructure/DisplayAttributeExtesions.cs
```diff
@@ -0,0 +1,33 @@
...
+        //if (PropertyInfo == null && typeof(ICustomTypeProvider).IsAssignableFrom(itemType))
+        //    PropertyInfo = ((ICustomTypeProvider)Item).GetCustomType().GetProperty(PropertyName);
+        if (propertyInfo == null)
+            return null;
+
+        var DisplayAtt = propertyInfo.GetCustomAttributes(typeof(DisplayAttribute), true).FirstOrDefault();
```
vnbaaij, line 21: We don't use `var` in this repo. Please use the type name

src/Microsoft.Fast.Components.FluentUI/Components/DataGrid/Infrastructure/DisplayAttributeExtesions.cs
```diff
@@ -0,0 +1,33 @@
...
+        if (DisplayAtt != null)
+            return ((DisplayAttribute)DisplayAtt).GetName();
+        else
+        {
+            var metadata = itemType.GetCustomAttribute(typeof(MetadataTypeAttribute)) as MetadataTypeAttribute;
```
vnbaaij, line 26: We don't use `var` in this repo. Please use the type name

vnbaaij (changes requested):

src/Microsoft.Fast.Components.FluentUI/Components/DataGrid/Infrastructure/DisplayAttributeExtesions.cs
ehsangfl: I can only disable it.
this is a warning for using reflection, so code analysis warn me that 'Trimming may remove property or functions that u need those'
if I add `DynamicallyAccessedMembers` attribute to 'type' parameter of `GetDisplayAttributeString`, warning will resolved. but I get the same warning on `OnParameterSet` function (the caller function). and we cannot add the same attribute to overrides function and the warning will be remain.

on the other hand, model is not in our assembly and we are not worry about triming models in our project
if you agree, I disable it

vnbaaij: Let's fix the warning in the `GetDisplayAttributeString` method (by adding the `DynamicallyAccessedMember` attribute) and add the following attribute to the `OnParameterSet` method:
```
[System.Diagnostics.CodeAnalysis.SuppressMessage("Trimming", "IL2072:Target parameter argument does not satisfy 'DynamicallyAccessedMembersAttribute' in call to target method. The return value of the source method does not have matching annotations.", Justification = "'DynamicallyAccessedMembersAttribute' cannot be added to overriden method")]
```

PR #275, opened 2022-12-22
Multiple Columns Sorting

ehsangfl:
user can apply a sort to a column, then by another, then by another and so on


vnbaaij:
Hi @ehsangfl,
I have checked out the PR locally and can get it to run...but there is a lot happening. Can you supply as description of what is changed and how all the changes are supposed to work?

Judging on what I've seen so far, this will be something for a 2.1 release, I think. I don't feel it is quite ready for prime time yet.

ehsangfl:
Hi @vnbaaij 

the idea for sortable column is the same as bindable column. so we have another interface that is ISortableColumn.
we dont have to use this interface and we can use existing IBindableCoolumn, but I use another interface called ISortableColumn , so user can bind column to a property and sort by another property. this feature is used for nested objects that we show to user a computed/complex column and sort/filter by another.

if we are going to allow users to sort multi column, then we must allow them to remove sort. so this is the reason of add `ColumnHeaderOptions` control.  so I created a `ColumnHeaderOptions` component to add remove sort. I tried to design it looks like excel.
on the other hand, if user add 'ColumnOtions' then there is tiwce dropdown on header. so all `ColumnOptions` fragment parts moved to `ColumnHeaderOptions` component without any change.

### ColumnBase :
1- All properties and functions from the razor page have been moved to razor.cs. Most of the properties have not changed and this is just to be similar to `Datagrid`, `DataRow`, etc
2- if we are going to allow users to sort multi column, then we must allow them to remove sort. so this is the reason of add `ColumnHeaderOptions` control.  so I created a `ColumnHeaderOptions` component to add remove sort. I tried to design it looks like excel.
on the other hand, if user add 'ColumnOtions', then there is tiwce dropdown on header. so all `ColumnOptions` fragment parts moved from `ColumnBase` to `ColumnHeaderOptions` component without any change.

3- some changes on sort funtionality.

### sort functionality:
1- I added `SortProvider` to add sort to data source of grid 
2- some changes to `GridItemsProviderRequest` to can get multi column insead of on column.
3- `SortDirection` enum removed and existing `System.ComponentModel.ListSortDirection` used.

### DataGrid
1- in collecting columns, we collect sortable columns
2- send all sortable columns to `GridItemsProviderRequest` instead of one column
3- apply sort to datasource
4- remove 'ColumnOptions' part as described in previouse
5- move 'ColumnOptions' css part to isolate css of `ColumnHeaderOptions` componemt

I completed some documentation

Ultimately, you decide whether or not these changes will be released or when they will be released.

But due to the changes in the sorting method, I suggest publishing in version 2. Because continuing to my development program will not cause the breaking changes. But here we have a breaking changes. In my opinion, if the breaking changes comes with the version 2, then  we don't have twice breaking changes, users will be more comfortable

In the end, I agree with whatever you decide
BR

ehsangfl:
> I like the functionality for the most part. From an UI perspective I think the result is still a bit 'messy'. From the comments in the examples, I understand the state the sorted columns end up in is dependent on the order you select/click the columns. That by itself is ok, but it is not being represented in the UI in any way, I believe. Correct? Any ideas on how to solve that?

I check some third parties like devexpress (pic 1), telerik (pic 2), we can use `badge` or simple text and bind it to `SortOrder` property, nearby of sortIcon. I think badge is beter option





> I'm fine with adding an extra interface but I want it to be an optional addition if possible.

I think you are going to implement previous  sort ability and new sort ability at the same time. is this true?
I'm not prefer to we have two same funtionality at the same time. but if you want to keep it, I will back it.

> I'm not happy yet with the way the merged `ColumnHeaderOptions` / `ColumnOptions` look but I think that is not so hard to change. Why did you call this `ColumnHeaderOptions`? The functionality it presents pertains to the whole column, not just the header. Would it be possible to integrate the functionality into `ColumnOptions`?

thats true and I will rename it to `ColumnOptions`

> By doing this, we lose the 'Auto' option and that requires users to specify `DefaultSort="ListSortDirection.Descending"` (or Ascending) to a lot of columns. I would like to stay as close to the QuickGrid functionality for this as possible and also to make it as much 'opt-in' for the library users

I dont understand mean of "auto". I think auto means 'No Sort' that is the same as null value of 'SortDirection' property. I added "remove sort" ability for this purpose. is there a miss understanding by me?

> As described above, I rather have `ColumnOptions` extended than something else added.

there is 2 functionality (user columnoptions and ability of remove sort ) at this time, and another one on next version (for internal filters ability). I dont have any idea other than this.
another important reason is this component need two generic argumant for sort (and filter on next version) but ColumnBase has one argumant (only TGridItem). if I dont add new component, I have to used more 'Reflection's on code.

> I see you altered/changed a lot of the existing examples. Can you make new samples as much as possible to showcase the new functionality? I'll work on a better way of exposing them (instead of everything on one page).

most changes on examples is only adding TValue parameter to  `TemplateColumn`,
I tried to explain functionality on _Multiple column sorting_ section. but I try to explain more

vnbaaij:
> I check some third parties like devexpress (pic 1), telerik (pic 2), we can use `badge` or simple text and bind it to `SortOrder` property, nearby of sortIcon. I think badge is beter option

Yes, a badge seems like a good solution for this.

> > I'm fine with adding an extra interface but I want it to be an optional addition if possible.
> I think you are going to implement previous sort ability and new sort ability at the same time. is this true? I'm not prefer to we have two same funtionality at the same time. but if you want to keep it, I will back it.

No, what I meant is that I would like to have the sorting behavior as it is now to be the 'default' situation. In the current grid it is very easy to get a one-column sort set up and I want it to stay like that.
 
> > By doing this, we lose the 'Auto' option and that requires users to specify `DefaultSort="ListSortDirection.Descending"` (or Ascending) to a lot of columns. I would like to stay as close to the QuickGrid functionality for this as possible and also to make it as much 'opt-in' for the library users
> I dont understand mean of "auto". I think auto means 'No Sort' that is the same as null value of 'SortDirection' property. I added "remove sort" ability for this purpose. is there a miss understanding by me?

The enum describes 'auto' as folows: Automatic sort order. When used with `FluentDataGrid<TGridItem>.SortByColumnAsync(ColumnBase<TGridItem>, SortDirection)`,
the sort order will automatically toggle between `Ascending` and `Descending` on successive calls, and
resets to `Ascending` whenever the specified column is changed. (So, basically `Ascending` as a default and fallback)

> there is 2 functionality (user columnoptions and ability of remove sort ) at this time, and another one on next version (for internal filters ability). I dont have any idea other than this. another important reason is this component need two generic argumant for sort (and filter on next version) but ColumnBase has one argumant (only TGridItem). if I dont add new component, I have to used more 'Reflection's on code.

I was referring to the renaming of `ColumnOptions` (but you already reverted that)

> most changes on examples is only adding TValue parameter to `TemplateColumn`, I tried to explain functionality on _Multiple column sorting_ section. but I try to explain more

Sounds good.

Two other things I thought of:
1) I see in `ApiDocumentation.razor` you are adding `TGridItem="MemberDescription"` to the columns. Can we try to use `CascadingTypeParameter` for that?

2) There is still quite some work to be done on this (and the next wave with filtering). I do not want to hold off on the 2.0 release for that long. Also, I think we need to (try to) include the rest of the community on defining and specifying how this should work. I'm fine with that eventually (perhaps) leading to some breaking changes/new major release. What I think we should do next is create a new topic in the 'Discussion' section and describe what the plans are, how we want to code it and if people think we need to change/add/remove things. As this is your brainchild, I think you should be the one to create the topic (and take the credit!). One is it created, I'll pin it, spotlight it and generate some attention for it on Discord etc. Agreed?

ehsangfl:
Hi @vnbaaij sorry about delay


> The enum describes 'auto' as folows: Automatic sort order. When used with `FluentDataGrid<TGridItem>.SortByColumnAsync(ColumnBase<TGridItem>, SortDirection)`,
> the sort order will automatically toggle between `Ascending` and `Descending` on successive calls, and
> resets to `Ascending` whenever the specified column is changed. (So, basically `Ascending` as a default and fallback)

I think there is three state for sorting of a column
1- no sort (original order)
2- ascending
3- descending

and we need three functionality to user can toggle between these states.

please consider a datagrid with one column sorting behavior (previouse behavior of datagrid).
at first, datagrid does not applied any sort and display original order of data source

1- user click on header, grid goes to `Ascending` state.
2- user click again on header, grid goes to 'Descending' state.
3- user can remove sort by click 'Remove Sort' btn and grid goes to 'No Sort' State

as you see, user can toggle between all states without need to 'Auto'.
I think 'Auto' is duplicate behavior. in current implementation of sort ability, I support all behaviors without need to auto.
can you tell me a behavior that new sort ability not supported and we need to back 'auto' to support that


> * I see in `ApiDocumentation.razor` you are adding `TGridItem="MemberDescription"` to the columns. Can we try to use `CascadingTypeParameter` for that?
> *

unfortunatly, this is not possible cause of another 'TValue' argumant. I suggest to remove TValue from template column and I make TValue as default 'string' type, and I add another extended column with TValue parameter.




> 2\. There is still quite some work to be done on this (and the next wave with filtering). I do not want to hold off on the 2.0 release for that long. Also, I think we need to (try to) include the rest of the community on defining and specifying how this should work. I'm fine with that eventually (perhaps) leading to some breaking changes/new major release. What I think we should do next is create a new topic in the 'Discussion' section and describe what the plans are, how we want to code it and if people think we need to change/add/remove things. As this is your brainchild, I think you should be the one to create the topic (and take the credit!). One is it created, I'll pin it, spotlight it and generate some attention for it on Discord etc. Agreed?

Excellent. Please pin your discussion to continue discuss
thx

vnbaaij:
> I think there is three state for sorting of a column 1- no sort (original order) 2- ascending 3- descending
> and we need three functionality to user can toggle between these states.
> please consider a datagrid with one column sorting behavior (previouse behavior of datagrid). at first, datagrid does not applied any sort and display original order of data source
> 
> 1- user click on header, grid goes to `Ascending` state.
2- user click again on header, grid goes to 'Descending' state. 
3- user can remove sort by click 'Remove Sort' btn and grid goes to 'No Sort' State
> 
> as you see, user can toggle between all states without need to 'Auto'. I think 'Auto' is duplicate behavior. in current implementation of sort ability, I support all behaviors without need to auto. can you tell me a behavior that new sort ability not supported and we need to back 'auto' to support that

I think you are correct, and we do not need `Auto`

> unfortunatly, this is not possible cause of another 'TValue' argument. I suggest to remove TValue from template column and I make TValue as default 'string' type, and I add another extended column with TValue parameter.
> 
No, I think we need to keep that as it is now. If we need to add extra argument, then so be it.

> > 2. There is still quite some work to be done on this (and the next wave with filtering). I do not want to hold off on the 2.0 release for that long. Also, I think we need to (try to) include the rest of the community on defining and specifying how this should work. I'm fine with that eventually (perhaps) leading to some breaking changes/new major release. What I think we should do next is create a new topic in the 'Discussion' section and describe what the plans are, how we want to code it and if people think we need to change/add/remove things. As this is your brainchild, I think you should be the one to create the topic (and take the credit!). One is it created, I'll pin it, spotlight it and generate some attention for it on Discord etc. Agreed?
> 
> Excellent. Please pin your discussion to continue discuss thx
I meant that you should create the post (as it is your work!) in the Discussion section and I'll pin and promote it.

Wanted to come back to the discussion w had on changing the existing samples...I looked at the source in you PR again and I see in `DataGridPage.razor` you changed the `DataGridCustumComparer` demo section to a Multiple column sorting example. You also made the necessarry changes in `DataGridCustumComparer` to support that. I would like to keep the original example (custom comparer) and that you just add a new Multiple column sorting example. That way it stays clearer for people what is happening in an example (I hope)

vnbaaij:
Due to lack of updating and future direction, we are not going to merge in these changes. The `FluentDataGrid` will basically stay a clone of the now productized `QuickGrid`.

PR #380, opened 2023-05-01
Add FluentAssert.Verify

dvoituron:
# Add FluentAssert.Verify method

## 📖 Description

The objective of this method is to generate the HTML code of the component to compare it to a content expected and saved in a sub-file "verified.hml". In case of error, a "received.html" file is generated to simplify the comparison of the two HTML contents.

Using a **Diff tool** such as VSCode or Visual Studio [File Differ extension](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.FileDiffer), will simplify problem comparisons.

This method will solve the current **Id** problem: where IDs are auto-generated and different for each test runs.

image

It's inspired from https://github.com/VerifyTests/Verify.Blazor

## ⏭ Next Steps

To update all existing unit tests to create these ".verified.html" files.


vnbaaij (approved):
Looks good to me!

vnbaaij:
Can you take a look at the conflict

dvoituron:
Done. And I added a method `FluentAssertOptions.ScrubLinesWithReplace` to remove or replace some HTML Attributes by static values. Example : replace `id="..."` by `id="xxx"`.

SimonCropp:
why not just use verify?

PR #381, opened 2023-05-01
Add IdentifyContext

dvoituron:
# IdentityContext

## 📖 Description

Each time a component is generated, a new **identifier** is automatically created and **is different each time**. This can cause problems in Unit Tests or in HTML code processing tools.

Using an **IdentifyContext**, inspired by this [blog post](https://dvoituron.com/2020/01/22/UnitTest-DateTime) it is possible to specify a method for generating identifiers, for example sequentially.

These contexts are Thread Safe. 

To use it, it's very simple:

```csharp
using var id = new IdentifierContext((n) => $"f{n:0000}");
```

image


dvoituron:
Abandoned to https://github.com/microsoft/fluentui-blazor/pull/382

PR #490, opened 2023-07-04
Users/vnbaaij/generic slider

vnbaaij:
# Pull Request

### General

- [ ] I have read the [CONTRIBUTING](https://github.com/Microsoft/fast-blazor/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://www.fast.design/docs/community/code-of-conduct/#our-standards) for this project.

### Component-specific

### DIFF
--- a/src/Microsoft.Fast.Components.FluentUI/Components/Base/InputHelpers.cs
+++ b/src/Microsoft.Fast.Components.FluentUI/Components/Base/InputHelpers.cs
@@ -1,4 +1,5 @@
 ﻿using System.Globalization;
+using Microsoft.AspNetCore.Components;
 
 namespace Microsoft.Fast.Components.FluentUI;
 
@@ -150,5 +151,26 @@ internal static class InputHelpers<TValue>
         {
             ValidateDecimalInput(max, min);
         }
+    }
+
+    /// <summary>
+    /// Formats the value as a string. 
+    /// </summary>
+    /// <param name = "value">The value to format.</param>
+    /// <returns>A string representation of the value.</returns>
+    internal static string? FormatValueAsString(TValue? value)
+    {
+        // Avoiding a cast to IFormattable to avoid boxing.
+        return value switch
+        {
+            null => null,
+            int @int => BindConverter.FormatValue(@int, CultureInfo.InvariantCulture),
+            long @long => BindConverter.FormatValue(@long, CultureInfo.InvariantCulture),
+            short @short => BindConverter.FormatValue(@short, CultureInfo.InvariantCulture),
+            float @float => BindConverter.FormatValue(@float, CultureInfo.InvariantCulture),
+            double @double => BindConverter.FormatValue(@double, CultureInfo.InvariantCulture),
+            decimal @decimal => BindConverter.FormatValue(@decimal, CultureInfo.InvariantCulture),
+            _ => throw new InvalidOperationException($"Unsupported type {value.GetType()}"),
+        };
     }
 }
--- a/src/Microsoft.Fast.Components.FluentUI/Components/Base/InputHelpers.cs
+++ b/src/Microsoft.Fast.Components.FluentUI/Components/Base/InputHelpers.cs
@@ -1,0 +1,154 @@
+﻿using System.Globalization;
+
+namespace Microsoft.Fast.Components.FluentUI;
+
+internal static class InputHelpers<TValue>
+{
+
+    public static string GetMaxValue()
+    {
+        Type? targetType = Nullable.GetUnderlyingType(typeof(TValue)) ?? typeof(TValue);
+        TypeCode typeCode = Type.GetTypeCode(targetType);
+        string value = typeCode switch
+        {
+            TypeCode.Decimal => decimal.MaxValue.ToString(CultureInfo.InvariantCulture),
+            TypeCode.Double => double.MaxValue.ToString(CultureInfo.InvariantCulture),
+            TypeCode.Int16 => short.MaxValue.ToString(),
+            TypeCode.Int32 => int.MaxValue.ToString(),
+            TypeCode.Int64 => "999999999999",
+            TypeCode.SByte => sbyte.MaxValue.ToString(),
+            TypeCode.Single => float.MaxValue.ToString(CultureInfo.InvariantCulture),
+            TypeCode.UInt16 => ushort.MaxValue.ToString(CultureInfo.InvariantCulture),
+            TypeCode.UInt32 => uint.MaxValue.ToString(),
+            TypeCode.UInt64 => "999999999999",
+            _ => ""
+        };
+
+        return value;
+    }
+
+    public static string GetMinValue()
+    {
+        Type? targetType = Nullable.GetUnderlyingType(typeof(TValue)) ?? typeof(TValue);
+        TypeCode typeCode = Type.GetTypeCode(targetType);
+
+        string value = typeCode switch
+        {
+
+            TypeCode.Decimal => decimal.MinValue.ToString(CultureInfo.InvariantCulture),
+            TypeCode.Double => double.MinValue.ToString(CultureInfo.InvariantCulture),
+            TypeCode.Int16 => short.MinValue.ToString(),
+            TypeCode.Int32 => int.MinValue.ToString(),
+            TypeCode.Int64 => "-999999999999",
+            TypeCode.SByte => sbyte.MinValue.ToString(),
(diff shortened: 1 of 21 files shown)


src/Microsoft.Fast.Components.FluentUI/Components/Base/InputHelpers.cs
```diff
@@ -0,0 +1,154 @@
+﻿using System.Globalization;
+
+namespace Microsoft.Fast.Components.FluentUI;
+
+internal static class InputHelpers<TValue>
```
dvoituron, line 5: Maybe a name like `InputNumberHelpers` (or `NumberFieldHelpers`) would be more precise and adapted.

src/Microsoft.Fast.Components.FluentUI/Components/Base/InputHelpers.cs
```diff
@@ -0,0 +1,154 @@
...
+{
+
+    public static string GetMaxValue()
```
dvoituron, line 8: Even if the class is `internal`, it's always better to put the methods in `internal` too (it's clearer).

src/Microsoft.Fast.Components.FluentUI/Components/NumberField/FluentNumberField.razor.cs
```diff
@@ -59,13 +59,13 @@ public partial class FluentNumberField<TValue> : FluentInputBase<TValue>
     /// Gets or sets the maximum value
     /// </summary>
     [Parameter]
-    public string? Max { get; set; } = GetMaxValue();
+    public string? Max { get; set; } = InputHelpers<TValue>.GetMaxValue();
```
dvoituron, line 62: Why to force to have a `Min` and `Max` value? Why not to keep the ` null` default value?

src/Microsoft.Fast.Components.FluentUI/Components/NumberField/FluentNumberField.razor.cs
```diff
@@ -59,13 +59,13 @@ public partial class FluentNumberField<TValue> : FluentInputBase<TValue>
...
 
     /// <summary>
     /// Gets or sets the minimum value
     /// </summary>
     [Parameter]
-    public string? Min { get; set; } = GetMinValue();
+    public string? Min { get; set; } = InputHelpers<TValue>.GetMinValue();
```
dvoituron, line 68: Since you are setting a generic component, it should be good to have typed properties: `TValue Step`, `TValue Max` and `TValue Min`. It's not logical to have numeric values and these properties as `string`.

src/Microsoft.Fast.Components.FluentUI/Components/Slider/FluentSlider.razor.cs
```diff
@@ -5,25 +5,25 @@
 
 namespace Microsoft.Fast.Components.FluentUI;
 
-public partial class FluentSlider : FluentInputBase<int>
+public partial class FluentSlider<TValue> : FluentInputBase<TValue>
```
dvoituron, line 8: If this `TValue` accepts only number, you should add a `where` clause:
- `where T : struct, IComparable, IComparable<T>, IConvertible, IEquatable<T>, IFormattable`
- Better since .NET7: ` where T : System.Numerics.INumber<T>`

src/Microsoft.Fast.Components.FluentUI/Components/Slider/FluentSlider.razor.cs
```diff
@@ -57,8 +65,48 @@ protected override bool TryParseValueFromString(string? value, out int result, [
...
+            targetType == typeof(decimal))
+        {
+            return "1";
+        }
+        else
+        {
+            throw new InvalidOperationException($"The type '{targetType}' is not a supported numeric type.");
+        }
+    }
+
```
dvoituron, line 110: Could you remove extra spaces?

src/Microsoft.Fast.Components.FluentUI/Components/Slider/FluentSliderLabel.razor.cs
```diff
@@ -26,4 +27,20 @@ public partial class FluentSliderLabel : FluentComponentBase
     /// </summary>
     [Parameter]
     public RenderFragment? ChildContent { get; set; }
+
+    protected string? FormatValueAsString(TValue? value)
+    {
+        // Avoiding a cast to IFormattable to avoid boxing.
+        return value switch
```
dvoituron, line 34: This code is also used in FluentSlider. Could you add a common method in the `InputHelpers` class to centralize these conversions?

dvoituron (commented):
Great job!

src/Microsoft.Fast.Components.FluentUI/Components/Slider/FluentSlider.razor.cs
vnbaaij: Solved by doing:
```
#if NET7_0_OR_GREATER
    where TValue : System.Numerics.INumber<TValue>
#else
    where TValue : struct, IComparable, IComparable<TValue>, IConvertible, IEquatable<TValue>, IFormattable
#endif
```

src/Microsoft.Fast.Components.FluentUI/Components/Slider/FluentSliderLabel.razor.cs
vnbaaij: Done. Using it in `FluentNumberField` now as well

src/Microsoft.Fast.Components.FluentUI/Components/NumberField/FluentNumberField.razor.cs
vnbaaij: Agreed, but I ran into issues with the underlying web component when not using strings. Will revise in v3 branch

src/Microsoft.Fast.Components.FluentUI/Components/NumberField/FluentNumberField.razor.cs
vnbaaij: Good question? Just can't think of the answer anymore 😁
Will leave it like this for v2 (to not break anything) and revise for v3

src/Microsoft.Fast.Components.FluentUI/Components/Base/InputHelpers.cs
vnbaaij: It is used from `FluentSlider` as well (also a `FluentInputBase` decendant)

src/Microsoft.Fast.Components.FluentUI/Components/Base/InputHelpers.cs
vnbaaij: Done!

src/Microsoft.Fast.Components.FluentUI/Components/NumberField/FluentNumberField.razor.cs
dvoituron: OK, but why to force to have a default value?
Without set a specific `Min=null` the HTML generated code will contain `min="..."`.
But you can use the `fluent-number-field` component without these min/max attributes.

Maybe check the generated "string" only if the dev set a value.
In addition, you can set `null` :-)

src/Microsoft.Fast.Components.FluentUI/Components/NumberField/FluentNumberField.razor.cs
dvoituron: Yes, but it's the Web Component that needs a "string" value. On our Blazor side, we need to type the property.

src/Microsoft.Fast.Components.FluentUI/Components/NumberField/FluentNumberField.razor.cs
vnbaaij: Will change for v3!

src/Microsoft.Fast.Components.FluentUI/Components/Slider/FluentSlider.razor.cs
dvoituron: Great :-)

src/Microsoft.Fast.Components.FluentUI/Components/NumberField/FluentNumberField.razor.cs
dvoituron: Perfect :-)

PR #494, opened 2023-07-05
Prevent NavMenuGroup text from overlapping expand/collapse button - #493

### LINKED ISSUE #493: fix: text overlaps expand/collapse button in navmenu
When a `FluentNavMenuGroup` has content that reaches the right-edge of the menu, it appears beneath the expand/collapse button.

mrpmorris:
Fixes #493 

Two problems.

1: The menu group seems to be about 2px slightly more to the start (left) than menu link, but I don't know what margin-inline-start unit to use.
2: I cannot enable RTL mode for some reason, so I cannot check if the expand/collapse appears on the left in that mode.


vnbaaij:
Closing this as it targets the wrong branch

PR #502, opened 2023-07-06
Nav menu group alignment

### LINKED ISSUE #501: fix: NavGroup alignment to match NavLink in NavMenu

mrpmorris:
Fix #501


src/Microsoft.Fast.Components.FluentUI/Components/NavMenu/INavMenuItemsHolder.cs
vnbaaij, line 1: Can you use naming convention like we use with [InternalDialogContex](https://github.com/microsoft/fluentui-blazor/blob/new-components/src/Microsoft.Fast.Components.FluentUI/Components/Dialog/Services/InternalDialogContext.cs)
Use `Register`/`Unregister`
And stay in line with components, so `INameMenuLink` + rename interface to `INavMenuLinkContainer` or just `INavMenuGroup`

vnbaaij (changes requested):

vnbaaij (approved):

PR #507, opened 2023-07-07
NavMenu Expanded binding support

### LINKED ISSUE #503: fix: flicker of expanding groups in navmenu when navigating to another page

### LINKED ISSUE #504: fix: Support two-way binding on Expand in NavMenu and NavTreeView
`<FluentNavMenuGroup Expanded>` will show a menu group expanded by default, and still allow the user to expand/collapse it.

If the FluentNavMenu has to re-render for any reason then this rebinds and causes the menu group to expand again.  The way to solve this is to implement 2-way binding (which would be good practice anyway) on `Expanded`.

However, having to create a state member for each group would be a lot of effort for the consumer if they only wish to default the groups to be expanded.

Suggested solution
1: Keep `Expanded`
2: Rename `OnExpandedChanged` to `ExpandedChanged` in order to support 2 way binding
3: Add a new `InitiallyExpanded` bool parameter that is used in `OnInitialized` - so it defaults the state on first render, but then doesn't force a new value when the group re-renders.

mrpmorris:
fix #504
fix #503


src/Microsoft.Fast.Components.FluentUI/Components/TreeView/FluentTreeView.razor.cs.orig
vnbaaij, line 1: Remove?

PR #528, opened 2023-07-13
Fix: Lots of properties are not working in fluent-design-system-provider

LuohuaRain:
# Pull Request

## 📖 Description
The units for these properties were incorrect and could not be applied even when values were set, due to the absence of measurement units: `px`. Therefore, it was decided to change their data type to string, allowing users to determine both the value and unit type as needed, while also maintaining consistency with the official FluentUI property types (ref: https://github.com/microsoft/fluentui/blob/57de08c18915440cdc54a1242bbdf948e2b0f562/packages/web-components/src/design-system-provider/index.ts#L426). 

For example, if the TypeRampMinus1FontSize property was assigned a value of float `12`, but did not include a unit such as "px" or "em", the value would not be applied correctly in the user interface. By changing the data type to string, users can now include the appropriate unit of measurement along with the value, providing greater flexibility in how the properties can be used.

### General

### Component-specific


LuohuaRain:

dvoituron (approved):

vnbaaij (approved):

PR #542, opened 2023-07-14
Fix menu items staying selected

vnbaaij:
Fix items staying selected

### DIFF
--- a/src/Core/Components/NavMenu/FluentNavMenu.razor.cs
+++ b/src/Core/Components/NavMenu/FluentNavMenu.razor.cs
@@ -222,6 +222,8 @@ public partial class FluentNavMenu : FluentComponentBase, INavMenuItemsOwner, ID
         if (menuItem is not null)
         {
             _selectedTreeItem = menuItem.TreeItem;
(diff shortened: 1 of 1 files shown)


src/Core/Components/NavMenu/FluentNavMenu.razor.cs
```diff
@@ -222,6 +222,8 @@ private void HandleNavigationManagerLocationChanged(object? sender, LocationChan
         if (menuItem is not null)
         {
             _selectedTreeItem = menuItem.TreeItem;
+            // menuItem.SelectedChanged.InvokeAsync(true);
```
dvoituron, line 225: Why to add these comments?

dvoituron (changes requested):

src/Core/Components/NavMenu/FluentNavMenu.razor.cs
vnbaaij: This was a test to unblock Peter and his work. As he created a new PR, this one will be not merged and closed

PR #544, opened 2023-07-14
Wasm not deselecting nav menu

### LINKED ISSUE #523: feat: Allow OnCustomAction on NavMenuGroup
Scenario: NavMenu is collapsed, user selects a group

The correct default behaviour is for the nav menu to expand and the group to appear expanded. My requirement is that a popup menu should appear instead.

Proposal:
Add a callback to FluentNavMenuGroup `EventCallback<FluentNavMenuGroupSelectingArgs> OnSelecting`

Ordinarily, the group is not selectable if it doesn't have a value set in its `Href` parameter.  This would be changed so that it is selectable if `Href` is set, or `OnSelecting` is set.

The args classes would be

```
public class ....EventArgs : EventArgs
{
  public FluentNavMenuGroup Group { get; }
  public bool Cancelled { get; }
  public void Cancel()
  {
    Cancelled = true;
  }
}
```

where Cancelled would default to `true` if the group had no `Href`.

This would allow the user to add an OnSelecting action to the group, handle it with a custom action, and then cancel the original navigation if appropriate.

If executing `OnCustomAction` calls `Cancel` then the item will not be selected

mrpmorris:
Fix #523


src/Core/Components/TreeView/FluentTreeItem.razor.cs
```diff
@@ -73,11 +73,36 @@ public partial class FluentTreeItem : FluentComponentBase
     /// </summary>
     public bool Collapsed => !Expanded;
 
+    private bool _disposed;
```
dvoituron, line 76: That's a detail, but can you put private declarations at the top of the file?

src/Core/Components/TreeView/FluentTreeView.razor.cs
```diff
@@ -44,11 +45,22 @@ public partial class FluentTreeView : FluentComponentBase
     [Parameter]
     public EventCallback<FluentTreeItem> OnExpandedChange { get; set; }
 
+    private readonly Dictionary<string, FluentTreeItem> _allItems = new();
```
dvoituron, line 48: Same here: can you put private declarations at the top of file?

src/Core/Components/TreeView/FluentTreeView.razor.cs
```diff
@@ -44,11 +45,22 @@ public partial class FluentTreeView : FluentComponentBase
...
+    private readonly Debouncer _currentSelectedChangedDebouncer = new();
+    private bool disposedValue;
```
dvoituron, line 50: By convention we prefer to prefix private variable with an underscore.

vnbaaij (approved):

PR #572, opened 2023-07-19
Add validation css and FluentValidationSummary component

vnbaaij:


dvoituron (changes requested):
Could you add a Unit Test to validate validations like this?
```csharp
    [MaxLength(10)]
    [Parameter]
    public string? Street { get; set; }

    [Required]
    [Range(1, 9999)]
    [Parameter]
    public string? Zip { get; set; }
```

src/Core/Components/Forms/FluentValidationSummary.razor
```diff
@@ -0,0 +1,21 @@
+﻿@namespace Microsoft.Fast.Components.FluentUI
+@using Microsoft.AspNetCore.Components.Forms
+@inherits ValidationSummary
+
+@{ var validationMessages = Model is null ? FluentEditContext.GetValidationMessages() : FluentEditContext.GetValidationMessages(new FieldIdentifier(Model, string.Empty));
```
dvoituron, line 5: This line of code should be in razor.cs file

src/Core/Components/Forms/FluentValidationSummary.razor
```diff
@@ -0,0 +1,21 @@
...
+
+    @if (validationMessages.Count() > 0)
+    {
+        <div @attributes="AdditionalAttributes">
```
dvoituron, line 9: Where this property is defined?

src/Core/Components/Forms/FluentValidationSummary.razor
```diff
@@ -0,0 +1,21 @@
...
+             <ul class="validation-errors">
+
+            @foreach (var error in validationMessages)
```
dvoituron, line 12: It would be nice to use a `RenderFragments<T>` for all `validationMessages` elements.

dvoituron (changes requested):

src/Core/Components/Forms/FluentValidationSummary.razor
vnbaaij: From ValidationSummary

src/Core/Components/Forms/FluentValidationSummary.razor
vnbaaij: Agree. But we will work on that for a later version. As of now, each `validationMessage` is just a `string`

src/Core/Components/Forms/FluentValidationSummary.razor
vnbaaij: Done

PR #579, opened 2023-07-20
Update to web-components.min.js v2.5.16 and update/extend relevant examples. Update version (2.4.2)

vnbaaij:
Update version (2.4.2)


examples/FluentUI.Demo.Shared/Pages/TextArea/Examples/TextAreaRowsAndCols.razor
```diff
@@ -0,0 +1,11 @@
+﻿﻿<p>
+    12 rows: <br />
+    <FluentTextArea @bind-Value=value1 Rows="12"></FluentTextArea>
```
dvoituron, line 3: I think we need to use a consistent convention for using double quotes for attributes. Example: https://mathiasbynens.be/notes/unquoted-attribute-values

In future development, it might be better to use `attribute="value"`.

dvoituron (approved):

PR #613, opened 2023-08-13
Add options to enable/disable primary/secondary action button(s) in FluentDialog

ProH4Ck:
# Pull Request

## 📖 Description

Added two new properties to `DialogParameters` to enable or disable primary and secondary action buttons.
Simple Dialog sample has been updated with an example on how to use these new api.

### General

### Component-specific

## ⏭ Next Steps

I'm working to implement a FluentDialog to handle forms and this PR is the first step to achieve this result.


dvoituron:
It's a good idea to have more customization features.

Why not go a step further? For example, by adding a Footer property of type RenderFragment. You could then add other buttons (Apply, Delete, etc.) and even disable existing ones.

ProH4Ck:
> It's a good idea to have more customization features.
> 
> Why not go a step further? For example, by adding a Footer property of type RenderFragment. You could then add other buttons (Apply, Delete, etc.) and even disable existing ones.

Yes. I'm already working on a blank FluentDialog that can be used to host a form (or any other component). 
In this case you need total control of `FluentDialog` because `<EditForm>` tag must enclose also submit button and current structure cannot be used.
With this PR I want to cover a more basic scenario when you just need to activate or deactivate standard action buttons based on logic in hosted component.
If you prefer, I can rename this PR to something like: "Customizable FluentDialog" and push all the code here.
Let me know!

Thanks.

Emanuele

vnbaaij (approved):

vnbaaij:
To prevent having another massive `new-components` PR, let's use small PRs for small additions. That way we can also do us to do smaller point releases (3.x) once 3.0 is out the door. Going to squash and merge this one now.

PR #650, opened 2023-08-25
New component: FluentFormDialog

ProH4Ck:
# Pull Request

## 📖 Description

This PR adds a new component: `FluentFormDialog` based on `FluentDialog`.

As most of the web applications use dialogs to show forms to user, the idea is to create a ready-to-use `FluentDialog`  that host an `EditForm`.

This component can also be used as reference on how to create reusable custom dialogs.

Changes in `FluentDialog`:
- `Hidden` property is now bindable to notify parent component when dialog shows/hides
- `HeaderTemplate` & `FooterTemplate`: two `RenderFragment` used to customize dialog's Header and/or Footer
- `FluentDialogHeader` & `FluentDialogFooter`: two sub-components used to share code between `FluentDialog` and `FluentFormDialog`

### General

### Component-specific

## ⏭ Next Steps

Documentation need to be improved. I'll work on this in the next few days.


vnbaaij:
Hi Emanuele,

We like the work you did for the `Hidden` parameter and  the `HeaderTemplate`/`FooterTemplate`. We think we can merge that in. 
What we don't want to merge in is the new `FluentFormDialog` component. We think that using the `DialogService` is always better and preferred. Also, we think this can be done already with just creating a component that implements the `IDialogContentComponent`.  If adding your component (and then perhaps later other specializations) we end up with dialogs in dialogs. Hope this makes sense.

ProH4Ck:
> Hi Emanuele,
> 
> We like the work you did for the `Hidden` parameter and the `HeaderTemplate`/`FooterTemplate`. We think we can merge that in.
> 
> What we don't want to merge in is the new `FluentFormDialog` component. We think that using the `DialogService` is always better and preferred. Also, we think this can be done already with just creating a component that implements the `IDialogContentComponent`.
> 
> If adding your component (and then perhaps later other specializations) we end up with dialogs in dialogs. Hope this makes sense.
> 
> Let's work together to see how we can/if we need to extend the `FluentDialog` component itself to reach the same goal.

Ok I can move all changed to `FluentDialog` component on a new branch forked from the current main (v3) and open a new PR...
Do you have ideas for other implementations on `FluentDialog` ?

Thanks.
Emanuele

vnbaaij:
Target the `dev` branch please. Thanks.

Would be interesting to see if you can indeed reach the same goal by using the building blocks that already there.

PR #655, opened 2023-08-30
Fix ProgressToast rendering error for WASM

cupsos:
# Pull Request

## 📖 Description

image

```
crit: Microsoft.AspNetCore.Components.WebAssembly.Rendering.WebAssemblyRenderer[100]
      Unhandled exception rendering component: Error: Unknown parameter type a
```

⬆️Bug from https://www.fluentui-blazor.net/Toast#progresstoasts
version: 3.0.0+a9833010226148eb96052927c6a42d51338495a0

`FluentProgress.Value`  is not bindable. There is no `[Parameter] public EventCallback<int?> ValueChanged` 

I found some ways to fix this bug.

- Change type of `ProgressToastContent.Progress` . `int?`  to `int` or `string?` (Like before cb5a3a29698a091f0b3c2e4f7d84d2685776cb33. but don't know why it works...)
- Make `FluentProgress.Value`  as bindable. Add `[Parameter] public EventCallback<int?> ValueChanged` 
- Don’t use bind

I think `FluentProgress`  don’t need bind. So I do that.

### General

- [ ] I have added tests for my changes.
- [x] I have tested my changes.
- [ ] I have updated the project documentation to reflect my changes.
- [x] I have read the [CONTRIBUTING](https://github.com/Microsoft/fast-blazor/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://www.fast.design/docs/community/code-of-conduct/#our-standards) for this project.

### Component-specific

- [ ] I have added a new component
- [x] I have modified an existing component

PR #657, opened 2023-08-30
ProgressToast related cahnges

vnbaaij:
FluentProgress: Add ValueChanged 
FluentProgressRing: Add ValueChanged 
Fix #655 and update example

### DIFF
--- a/examples/Demo/Shared/Pages/Toast/Examples/ToastProgressToasts.razor
+++ b/examples/Demo/Shared/Pages/Toast/Examples/ToastProgressToasts.razor
@@ -10,6 +10,7 @@
 @code {
     private FluentButton? _button;
     private static string _id = "progressToast";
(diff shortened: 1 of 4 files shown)


dvoituron:
Why do you want to set this `Value` property, as a **bindable** property?
In my opinion, this value cannot be modified from within the component itself, but only from the outside.
Fix #655 should be enough.

cupsos:
@vnbaaij Did you forget to revert `FluentProgress.razor.cs` and `FluentProgressRing.razor.cs`?
`ValueChanged` is in there. so `Value` is bindable. But `ValueChanged` is never invoked in code. This could be confusing to user.

FYR: https://learn.microsoft.com/en-us/aspnet/core/blazor/components/data-binding?view=aspnetcore-7.0#binding-with-component-parameters

Issue #643, opened 2023-08-24
fix: Visual Studio 2022 freezes and then crashes with v3-RC-1


StevenRasmussen:
# 🐛 Bug Report

After updating the NuGet packages from v2.4.2 to v3.0.0-RC-1 on a .Net 8 Preview 7 Blazor project, Visual Studio 2022 v17.8.0 Preview 1.0 freezes and then crashes within a couple of minutes while trying to edit a razor file.

## 💻 Repro or Code Sample

Note: The following should provide a way for you to reproduce the issue, but the issue manifests itself without doing these things either... it just takes a little longer to crash.  I feel like the issue has to do with the `Icons` package somehow.

1) Clone the repo here: [https://github.com/StevenRasmussen/FluentUIBlazorIssue](https://github.com/StevenRasmussen/FluentUIBlazorIssue)
2) Open the solution
3) Open the `Index.razor` file
4) Try "Going to definition" on the `Delete` icon name or the `Icon` parameter of the `FluentIcon` component on line 9.
5) Then move your cursor back and forth between those two and hover over them to get the tooltip to show.  Usually within performing this 4-5 times, VS will freeze and crash.  Here's a video of it:

![2023-08-24_15-27-20](https://github.com/microsoft/fluentui-blazor/assets/11198735/b549cb76-9935-4016-8c01-990ac3af674c)

## 🤔 Expected Behavior

VS does not freeze and crash

## 🌍 Your Environment

Windows 10 - Latest updates installed
VS 2022 v17.8.0 Preview 1.0
.Net 8.0.100-preview.7
Microsoft.Fast.Components.FluentUI v3.0.0-rc-1
Microsoft.Fast.Components.FluentUI.Icons v3.0.0-rc-1

AathifMahir:
Facing same issue with .Net 7 and Fluent Blazor v3 Stable on Visual Studio 17.7.2 Latest Stable

vnbaaij:
I'm trying to get the Visual Studio team involved on this.

AathifMahir:
> I'm trying to get the Visual Studio team involved on this.

I guess the analyzers are the ones causing this issue. Let's hope this is just a minor bug

dvoituron:
After opening the **Event Viewer**, I saw this error `Failed to read environment variable [DOTNET_STARTUP_HOOKS]` 
Same for you?

I found this article
https://github.com/dotnet/runtime/issues/87687#issuecomment-1635628045

[...]

PR #685, opened 2023-09-04
Rewrite the Icons.AllIcons property using relfection

dvoituron:
To try to solve a VS2022 bug (#643), this PR rewrites the `Icons.AllIcons` property with a `GetAllIcons` method using reflection


vnbaaij (approved):

AathifMahir:
@dvoituron Awesome man, Thanks for shift response on fixing the issue

AathifMahir:
@vnbaaij Do we have an ETA on the release that includes this fix? aka v3.0.1

vnbaaij:
v3.0.1-preview.1 package is building as we speak. Will share it here so you can try it before we release on NuGet

AathifMahir:
> v3.0.1-preview.1 package is building as we speak. Will share it here so you can try it before we release on NuGet. As a heads up that package will not be signed...
> 

Sure, looking forward to it

vnbaaij:
The changes needed are in the **Icons** package only. Please try the attached package (rename and remove `.zip`) and let us know if this helps/fixes the issue.

[Microsoft.Fast.Components.FluentUI.Icons.3.0.1-preview.1.nupkg.zip](https://github.com/microsoft/fluentui-blazor/files/12512261/Microsoft.Fast.Components.FluentUI.Icons.3.0.1-preview.1.nupkg.zip)

AathifMahir:
> The changes needed are in the **Icons** package only. Please try the attached package (rename and remove `.zip`) and let us know if this helps/fixes the issue.
> 
> [Microsoft.Fast.Components.FluentUI.Icons.3.0.1-preview.1.nupkg.zip](https://github.com/microsoft/fluentui-blazor/files/12512261/Microsoft.Fast.Components.FluentUI.Icons.3.0.1-preview.1.nupkg.zip)
> 

Sure, will try the latest icons package and get back to you shortly

vnbaaij:
It might be less performant indeed, although wit our limited internal testing it does not look bad at all. 

If, for now, we need to choose between a bit less performant or having VS crash, I guess you understand which one we will choose. Once we know the bug in VS/ILSpy is solved we can decide to change it back again.

dgrunwald:
Note that the issue here is a huge number of `yield return` in a linear sequence.
This results in a large amount of control flow blocks, which are processed recursively, causing a stack overflow in the decompiler.
Replacing `yield return a; yield return b; yield return c;` with `return new[] { a, b, c };` would avoid all that extra control flow, no need for reflection.

PR #696, opened 2023-09-05
Apply the Immediate property to "text" components

dvoituron:
Use the `oninput` attribute to apply the `Immediate` and `ImmediateDelay` properties.

These properties are only used with these components:
- FluentNumberField
- FluentTextField
- FluentSearch
- FluentTextArea

## Example

```xml
<FluentTextField @bind-Value="value1" Immediate="true" ImmediateDelay="500" />
<p>You entered: @value1</p>
```
text-immediate

### DIFF
--- a/src/Core/Components/NumberField/FluentNumberField.razor
+++ b/src/Core/Components/NumberField/FluentNumberField.razor
@@ -24,7 +24,8 @@
                      name=@Name
                      required="@Required"
                      appearance="@Appearance.ToAttributeValue()"
-                     @onchange="@(e => SetCurrentValueAsString((string?)e.Value))"
+                     @onchange="@ChangeHandlerAsync"
+                     @oninput="@InputHandlerAsync"
                      @attributes="AdditionalAttributes">
     @ChildContent
 </fluent-number-field>
--- a/src/Core/Components/Search/FluentSearch.razor
+++ b/src/Core/Components/Search/FluentSearch.razor
@@ -19,7 +19,8 @@
                name=@Name
                required="@Required"
                appearance="@Appearance.ToAttributeValue()"
-               @onchange=@(e => SetCurrentValue((string?)e.Value))
+               @onchange="@ChangeHandlerAsync"
+               @oninput="@InputHandlerAsync"
                @attributes="AdditionalAttributes">
     @ChildContent
 </fluent-search>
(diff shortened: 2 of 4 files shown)


vnbaaij (approved):

geowong2022:
> Hi. Yes that could be a good idea. I will add that in my backlog 😊

Thank you so much for considering it. 😊

geowong2022:
> Hi. Yes that could be a good idea. I will add that in my backlog 😊

Hi @dvoituron by any chance do you have an expected time for this?

- Add the ImmediateDelay to FluentAutoComplete? This will be helpful before raising the event OnOptionsSearch when user is still typing.

dvoituron:
> > Hi. Yes that could be a good idea. I will add that in my backlog 😊
> 
> Hi @dvoituron by any chance do you have an expected time for this?
> 
> * Add the ImmediateDelay to FluentAutoComplete? This will be helpful before raising the event OnOptionsSearch when user is still typing.

Hi.
Yes, it's always in our task list. But we'd like to include this component with the same management as the other List components and add this `ImmediateDelay` feature to it. It's a bit more complex than I initially imagined, so it'll take a bit longer...
But we won't forget you :-)

geowong2022:
> > > Hi. Yes that could be a good idea. I will add that in my backlog 😊
> > 
> > 
> > Hi @dvoituron by any chance do you have an expected time for this?
> > 
> > * Add the ImmediateDelay to FluentAutoComplete? This will be helpful before raising the event OnOptionsSearch when user is still typing.
> 
> Hi. Yes, it's always in our task list. But we'd like to include this component with the same management as the other List components and add this `ImmediateDelay` feature to it. It's a bit more complex than I initially imagined, so it'll take a bit longer... But we won't forget you :-)

Alright, that's a good one! 😊 May I know is there any way I can keep track of this update? I would like to get inform as we have a production issue due to this. User type something in the AutoComplete and it will immediately search from DB (async task). While users are typing fast, the call to DB (take some time) making the AutoComplete control to behave weird (text removing by itself).

Don't get me wrong, I am not pushing, just wish that I could track this so I could update my project once it is done. 😊

Once again, many thanks for your hard work.

dvoituron:
We will tag you in this discussion when that will be done (in addition of the Release notes)

PR #729, opened 2023-09-13
[FluentMenu/FluentPopover/FluentOverlay] Close menu/popover automatically when user clicks outside of it (by using overlay)

vnbaaij:
So a menu can be closed by clicking outside of it.

### DIFF
--- a/src/Core/Components/Menu/FluentMenu.razor
+++ b/src/Core/Components/Menu/FluentMenu.razor
@@ -3,7 +3,7 @@
 <CascadingValue Value="this" IsFixed="true">
     @if (Open)
     {
-        <FluentOverlay @bind-Visible=Open OnClose="@CloseAsync" Transparent="true" FullScreen="true" />
+        <FluentOverlay @bind-Visible="@Open" OnClose="@CloseAsync" Transparent="true" FullScreen="true" />
         @if (Anchored)
         {
             <FluentAnchoredRegion Anchor="@Anchor"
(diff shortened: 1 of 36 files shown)


src/Core/Components/Menu/FluentMenu.razor
```diff
@@ -3,7 +3,7 @@
 <CascadingValue Value="this" IsFixed="true">
     @if (Open)
     {
-        <FluentOverlay OnClose="@CloseAsync" Transparent="true" />
+        <FluentOverlay @bind-Visible=Open OnClose="@CloseAsync" Transparent="true" FullScreen="true" />
```
dvoituron, line 6: It's more by "naming convention", but could you always use double-quotes with attributes? `@bind-visible="@Open"`

src/Core/Components/Popover/FluentPopover.razor
```diff
@@ -3,7 +3,7 @@
 
 @if (Open)
 {
-    <FluentOverlay OnClose="@CloseAsync" Transparent=true Dismissable=true FullScreen=true/>
+    <FluentOverlay @bind-Visible=Open OnClose="@CloseAsync" Transparent=true FullScreen=true />
```
dvoituron, line 6: Same here

dvoituron:
Small detail...
It would be interesting to have more "explicit" PR titles. By default, this title will be the commit in the `dev` history.
Something like "[FluentMenu & FluentOverlay] Close menu/overlay automatically when user clicks out".

dvoituron (approved):

PR #731, opened 2023-09-14
Add `EmptyContent` parameter to the DataGrid

vnbaaij:
Adding a new `EmptyContent` parameter to the DataGrid:

DataGridEmptyContent

Implemented as follows:
- Add parameter `EmptyContent` (of type `RenderFragment`)  (name chosen to stay in line with new parameter added on standard Blazor `Virtualize` component)
- Add a class to the isolated `FluentDataGrid.razor.css`:
```
::deep .empty-content {
    width: 100%;
    height: 100%;
    display: flex;
    justify-content: center;
    align-items: center;
    font-weight: 600;
}
```
- The `FluentDataGrid` component will render this when Items/ItemsProvider is empty/has a count of 0:
```
 <div class="empty-content">
         @if (EmptyContent is null)
          {
                 @("No data to show!");
          }
          else
          {
                  @EmptyContent
           }
</div>
``` 
This allows you to specify your own text and/or layout for when there is no content or uses a sensible default when none is provided.
- Only caveat is that if you specify your own content for the `EmptyContent` parameter, the `PropertyColumn`/`TemplateColumn` definitions need to be wrapped in a `ChildContent` element, i.e.:
```
<FluentDataGrid Style="height: 400px;" Items=@items Virtualize="true" ItemSize="32" GridTemplateColumns="1fr 1fr 1fr 1fr">
        <ChildContent>
            <PropertyColumn Property="@(c => c.Item1)" Sortable="true" />
            <PropertyColumn Property="@(c => c.Item2)" />
            <PropertyColumn Property="@(c => c.Item3)" Align="Align.Center" />
            <PropertyColumn Property="@(c => c.Item4)" Align="Align.End" />
        </ChildContent>
        <EmptyContent>Nothing to see here!</EmptyContent>
    </FluentDataGrid>
```

### DIFF
--- a/src/Core/Components/DataGrid/FluentDataGrid.razor
+++ b/src/Core/Components/DataGrid/FluentDataGrid.razor
@@ -29,19 +29,34 @@
                     @_renderColumnHeaders
                 </FluentDataGridRow>
             }
-
-            @if (Virtualize)
+            @if (Items?.Count() == 0 || _ariaBodyRowCount == 0)
             {
-                <Virtualize @ref="@_virtualizeComponent"
-                            TItem="(int RowIndex, TGridItem Data)"
-                            ItemSize="@ItemSize"
-                            ItemsProvider="@ProvideVirtualizedItems"
-                            ItemContent="@(item => builder => RenderRow(builder, item.RowIndex, item.Data))"
-                            Placeholder="@(placeholderContext => builder => RenderPlaceholderRow(builder, placeholderContext))" />
+                <div class="empty-content">
+                @if (EmptyContent is null)
+                {
+                     @("No data to show!");
+                }
+                else
+                {
+                    @EmptyContent
+                }
+                </div>
             }
             else
             {
-                @_renderNonVirtualizedRows
+                @if (Virtualize)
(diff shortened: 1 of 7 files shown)


src/Core/Components/DataGrid/FluentDataGrid.razor
```diff
@@ -29,19 +29,34 @@
                     @_renderColumnHeaders
                 </FluentDataGridRow>
             }
-
-            @if (Virtualize)
+            @if (Items?.Count() == 0 || _ariaBodyRowCount == 0)
```
dvoituron, line 32: It's better to use `.Any()`... to avoid to compute the number of all items.

dvoituron (approved):

src/Core/Components/DataGrid/FluentDataGrid.razor
vnbaaij: True, but in this case using `.Any()` gives Operator '||' cannot be applied to operands of type 'bool?' and 'bool'. Choose the easy way out :)

PR #758, opened 2023-09-20
Add PreventScroll to Dialog and Overlay

vnbaaij:
Added this parameter and associated code to enable scenario where content being shown behind dialog/overlay can not be scrolled


dvoituron:
Why did you add a parameter ´PreventScroll`, In my opinion, this should always be the case once an overlay is active.

vnbaaij:
I think so too, that is why it defaults to `true`. But it is a chang3efrom how it works now so wanted to a fallback option to go back to the current behavior

dvoituron (approved):

PR #791, opened 2023-09-26
Add a Tooltip parameter to DataGrid columns 

### LINKED ISSUE #787: feat: add tooltip to PropertyColumn
# 🙋 Feature Request

Would it be possible to add a tooltip property to PropertyColumn?

## 🤔 Expected Behavior

The display of a property column that is too small is that the overfllow is trimmed with an ellipsis. This looks great. But to see the full value, your only option is to expand the column width. It'd be nice to be able to hover over the truncated value and get the full value.

To be more flexible, this could just be a generalized ToolTip property on the column so the dev could put anything they wanted (even if different from the content).

## 😯 Current Behavior

Cell values can be truncated if the width is too small to display it and expanding the width of the column is the only way to see the full value

## 💁 Possible Solution

Adding a `title=""` attribute to the underlying fluent-data-grid-cell does the right thing here, so we just need a way to fill that in. There's already a Title property on PropertyColumn, so maybe a ToolTip property? 

## 🔦 Context

This can be achieved now by using a TemplateColumn, but having to convert everything from the simple PropertyColumn to TemplateColumns for this purpose seemed like a task better handled by the component itself.

## 💻 Examples

vnbaaij:
Also cleaned up the What's New pages a bit by movig pre-v3 content to an archive page
Fix #787

### DIFF
--- a/src/Core/Components/DataGrid/Columns/PropertyColumn.cs
+++ b/src/Core/Components/DataGrid/Columns/PropertyColumn.cs
@@ -96,5 +96,7 @@ public class PropertyColumn<TGridItem, TProp> : ColumnBase<TGridItem>, IBindable
 	protected internal override void CellContent(RenderTreeBuilder builder, TGridItem item)
 		=> builder.AddContent(0, _cellTextFunc!(item));
 
+    protected internal override string? RawCellContent(TGridItem item)
+        => _cellTextFunc!(item);
 
 }
--- a/src/Core/Components/DataGrid/Columns/PropertyColumn.cs
+++ b/src/Core/Components/DataGrid/Columns/PropertyColumn.cs
@@ -94,9 +94,9 @@ public class PropertyColumn<TGridItem, TProp> : ColumnBase<TGridItem>, IBindable
 
 	/// <inheritdoc />
 	protected internal override void CellContent(RenderTreeBuilder builder, TGridItem item)
-		=> builder.AddContent(0, _cellTextFunc!(item));
+		=> builder.AddContent(0, _cellTextFunc?.Invoke(item));
 
     protected internal override string? RawCellContent(TGridItem item)
-        => _cellTextFunc!(item);
+        => _cellTextFunc?.Invoke(item);
 
 }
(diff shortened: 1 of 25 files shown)


src/Core/Components/DataGrid/Columns/PropertyColumn.cs
```diff
@@ -96,5 +96,7 @@ protected override void OnParametersSet()
 	protected internal override void CellContent(RenderTreeBuilder builder, TGridItem item)
 		=> builder.AddContent(0, _cellTextFunc!(item));
 
+    protected internal override string? RawCellContent(TGridItem item)
+        => _cellTextFunc!(item);
```
dvoituron, line 100: Why to use `!` because at the end of the `OnParametersSet` method, the `_cellTextFunc` can be null.
Maybe better to use `?` (same in `CellContent` method).

src/Core/Components/DataGrid/FluentDataGrid.razor
```diff
@@ -83,7 +83,13 @@
             {
                 var col = _columns[colIndex];
 
-                <FluentDataGridCell GridColumn=@(colIndex+1) Class="@ColumnClass(col)" @key="@col" TGridItem="TGridItem" Item="@item">
+                string? tooltip = null;
```
dvoituron, line 86: Can be simplified to `string? tooltip = col.Tooltip ? col.RawCellContent(item) : null;`

vnbaaij: Done!

src/Core/Components/DataGrid/Columns/PropertyColumn.cs
vnbaaij: Can use  `?` there because the you get error saying it can't convert `Func<TGridItem, string?>` to `bool`

dvoituron: I think you have to use the Invoke method: `_cellTextFunc?.Invoke(item)`

vnbaaij: Yes, that works. Changing it now

PR #797, opened 2023-09-27
Fix MessageBar issues

vnbaaij:
- OnMessageItemsUpdatedAsync was not handeled
- Title is now nullable and will not be rendered if null
- Link is not generating an object anymore on initialization so also nor being rendered if null
- Sample had been adapted to show this behavior

### DIFF
--- a/src/Core/Components/MessageBar/FluentMessageBar.razor
+++ b/src/Core/Components/MessageBar/FluentMessageBar.razor
@@ -15,7 +15,10 @@
             
             @* Message *@
             <div class="fluent-messagebar-message">
-                <span class="title">@((MarkupString)Title)</span>
+                @if (!String.IsNullOrEmpty(Title))
+                {
+                    <span class="title">@((MarkupString)Title)</span>
+                }
                 @if (ChildContent is not null)
                 {
                     @ChildContent
--- a/src/Core/Components/MessageBar/FluentMessageBar.razor
+++ b/src/Core/Components/MessageBar/FluentMessageBar.razor
@@ -79,9 +79,12 @@
                 </div>
 
             @* Message *@
-            <div class="fluent-messagebar-notification-message" title="@Content">
-                @((MarkupString)Title)
-            </div>
+            @if (!String.IsNullOrEmpty(Title))
+            {
+                <div class="fluent-messagebar-notification-message" title="@Title">
+                    @((MarkupString)(Title))
+                </div>
+            }
 
             @* Close *@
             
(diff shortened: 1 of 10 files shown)


src/Core/Components/MessageBar/FluentMessageBarContainer.razor.cs
```diff
@@ -83,19 +83,29 @@ protected IEnumerable<Message> MessagesToShow
...
+            InvokeAsync(StateHasChanged);
+        });
+    }
+
     /// <summary />
     public void Dispose()
     {
-        MessageService.OnMessageItemsUpdated -= OnAlertUpdateHandler;
+        MessageService.OnMessageItemsUpdated -= OnMessageItemsUpdatedHandler;
+        MessageService.OnMessageItemsUpdatedAsync -= OnMessageItemsUpdatedHandlerAsync;
```
dvoituron, line 109: I think you should replace `!` by `?` on `await OnMessageItemsUpdatedAsync!.Invoke();`, in `MessageService` class (line 222)

src/Core/Components/MessageBar/FluentMessageBar.razor
```diff
@@ -15,7 +15,10 @@
             
             @* Message *@
             <div class="fluent-messagebar-message">
-                <span class="title">@((MarkupString)Title)</span>
+                @if (!String.IsNullOrEmpty(Title))
```
dvoituron, line 18: Could you also add a test here to avoid a warning during the compilation?
image

dvoituron: And, not the subject of this PR, but to avoid another warning in my previous code, could you update the `WithColor`method like this? (to avoid to create a small PR with that :-))
```csharp
public virtual Icon WithColor(string? color)
{
    if (!string.IsNullOrEmpty(color))
    {
        Color = color;
    }
        
    return this;
}
```

dvoituron (changes requested):

src/Core/Components/MessageBar/FluentMessageBarContainer.razor.cs
vnbaaij: No, can't change because of `await`

src/Core/Components/MessageBar/FluentMessageBar.razor
vnbaaij: Done!

dvoituron (approved):

PR #802, opened 2023-09-28
Several changes and fixes

### LINKED ISSUE #800: fix: NavMenu changing to base route by itself
# 🐛 Bug Report

This issue might be related to #789 fix. It seems that I can not go to child routes of a sub page whenever I have the `<Navigating>` component of the `<Router>` parent component. This does not happen when the base route is `/` in the `FluentNavMenu` component. Be aware that I'm using the latest FluentUI package that contains the fix for #789 .

## 💻 Repro or Code Sample

1. Clone [this repo](https://github.com/Narvalex/FluentUIBlazorTests)
2. Go to the `/sub-page` route
3. Try to go to a child route, like Counter page or Fetch Data page.
image

## 🤔 Expected Behavior

You should be able to go to `/sub-page/counter` and `/sub-page/fetchdata` and stay in that location.

## 😯 Current Behavior

The app redirects you to the `/sub-page` route after a few miliseconds.

## 💁 Possible Solution

If you remove the `<Navigating>` component from the `<Router>` the automatic redirection will not happen. 

## 🔦 Context

I'm trying to use the `FluentNavMenu` component without getting rid of the `Navigating` component.

## 🌍 Your Environment

.NET 7 and Fluent UI 3.1.1

vnbaaij:
- Fix #800 by doing more checks.
- Tooltip example uses `DemoLogger` now 
- Tooltip: add delegate check 
- `TemplateColumn`: Add `TooltipText` parameter. As `TemplateColumn` produces `RenderFragment` it cannot be converted to a string value used for the title attribute. Looking at adding this for `PropertyColumn` as well


nuget.config
```diff
@@ -9,8 +9,5 @@
        </repository>
      </trustedSigners>
    </config>
-   <packageSources>
-        <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
```
dvoituron, line 13: Why did you remove this line?

src/Core/Components/DataGrid/Columns/ColumnBase.razor
```diff
@@ -26,7 +26,7 @@
             
             if (Sortable.HasValue ? Sortable.Value : IsSortableByDefault())
             {
-                <FluentButton Appearance="Appearance.Stealth" class="col-sort-button" @onclick="@(() => Grid.SortByColumnAsync(this))" aria-label="Sort on this column">
+                <FluentButton Appearance="Appearance.Stealth" class="col-sort-button" @onclick="@(() => Grid.SortByColumnAsync(this))" aria-label="@($"Sort on '{@tooltip}'")" title="@tooltip">
```
dvoituron, line 29: To be compliant with Accessibility and to avoid localization issue, I would put `aria-label="@tooltip"` This way, the responsibility for writing a correct label is in the hands of the developer.

src/Core/Components/DataGrid/Columns/ColumnBase.razor.cs
```diff
@@ -36,10 +37,14 @@ public abstract partial class ColumnBase<TGridItem>
...
     /// <summary>
-    /// If true, generate a title attribute for the cell contents
+    /// If true, generate a title and aria-label attribute for the cell contents
     /// </summary>
     [Parameter] public bool Tooltip { get; set; } = false;
 
+    /// <summary>
+    /// Defines the value to be used as the tooltip and aria-label in this column's cells
+    /// </summary>
+    [Parameter] public Expression<Func<TGridItem, string?>>? TooltipText { get; set; }
```
dvoituron, line 47: Why did you use an `Expression` and not `Func<TGridItem, string?>`?

In my opinion, most of the time you'll need Func or Action if you're simply executing code. You need Expression when the code needs to be analyzed, serialized or optimized before being executed. Expression is used to think the code, Func/Action to execute it.

src/Core/Components/DataGrid/Columns/PropertyColumn.cs
```diff
@@ -26,12 +27,12 @@ public class PropertyColumn<TGridItem, TProp> : ColumnBase<TGridItem>, IBindable
 	/// </summary>
 	[Parameter, EditorRequired] public Expression<Func<TGridItem, TProp>> Property { get; set; } = default!;
```
dvoituron, line 28: Same here, why an Expression and not a Func?

nuget.config
vnbaaij: Because it was in mu list of package sources twice. Is it needed for your setup?

src/Core/Components/DataGrid/Columns/ColumnBase.razor
vnbaaij: Agreed!. Changing it.

nuget.config
dvoituron: OK. I don't think so (not checked :-)) .

src/Core/Components/DataGrid/Columns/ColumnBase.razor.cs
vnbaaij: I'm using the same construct that was already available in the `PropertyColumn` for the `Property` parameter. Short answer: it is faster. See also https://stackoverflow.com/questions/793571/why-would-you-use-expressionfunct-rather-than-funct?rq=2

src/Core/Components/DataGrid/Columns/PropertyColumn.cs
vnbaaij: Same reason

src/Core/Components/DataGrid/Columns/ColumnBase.razor.cs
dvoituron: I'm not sure that `Expression` is faster than `Func`. Probably for parsing the tree when you need to generate an SQL query, but not for executing a simple method like `c => c.Name`. We don't need to change this now, but I found the compiled code more complex

PR #808, opened 2023-09-30
Theme Services

AathifMahir:
# Pull Request

Theme Services that Merges Most Used Design Tokens to Create Centralized Theme System Where User can just Inject Theme Service and Do the Theme Changes Instead Inject Lots of Different Design Tokens

## 📖 Description

This pull request makes easy for users to do the theme changes without much of context

### General

### Component-specific


AathifMahir:
@vnbaaij This is initial Prototype of Theme Service that i talked about, I would like you opinion on that, Therefore i can move forward on this start readying this for production and also i would like to listen to your opionion on Best Place to Initialize the Theme Service, Are we giving this to users options to initialize the theme service in there desired layout or maybe creating a ThemeContainer ourself that does the Initialization

AathifMahir:

vnbaaij:
Hi @AathifMahir,

I've looked at your PR with my fellow maintainer @dvoituron. It needs quite a bit more work before we can think of merging this. We do see value in an enhancement like this.

Here are some of the things we noticed on the current state of the PR
- We adhere to certain coding standards. Please look at the rest of the source to see if your style is matching ours. Couple of examples: We don't use `Is...` or 'Has...` as a prefix for parameters, variables or methods. `if...then...` always have `{}`, etc.
- If we want do do persistence on a theme, it needs to be done through an interface so we can also enable storing the data in a cookie, file, local storage, etc.
- The goal of the demo is to showcase as many of the possibilities as possible. The option to select a color for the theme is one of them. Your PR abstract too much of that functionality.
- We are already way behind on having tests for our code. We do not take PRs that do not contain test for the functionality it contains.
- Some of the things you moved in to the theme service js file are very specific for the demo environment (like the `isMobile` check and the switching of the highlighter stylesheet. These should not be part of a generic service.

We hope this gives you some handles to move forward with this. Thanks for the effort so far!

AathifMahir:
> Hi @AathifMahir,
> 
> I've looked at your PR with my fellow maintainer @dvoituron. It needs quite a bit more work before we can think of merging this. We do see value in an enhancement like this.
> 
> Here are some of the things we noticed on the current state of the PR
> - We adhere to certain coding standards. Please look at the rest of the source to see if your style is matching ours. Couple of examples: We don't use `Is...` or 'Has...` as a prefix for parameters, variables or methods. `if...then...` always have `{}`, etc.
> - If we want do do persistence on a theme, it needs to be done through an interface so we can also enable storing the data in a cookie, file, local storage, etc.
> - The goal of the demo is to showcase as many of the possibilities as possible. The option to select a color for the theme is one of them. Your PR abstract too much of that functionality.
> - We are already way behind on having tests for our code. We do not take PRs that do not contain test for the functionality it contains.
> - Some of the things you moved in to the theme service js file are very specific for the demo environment (like the `isMobile` check and the switching of the highlighter stylesheet. These should not be part of a generic service.
> 
> We hope this gives you some handles to move forward with this. Thanks for the effort so far!

Thanks for the reply,

I'll start enhancing this based on the Fluent UI standards

AathifMahir:
Would you mind clarifying the second point, as that describes that Storage Related stuff done by interface? The current prototype includes an interface to do the persistent/storage or else I'm catching it? @vnbaaij

vnbaaij:
Yeah, you have an interface indeed. @dvoituron  do you want to chime in here?

examples/Demo/Shared/App.razor
```diff
@@ -1,11 +1,11 @@
 ﻿<Router AppAssembly="@typeof(App).Assembly">
     <Found Context="routeData">
-        <RouteView RouteData="@routeData" DefaultLayout="@typeof(DemoMainLayout)" />
+        <RouteView RouteData="@(routeData)" DefaultLayout="@typeof(DemoMainLayout)" />
```
dvoituron, line 3: Why do you change this code?

examples/Demo/Shared/Components/SiteSettings.razor.cs
```diff
@@ -8,46 +8,45 @@ public partial class SiteSettings
 {
     private IDialogReference? _dialog;
     
-    [Inject]
-    private GlobalState GlobalState { get; set; } = default!;
+    //[Inject]
```
dvoituron, line 11: Could you remove the unused code?

examples/Demo/Shared/Components/SiteSettingsPanel.razor
```diff
@@ -22,8 +21,8 @@
         </FluentSwitch>
         <FluentSwitch @bind-Value="@_rtl" title="Switch direction" AfterBindValue="UpdateDirection">
             <span>Direction</span>
-            <span slot="unchecked-message">LTR</span>
-            <span slot="checked-message">RTL</span>
+            <span slot="unchecked-message">RTL</span>
```
dvoituron, line 24: Why did you change this order? LTR is by default (and unchecked)

examples/Demo/Shared/Components/SiteSettings.razor.cs
```diff
@@ -8,46 +8,45 @@ public partial class SiteSettings
...
-        {
-            GlobalState? state = result.Data as GlobalState;
+    //    if (result.Data is not null)
+    //    {
+    //        GlobalState? state = result.Data as GlobalState;
 
-            GlobalState.SetDirection(state!.Dir);
-            GlobalState.SetLuminance(state.Luminance);
-            GlobalState.SetColor(state!.Color);
+    //        GlobalState.SetDirection(state!.Dir);
```
dvoituron, line 43: Could you remove the unused code?

examples/Demo/Shared/Components/SiteSettingsPanel.razor
```diff
@@ -7,11 +7,10 @@
             <FluentSelect Id="color"
                           Items=@(Enum.GetValues<OfficeColor>())
                           OptionValue="@(c => c.ToAttributeValue())"
-                          OptionSelected="@(c => c == _selectedColorOption)"
+                          OptionSelected="@(c => c == ThemeService.SelectedAccentColor)"
                           TOption="OfficeColor"
                           Position="SelectPosition.Below"
-                          @bind-SelectedOption="@_selectedColorOption"
-                          @onchange="@(x => HandleColorChange(x))"
+                          @onchange="@(x => ThemeService.SetAccentColorAsync(x.Value!.ToString()!.GetEnumByDescription<OfficeColor>()))"
```
dvoituron, line 13: Could you set the code in the .cs file. So, it's better to create a `HandleThemeChange` method with this code.

src/Core/Extensions/EnumExtensions.cs
```diff
@@ -35,4 +35,17 @@ public static class EnumExtensions
 
         return description;
     }
+
+    public static TEnum GetEnumByDescription<TEnum>(this string description) where TEnum : struct, Enum
```
dvoituron, line 39: Please, allow a nullable description and add an optional argument to ignore case sensitive description.
```csharp
public static TEnum? GetEnumByDescription<TEnum>(this string? description, bool ignoreCase = false)
     where TEnum : struct, Enum
{
    return Enum.GetValues<TEnum>()
                .FirstOrDefault(i => string.Compare(GetDescription(i), description, ignoreCase) == 0);
}
```

examples/Demo/Shared/Components/SiteSettingsPanel.razor
dvoituron, line 13: Avoid to use an exclamation mark to get a property value. If the compilator say that this object can be null, please check the value :-) And don't use .ToString to convert an object to a string but `Convert.ToString`

examples/Demo/Shared/Shared/DemoMainLayout.razor.cs
```diff
@@ -26,32 +26,35 @@ public partial class DemoMainLayout : IAsyncDisposable
     ErrorBoundary? errorBoundary;
 
 
-    [Inject]
-    private GlobalState GlobalState { get; set; } = default!;
+    //[Inject]
```
dvoituron, line 29: Remove unused code

src/Core/DesignTokens/Services/IThemeService.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Microsoft.AspNetCore.Components;
+
+namespace Microsoft.Fast.Components.FluentUI.DesignTokens;
+
+public interface IThemeService
```
dvoituron, line 5: Why to have properties Read Only but updatable via methods?
And not...
```csharp
public interface IThemeService
{
    StandardLuminance Theme { get; set; }
    OfficeColor AccentColor { get; set; }
    LocalizationDirection Direction { get; set; }
    Task SetDarkModeAsync(bool dark);
}
```

src/Core/DesignTokens/Services/IThemeService.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    StandardLuminance SelectedTheme { get; }
```
dvoituron, line 7: Missing comments to generate the documentation

src/Core/DesignTokens/Services/ThemeConstant.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace Microsoft.Fast.Components.FluentUI.DesignTokens;
+
+internal sealed class ThemeConstant
```
dvoituron, line 3: These constants are used only one time. It's better to define in the associated class.

src/Core/DesignTokens/Services/IThemeStorageService.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Microsoft.Fast.Components.FluentUI.DesignTokens;
+
+internal interface IThemeStorageService : IAsyncDisposable
```
dvoituron, line 3: What's the difference between this IThemeService and IThemeStorageService? 
The persistence service should be outsourced. It is not the purpose of this library to use LocalStorage. The best would be to have a `ThemeSerialized` property (for example) that allows you to retrieve a json with all the theme data, and update this property to reapply the theme.

src/Core/DesignTokens/Services/ThemeService.cs
```diff
@@ -0,0 +1,147 @@
+﻿using System.Drawing;
+using Microsoft.AspNetCore.Components;
+
+namespace Microsoft.Fast.Components.FluentUI.DesignTokens;
+
+internal sealed class ThemeService : IThemeService
```
dvoituron, line 6: Why the service is not public?

src/Core/Enums/LocalizationDirection.cs
```diff
@@ -14,4 +14,17 @@ public enum LocalizationDirection
     /// Right to left.
     /// </summary>
     rtl
+}
+
+public static class LocalizationDirectionExtension
```
dvoituron, line 19: This class should be in a file named `LocalizationDirectionExtension.cs`

src/Core/Enums/LocalizationDirection.cs
```diff
@@ -14,4 +14,17 @@ public enum LocalizationDirection
...
+{
+    public static bool IsRTL(this LocalizationDirection direction)
+    {
+        return direction is LocalizationDirection.rtl;
+    }
+
+    public static LocalizationDirection GetDirectionFromBoolean(this bool boolValue)
```
dvoituron, line 26: It's not a good idea to add an extension to a system type (bool): all uses of boolean will display this method.

src/Core/wwwroot/js/themeService.js
```diff
@@ -0,0 +1,32 @@
+export function switchHighlightStyle(dark) {
```
dvoituron, line 1: Why not place this JS file under the razor file `FluentThemeProvider.razor` and only import it when the dev uses it.

dvoituron:
Hi @AathifMahir
I think that the persistence service should be outsourced. It is not the purpose of this library to use LocalStorage. The best would be to have a `ThemeSerialized` property/method (for example) that allows you to retrieve a json with all the theme data, and update this property to reapply the theme.

I quickly reviewed your code and added a few comments. Please don't hesitate to contact me for further details.

dvoituron (changes requested):

AathifMahir:
> Hi @AathifMahir
> I think that the persistence service should be outsourced. It is not the purpose of this library to use LocalStorage. The best would be to have a `ThemeSerialized` property/method (for example) that allows you to retrieve a json with all the theme data, and update this property to reapply the theme.
> 
> I quickly reviewed your code and added a few comments. Please don't hesitate to contact me for further details.

Sure, I'll go through it. Thanks 👍

examples/Demo/Shared/App.razor
AathifMahir: Nothing Changed there, I had some test code there and removed it, seems like format gets messed up, I'll fix that

examples/Demo/Shared/Components/SiteSettings.razor.cs
AathifMahir: Sure, I'll make sure to remove all the unused codes

src/Core/DesignTokens/Services/ThemeService.cs
AathifMahir: Usually, I prefer to have concrete implementation as internal, seems like that applied here. I'll get that changed to public

dvoituron: Yes, if the service is used only in the Lib. But in this case is it a Service usable by users/devs? No?

examples/Demo/Shared/Components/SiteSettingsPanel.razor
AathifMahir: I'll undo that 👍

src/Core/DesignTokens/Services/IThemeService.cs
AathifMahir: I have exposed the properties on the intention of providing users an easier option to bind it to controls and components.

We have a 2 option here, either we keep the properties and add a setter and do the theme and other changes using property only or else we can remove the read-only properties and add method to get the current theme and others

examples/Demo/Shared/Components/SiteSettingsPanel.razor
AathifMahir: I'll make sure to follow that in the future PRs and update the current one as well

src/Core/Extensions/EnumExtensions.cs
AathifMahir: Sure, I'll make that change in the future commits

src/Core/DesignTokens/Services/IThemeStorageService.cs
AathifMahir: Seems like that's understandable, how about we create an nullable initial-data parameter that FluentThemeProvider component accepts it.

If we want to go an extra step, we can include a model aka Data Blueprint that the users pass directly to FluentThemeProvider component as initial data that we could have instant access to without serialization since the user does the serialization/deserialization from wherever they keep that data persistent.

In this case, if we have initial data we apply the theme based on that or revert back to default

src/Core/wwwroot/js/themeService.js
AathifMahir: That's the perfect place to keep it. I'll make sure to move under the components tree

src/Core/DesignTokens/Services/IThemeService.cs
dvoituron: I think it's best to have Get/Set properties. If possible.
Another "detail" is probably to convert the OfficeColor to a class with constants (ex. `OfficeColor.Access = "#a4373a"`). You can therefore use ´AccentColor` as a string. 
In this way, the user can also use the color of his choice, or take predefined colors from this list of constants.
@vnbaaij what is your opinion?

AathifMahir: > Why to have properties Read Only but updatable via methods?
> And not...
> ```csharp
> public interface IThemeService
> {
>     StandardLuminance Theme { get; set; }
>     OfficeColor AccentColor { get; set; }
>     LocalizationDirection Direction { get; set; }
>     Task SetDarkModeAsync(bool dark);
> }
> ```

We might need to go with the Method route since we are using asynchronous functions here to change the theme, accent color, etc..

In my opinion, it's best to keep the read-only property to get the current/selected Theme, color, etc.. and methods for changing themes etc..

c0g1t8:
Great code review - learned some things.

src/Core/DesignTokens/Services/ThemeService.cs
AathifMahir: I guess, around 99% use cases we don't need to expose the concrete implementation but the other 1% we might have a use-case where the user must be using a different DI container or maybe doing some custom implementation etc..

Anyway, we'll expose the concrete implementation and let the users decide what to do.

src/Core/DesignTokens/Services/IThemeStorageService.cs
AathifMahir: I have added Initial Data Support on the component and Model on Latest Commits, feel free to take a look and let me know. If it's ideal on this Scenerio

AathifMahir:
@vnbaaij @dvoituron any update on this PR, I have already updated most of the implementation to production ready with some minor exception

dvoituron:
> @vnbaaij @dvoituron any update on this PR, I have already updated most of the implementation to production ready with some minor exception

I haven't had time to finalize this review yet. I'll have a look as soon as I can.

vnbaaij:
Hi @AathifMahir, I propose we close this PR in favor of what we are discussing here: #1049

AathifMahir:
> Hi @AathifMahir, I propose we close this PR in favor of what we are discussing here: #1049 

Seems like it's achieves the same as this one, I guess I makes to sense to close. Go ahead with it 👍

PR #819, opened 2023-10-04
New NavMenu, NavGroup and NavLink components

vnbaaij:
The existing NavMenu components utilized the TreeView components for displaying the hierarchical structure and was not using the native NavLink capabilities to render links. This meant we need to do link/url/route checking ourselves which did not work _quite_ right.

This PR introduces new `FluentNavMenu`, `FluentNavGroup` and `FluentNavLinks` components which are based on the native `NavLink` component. 

The original `FluentNavMenu` component had been renamed to `FluentNavMenuTree`. The `FluentNavMenuGroup` and `FluentNavMenuLink` have remained unchanged. To keep using the original implementation, the only change that needs to be done is to do this rename action in your code. These 3 components are now marked obsolete and will be removed in a future version.

To upgrade to the new behavior, instructions have been added to the `Upgrade guide` 

Documentation and examples have been updated.


src/Core/Components/NavMenu/FluentNavBase.cs
```diff
@@ -0,0 +1,98 @@
...
+    [Parameter]
+    public NavLinkMatch Match { get; set; } = NavLinkMatch.Prefix;
+
+    [CascadingParameter(Name = "NavMenuExpanded")]
+    protected bool NavMenuExpanded { get; private set; }
+
+    /// <summary>
+    /// Returns <see langword="true"/> if the item has an <see cref="Icon"/> set.
+    /// </summary>
+    public bool HasIcon => Icon is not null;
```
dvoituron, line 70: Could you set this property `protected` or `internal`?

src/Core/Components/NavMenu/FluentNavBase.cs
```diff
@@ -0,0 +1,98 @@
...
+    [Parameter]
+    public bool ForceLoad { get; set; }
+
+    [Inject]
+    private NavigationManager NavigationManager { get; set; } = default!;
+
+    protected async Task OnClickHandler(MouseEventArgs ev)
+    {
+        if (Disabled)
+            return;
```
dvoituron, line 87: Could you set this syntax (Clean Code):
```csharp
if (Disabled)
{
    return;
}
```

src/Core/Components/NavMenu/FluentNavGroup.razor
```diff
@@ -0,0 +1,69 @@
...
+            </FluentNavMenu>
+        </FluentCollapsibleRegion>
+    </div>
+}
+@code {
+    private void RenderContent(RenderTreeBuilder __builder)
+    {
+        @if (Icon is not null)
+        {
+            <FluentIcon Value="@Icon" Width="20px" Color="@IconColor" Class="fluent-nav-icon" />
```
dvoituron, line 43: Do we need to set `20px` in a C# constant or in a CSS variable to allow future updates (of dev changes)?

src/Core/Components/NavMenu/FluentNavGroup.razor.cs
```diff
@@ -0,0 +1,112 @@
...
+    protected string? ClassValue =>
+        new CssBuilder("fluent-nav-group")
+            .AddClass("fluent-nav-item")
+            .AddClass("expanded", Expanded)
+            .AddClass("disabled", Disabled)
+            .AddClass(Class)
+            .Build();
+
+    internal string? StyleValue => new StyleBuilder(Style)
+       .AddStyle("margin", $"{Gap} 0" , !string.IsNullOrEmpty(Gap))
```
dvoituron, line 21: I'm not sure, but I think you have to set the unit after zero: `0px`

src/Core/Components/NavMenu/FluentNavGroup.razor.cs
```diff
@@ -0,0 +1,112 @@
...
+    public string? Gap { get; set; } = "10px";
+
+    /// <summary>
+    /// If set, overrides the default expand icon.
+    /// </summary>
+    [Parameter]
+    public Icon ExpandIcon { get; set; } = new CoreIcons.Regular.Size12.ChevronRight();
+
+    [Parameter]
+    public EventCallback<bool> ExpandedChanged { get; set; }
```
dvoituron, line 77: Could you add a comment for the doc?

src/Core/Components/NavMenu/FluentNavGroup.razor.cs
```diff
@@ -0,0 +1,112 @@
...
+        _renderButton = RenderButton;
+    }
+
+    private Task ToggleExpandedAsync() => SetExpandedAsync(!Expanded);
+   
+    private async Task HandleExpanderKeyDownAsync(KeyboardEventArgs args)
+    {
+        Task handler = args.Code switch
+        {
+            "Enter" => SetExpandedAsync(!Expanded),
```
dvoituron, line 91: You have also `NumpadEnter`

src/Core/Components/NavMenu/FluentNavLink.razor.cs
```diff
@@ -0,0 +1,33 @@
...
+using Microsoft.AspNetCore.Components.Web;
+using Microsoft.Fast.Components.FluentUI.Utilities;
+
+namespace Microsoft.Fast.Components.FluentUI;
+
+public partial class FluentNavLink : FluentNavBase
+{
+    private readonly RenderFragment _renderContent;
+
+    internal string? ClassValue => new CssBuilder("fluent-nav-item")
```
dvoituron, line 11: You always need to create the CssBuilder using the user Class: `new CssBuilder(Class).AddClass("fluent-nav-item")`.
To ensure the correct order of optional overloads.

src/Core/Components/NavMenu/FluentNavMenu.razor.cs
```diff
@@ -104,139 +80,13 @@ public FluentNavMenu()
...
-        }
-    }
+    private Task ToggleExpandedAsync() => SetExpandedAsync(!Expanded);
 
     private async Task HandleExpandCollapseKeyDownAsync(KeyboardEventArgs args)
     {
         Task handler = args.Code switch
         {
-            "Enter" => SetExpandedAsync(true),
+            "Enter" => SetExpandedAsync(value: !Expanded),
```
dvoituron, line 89: Same `NumpadEnter`

src/Core/Components/CollapsibleRegion/FluentCollapsibleRegion.razor.cs
```diff
@@ -0,0 +1,111 @@
...
+{
+
+    internal enum CollapseState
+    {
+        Entering, Entered, Exiting, Exited
+    }
+
+    //internal double _height;
+    private bool _expanded, _isRendered;  //, _updateHeight;
+    //private ElementReference _wrapper;
```
dvoituron, line 20: Could you remove unused code?

src/Core/Components/CollapsibleRegion/FluentCollapsibleRegion.razor.cs
```diff
@@ -0,0 +1,111 @@
...
+    //internal CollapseState _state = CollapseState.Exited;
+
+    protected string? StyleValue =>
+        new StyleBuilder()
```
dvoituron, line 24: You always need to create a style starting by the dev style:  `new StyleBuilder(Style)...`

src/Core/Components/CollapsibleRegion/FluentCollapsibleRegion.razor.cs
```diff
@@ -0,0 +1,111 @@
...
+            .AddStyle("max-height", MaxHeight, MaxHeight is not null)
+            .AddStyle("height", "auto", Expanded)
+            .AddStyle("height", "0", !Expanded)
+            .AddStyle(Style)
+            .Build();
+
+    protected string? ClassValue =>
```
dvoituron, line 31: Same here: `new CssBuilder(Class)...`

src/Core/Components/CollapsibleRegion/FluentCollapsibleRegion.razor.cs
```diff
@@ -0,0 +1,111 @@
...
+    /// <summary>
+    /// If true, the region is expaned, otherwise it is collapsed.
+    /// </summary>
+    [Parameter]
+    public bool Expanded
+    {
+        get => _expanded;
+        set
+        {
+            if (_expanded == value)
```
dvoituron, line 48: Could you use the syntax (Clean Code)

```csharp
if (_expanded == value)
{
    return;
}
```

src/Core/Components/CollapsibleRegion/FluentCollapsibleRegion.razor.cs
```diff
@@ -0,0 +1,111 @@
...
+                return;
+            _expanded = value;
+
+            if (_isRendered)
```
dvoituron, line 52: Could you remove these unused lines of code?

src/Core/Components/CollapsibleRegion/FluentCollapsibleRegion.razor.cs
```diff
@@ -0,0 +1,111 @@
...
+    public RenderFragment? ChildContent { get; set; }
+
+    [Parameter]
+    public EventCallback OnAnimationEnd { get; set; }
+
+    [Parameter]
+    public EventCallback<bool> ExpandedChanged { get; set; }
+
+    
+    //internal async Task UpdateHeight()
```
dvoituron, line 86: Could you remove this unused property?

dvoituron (commented):
First verification phase (syntax only).
I'd like to test the code live on my machine.
Can you provide some unit tests?

src/Core/Components/NavMenu/FluentNavGroup.razor.cs
tlmii: > Specifying the length unit is optional if the number is 0.

[MDN length](https://developer.mozilla.org/en-US/docs/Web/CSS/length)

Some of the CSS editors actually complain if you _do_ add a unit to 0. I actually thought it wasn't supposed to be there because of that, not just optional.

src/Core/Components/NavMenu/FluentNavGroup.razor
vnbaaij: Added `internal const string ICON_WIDTH = 20px;`

src/Core/Components/NavMenu/FluentNavGroup.razor.cs
vnbaaij: In Dev Tools it show as `0` when combined and as `0px` when separated out:
image
I'll leave it as-is.

src/Core/Components/NavMenu/FluentNavGroup.razor.cs
vnbaaij: Added. Also added `NumpadArrowLeft` and `...Right` (also in `FluentNavMenu.razor.cs`)

dvoituron:
Some questions.
1. Why the Item 5 seems accessible, but the property `Disabled=true`?
peek

2. By default, I don't think that the `Gap` between a `NavLink` and a `NavGroup` must be different.
image

dvoituron:
Probably interesting to add an maximum width and ellipsis when the text is too long.

image

src/Core/Components/NavMenu/FluentNavMenu.razor
```diff
@@ -1,39 +1,45 @@
...
+    {
+        title = null;
+    }
+    else 
+{
+        title = Title;
+    }
+}
+
+<div id="@Id" @attributes="AdditionalAttributes" class="@ClassValue" style="@StyleValue" aria-label="@title">
```
dvoituron, line 17: Could you set `aria-label="@(Collapsible ? null : title)"` and remove the code above?

src/Core/Components/NavMenu/FluentNavMenu.razor.css
```diff
@@ -1,95 +1,112 @@
...
+    box-sizing: border-box;
+    background: var(--neutral-fill-stealth-rest);
+    border: calc(var(--stroke-width) * 1px) solid transparent;
+    border-radius: calc(var(--control-corner-radius) * 1px);
+    min-height: calc(((var(--base-height-multiplier) + var(--density)) * var(--design-unit) + 1) * 1px);
 }
 
-::deep .navmenu.collapsed .treeitem-text, ::deep .navmenu.collapsed fluent-tree-item.navmenu-child-element::part(expand-collapse-button) {
-    display: none;
+/* NavGroup margins */
```
dvoituron, line 63: Could you remove the unused code?

vnbaaij:
> Some questions.
> 
> 1. Why the Item 5 seems accessible, but the property `Disabled=true`?

Small CSS issue. Was only checking for `a.disabled` instead of just `.disabled`

> 2. By default, I don't think that the `Gap` between a `NavLink` and a `NavGroup` must be different.
Agreed. Setting it to empty string as default

dvoituron (approved):

PR #822, opened 2023-10-04
[Input Label] FluentInputLabel Add a Label and update aria-label

dvoituron:
# FluentInputLabel

Add a new component to display a label in front of **all** Input Controls. And update the `aria-label` attribute of the associated input.
```razor
<FluentInputLabel ForId="@Id" Label="@Label" AriaLabel="@AriaLabel">@LabelTemplate</FluentInputLabel>
```

- `Label` is the text to be displayed, just above the component.
- `LabelTemplate` is the formatted content to be displayed, just above the component.
- `AriaLabel` is the text to be used as the `aria-label` attribute of the input. Default is `Label`

Components updated (no breaking changes if new new properties `Label` and `AriaLabel` are empty):
- FluentCheckbox
- FluentDatePicker
- FluentTimePicker
- FluentAutocomplete
- FluentCombobox
- FluentListbox
- FluentSelect
- FluentNumberField
- FluentRadio
- FluentRadioGroup
- FluentSearch
- FluentSlider
- FluentSwitch
- FluentTextArea
- FluentTextField

Example of update for **FluentTextField**.
```html
@namespace Microsoft.Fast.Components.FluentUI
@inherits FluentInputBase<string?>
<FluentInputLabel ForId="@Id" Label="@Label" AriaLabel="@AriaLabel" />
<fluent-text-field @ref=Element ...>
    @ChildContent
</fluent-text-field>
```

Example before using this ´FluentInputLabel`

image


src/Core/Components/Label/FluentInputLabel.razor.css
```diff
@@ -0,0 +1,11 @@
+﻿.fluent-input-label {
+    display: block;
+    color: var(--neutral-foreground-rest);
+    cursor: pointer;
+    font-family: var(--body-font);
+    font-size: var(--type-ramp-base-font-size);
+    line-height: var(--type-ramp-base-line-height);
+    font-weight: initial;
+    font-variation-settings: var(--type-ramp-base-font-variations);
+    margin-bottom: 4px;
```
vnbaaij, line 10: I would use `calc(var(--design-unit) * 1px)` so it scales along if someone changes this variable

src/Core/Components/Label/FluentInputLabel.razor.js
```diff
@@ -0,0 +1,10 @@
+﻿export function setInputAriaLabel(id, value) {
+    const fieldElement = document.querySelector("#" + id)?.shadowRoot?.querySelector("input");
+
+    console.log(id, value);
+    console.log(fieldElement);
```
vnbaaij, line 5: These are just here for development, I assume?

vnbaaij (approved):

src/Core/Components/Label/FluentInputLabel.razor.css
dvoituron: Done

dvoituron:
Question: I added the 'Title' property to represent the contents of `aria-label` because this property name is already used.
But found this name weird, for this purpose.
Maybe another name would be more appropriate like `AriaLabel`?

vnbaaij:
> Question: I added the 'Title' property to represent the contents of `aria-label` because this property name is already used.
> But found this name weird, for this purpose.
> Maybe another name would be more appropriate like `AriaLabel`?

Agreed. Think AriaLabel is better than Title

dvoituron:
> > Question: I added the 'Title' property to represent the contents of `aria-label` because this property name is already used.
> > But found this name weird, for this purpose.
> > Maybe another name would be more appropriate like `AriaLabel`?
> 
> Agreed. Think AriaLabel is better than Title

I think so... but we will have a Breaking change on some component already using `Title` : `FluentComponentBase.Title`. I will set this property `Obsolete`.

dvoituron:
Property ´Title` renamed to `AriaLabel`

Issue #771, opened 2023-09-23
fix: Dropdown appears below grid title in FluentSelect


JamesNK:
# 🐛 Bug Report

Select dropdown appears below the sticky grid title.

## 💻 Repro or Code Sample

![image](https://github.com/microsoft/fluentui-blazor/assets/303201/9ca9f5e3-a72a-49ee-8955-e7f5fd0b2dc3)

## 🤔 Expected Behavior

Appears above the sticky grid title.

## 😯 Current Behavior

Appears below the sticky grid title.

## 🌍 Your Environment

* Windows
* Chrome
* .NET 8 and Fluent Blazor 3.1

vnbaaij:
I'm not seeing this when combining a `FluentSelect` with a `FluentDataGrid` sample:

![select_with_grid](https://github.com/microsoft/fluentui-blazor/assets/1761079/63886fc6-d121-463d-8760-56c2bbd6078c)

You can see the sample at https://preview.fluentui-blazor.net/Lab/Overview.

Can you share some source (privately) perhaps?

JamesNK:
Your test example has overridden the z-index of the select listbox to a huge value:

![image](https://github.com/microsoft/fluentui-blazor/assets/303201/c102cab9-ddf5-461b-ba94-159957f81eda)

My listbox has a z-index of 1.

vnbaaij:
Ah, true but we do that for every `FluentSelect` **but only if** it has a value for the `Height` parameter

From `FluentSelect.razor`:
```
if (!String.IsNullOrEmpty(Height))
{
    <style>
        @($"#{Id}::part(listbox) {{ max-height: {Height}; z-index: {ZIndex.SelectPopup} }}")
        @($"#{Id}::part(selected-value) {{ white-space: nowrap; overflow: hidden; text-overflow: ellipsis; }}")
        @($"fluent-anchored-region[anchor='{Id}'] div[role='listbox'] {{ height: {Height}; }}")
    </style>
}
```

JamesNK:
Adding a height fixed it. The dropdown should probably have a max height anyway.

vnbaaij:
Closing this. Made a note on the min-height/default behavior

dvoituron:
I pushed this PR #774 to fix this issue.

Issue #744, opened 2023-09-19
fix: Button displayed above sticky header in FluentDataGrid


JamesNK:
# 🐛 Bug Report

A data grid with a sticky header doesn't work well with buttons in the grid. Buttons display above the sticky header instead of being hidden by it.

## 💻 Repro or Code Sample

```html
<div style="max-height: 600px; width: 100%; overflow-y: scroll;" class="StructuredLogsOverflow" tabindex="-1">
    <FluentDataGrid Class="StructuredLogsDataGrid" Virtualize="true" GenerateHeader="GenerateHeaderOption.Sticky" ItemSize="48" ResizableColumns="true" Items="@(_logEntries)" GridTemplateColumns="1fr 1fr 4fr 1fr 2fr 1fr">
        <PropertyColumn Title="Severity" Sortable="true" Property="@(context => context.Severity)" />
        <PropertyColumn Title="Timestamp" Sortable="true" Property="@(context => context.TimeStamp.ToLocalTime().ToLongTimeString())" />
        <PropertyColumn Title="Format" Property="@(context => context.Message)" />
        <PropertyColumn Title="Span" Property="@(context => context.SpanId)" />
        <PropertyColumn Title="Trace Id" Property="@(context => context.TraceId)" />
        <TemplateColumn>
            <FluentButton Appearance="Appearance.Hypertext"
                          OnClick="async() => await OnShowProperties(context)">View</FluentButton>
        </TemplateColumn>
    </FluentDataGrid>
</div>
```

![image](https://github.com/microsoft/fluentui-blazor/assets/303201/afc4fc9a-28ea-4250-8b31-b3fe11d5e43b)

Note, I tried changing the button appearance to "Hypertext", but it was still a button. Is that another bug?

## 🤔 Expected Behavior

The "View" button is hidden when it scrolls out of view.

## 😯 Current Behavior

The "View" button is above the header when scrolling.

## 🌍 Your Environment

Windows
Chrome
.NET 8 RC2 daily build

vnbaaij:
This is fixed in the next release (due this week)

PR #835, opened 2023-10-09
fix: remove FluentDataGridRow z-index property

LuohuaRain:
# Pull Request

## 📖 Description

When the z-index of `FluentDataGridRow` is set to 3, it conflicts with the styles of some components, such as `FluentSelect`. 
image

Therefore, the z-index property is removed from FluentDataGridRow.
image

### General

### Component-specific

### DIFF
--- a/src/Core/Components/DataGrid/FluentDataGridRow.razor.css
+++ b/src/Core/Components/DataGrid/FluentDataGridRow.razor.css
@@ -1,8 +1,6 @@
 ﻿.header {
     padding: 0;
-    z-index: 3;
 }
 
 .sticky-header {
-    z-index: 3;
 }


vnbaaij:
The z-index had been given this value for a reason. Please see #744. 
A solution for the list to appear above the grid was given in #771. Please confirm that this solution works for you as well. If not, then we can see how we can change this but this PR will not be merged in its current form.

LuohuaRain:
Thank you for pointing me to issues #744 and #771. I have reviewed the suggested solution in #771 and confirmed that it works for my case as well. Given that the current z-index value has its reasons, I agree that this PR is not needed in its current form. Thanks for your guidance.

PR #849, opened 2023-10-12
Add a custom web-components script to enable setting theme pref before Blazor starts

vnbaaij:
- Sets a darker background in Server app when pref is dark mode
- Add code to site.js to set theme based on browser setting
- Clean up layout


src/Core/wwwroot/Microsoft.Fast.Components.FluentUI.lib.module.js
```diff
@@ -159,7 +159,7 @@ export function afterStarted(blazor) {
 export function beforeStart(options, extensions) {
     var wcScript = document.createElement('script');
     wcScript.type = 'module';
-    wcScript.src = './_content/Microsoft.Fast.Components.FluentUI/js/web-components-v2.5.16.min.js';
+    wcScript.src = './_content/Microsoft.Fast.Components.FluentUI/js/web-components-v2.5.16-custom.min.js';
```
dvoituron, line 162: Why did you rename the web-component JS file to a "custom" file?

dvoituron (changes requested):

src/Core/wwwroot/Microsoft.Fast.Components.FluentUI.lib.module.js
vnbaaij: To make it perfectly clear that this is, in fact, a customized version of the standard script

dvoituron: Oh? I just try to compare https://cdn.jsdelivr.net/npm/@fluentui/web-components@2.5.16/dist/web-components.min.js and I did not find any difference. What are the changes? Is it possible to keep the "default" JS file and to include another JS function?

vnbaaij: No, I can't include another JS function. It is a change in the lib itself:
From 
`const baseLayerLuminance = create('base-layer-luminance').withDefault(StandardLuminance.LightMode);`
to :
`const baseLayerLuminance = create('base-layer-luminance').withDefault(window['DefaultBaseLayerLuminance'] || StandardLuminance.LightMode);`

This will allow people to adjust the start situation for the theme instead of always going with light mode first. There will be an example of this in the demo site.

dvoituron (approved):

Issue #864, opened 2023-10-17
fix: Accessibility Issue in Autocomplete with fluent-flipper


tlmii:
# 🐛 Bug Report

The fluent-flippers [included in the implementation](https://github.com/microsoft/fluentui-blazor/blob/106dae25e0c56efc109a9735e6030f6e03ec996f/src/Core/Components/List/FluentAutocomplete.razor#L26C1-L39C71) of the FluentAutoComplete don't have title or aria-labels set, so they fail the aria-command-name check in Accessibility Insights:

![image](https://github.com/microsoft/fluentui-blazor/assets/9613109/36475692-49b0-463c-a235-2d5a41ba46af)

## 💻 Repro or Code Sample

You can see this on https://www.fluentui-blazor.net/Autocomplete if you add enough elements to the first auto complete and the flippers are rendered.

## 🤔 Expected Behavior

Title or labels are rendered so that Accessibility Insights tests pass

## 😯 Current Behavior

See error in screenshot above

## 🌍 Your Environment
.NET 8, FluentUI Blazor 3.2, Windows 11, Edge, Accessibility Insights for Web

dvoituron:
Thanks for your feedback.
I will fix that quickly.

dvoituron:
Will be fixed with this PR, in the next Release: #866

PR #866, opened 2023-10-18
[Autocomplete] Update some Accessibility issues

dvoituron:
According to issue #864:
1. I added the properties `TitleScrollToPrevious` and `TitleScrollToNext` (by default defined with an English explanation).
2. The Autocomplete Popup contains a new ID value based on the `IdPopup` property (`= $"{Id}-popup"`).

Not all accessibility rules are fully validated yet, but the most important ones are set.

> Unit tests are updated. I added a file `_UpdateVerifiedFiles.cmd` to automatically updated all invalid `.verified.html` files.

image


src/Core/Components/List/FluentAutocomplete.razor.cs
```diff
@@ -127,6 +127,18 @@ public override bool Multiple
...
+    /// Title and Aria-Label for the Scroll to previous button.
+    /// </summary>
+    [Parameter]
+    public string TitleScrollToPrevious { get; set; } = "Scroll to previous";
+
+    /// <summary>
+    /// Title and Aria-Label for the Scroll to next button.
+    /// </summary>
+    [Parameter]
+    public string TitleScrollToNext { get; set; } = "Scroll to next";
```
vnbaaij, line 140: I would just set default for these to 'Previous' and 'Next'

vnbaaij (approved):

src/Core/Components/List/FluentAutocomplete.razor.cs
dvoituron: Why not... Fixed

PR #869, opened 2023-10-18
fix!: Implement DesignToken.WithDefault and use in Demo site

tlmii:
# Pull Request

## 📖 Description

FAST/FluentUI DesignTokens support setting a default value for a particular token as well as setting a (non-default) value for a particular container element (and its descendants). FluentUI-Blazor supports the latter, with DesignToken.SetValueFor, but WithDefault is currently not implemented.

When you truly want to set the value for a particular design token for all elements on the page (not just a subset), WithDefault is useful. It has the side benefit of not polluting the DOM with a large style attribute on whatever container element is used in SetValueFor.

Using WithDefault doesn't prevent you from using SetValueFor in other places in the page, as evidenced by the example pages for buttons and toolbars, and the design token page itself. 

This is technically a breaking change since it changes the signature of the WithDefault method from `DesignToken<T> WithDefault(T value)` to `ValueTask<DesignToken<T> WithDefault(T value)`, though it is unlikely anyone was using WithDefault previously since it had no effect at all. 

### 🎫 Issues

No issues currently that I'm aware of, but there's this discussion item: https://github.com/microsoft/fluentui-blazor/discussions/427

## 👩‍💻 Reviewer Notes

I'm throwing this up for discussion as I'm not sure what the intent is with WithDefault in this library. The discussion above (and the fact that it was left unimplemented) makes it seem intentional, but it feels like a gap to me. I made the change to the example/demo site so that it uses WithDefault for theme changes and it all works as far as I can tell.

## 📑 Test Plan

I'm not sure (though would be surprised) if this has any impact on unit tests, though they all pass for me locally.

### General

### Component-specific

## ⏭ Next Steps

If implemented, we'd probably want to update the design token documentation to reference WithDefault in addition to SetValueFor. But I didn't want to spend the time wordsmithing things yet if this wasn't a direction you wanted to go for the library.

### DIFF
--- a/examples/Demo/Shared/Components/SiteSettingsPanel.razor.cs
+++ b/examples/Demo/Shared/Components/SiteSettingsPanel.razor.cs
@@ -64,7 +64,7 @@ public partial class SiteSettingsPanel : IDialogContentComponent<GlobalState>, I
         Content.Dir = _dir;
 
         await _jsModule!.InvokeVoidAsync("switchDirection", _dir.ToString());
-        await Direction.SetValueFor(_container, _dir.ToAttributeValue());
+        await Direction.WithDefault(_dir.ToAttributeValue());
 
         StateHasChanged();
     }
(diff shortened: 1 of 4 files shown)


vnbaaij:
It was never implemented that way as I didn't know it would also apply the value given to the withdefault.
Think this is a super useful addition.
I agree with the statement on the breaking change not being impact full. Pretty sure we will just change it to this.
Not getting the huge style block is a big win as well.

dvoituron (approved):

vnbaaij (approved):

PR #874, opened 2023-10-20
Fix #861 and also fix rendering in case of manual grid

### LINKED ISSUE #861: fix: Accessibility issue in EmptyContent of DataGrid depending on content
# 🐛 Bug Report

When the FlentDataGrid is rendered and empty, a structure like this is generated (truncated for brevity):

```html
<fluent-data-grid role="grid">
  <fluent-data-grid-row role="row">
    <fluent-data-grid-cell role="columnheader">
      Header Content
    </fluent-data-grid-cell>
  </fluent-data-grid-row>
  <div class="empty-content">
    No data to show!
  </div>
</fluent-data-grid>
```

ARIA rules specify that the elements inside an element of role="grid" must have role="row" (or rowgroup, which contains row). Then of course inside an element with role="row" must have role="cell" (or gridcell, columnheader, rowheader)

If you run Accessibility Insights on a grid with the the empty content div rendered, this will pass even though it doesn't strictly meet the above rules.

But if you add an icon to the empty template, you'll get flagged for `aria-required-children`:

image

because of the aria-hidden attribute of the icon.

## 💻 Repro or Code Sample

This repros on any FluentDataGrid when rendered empty with an icon in it, though what the error in Accessibility Insights points to differs depending on what content the user puts in the EmptyContent

## 🤔 Expected Behavior

FluentDataGrid with no data passes Accessibility Insights testing without consumers of the component needing to worry about the structure of the data grid.

## 😯 Current Behavior

FluentDataGrid with no data fails Accessibility Insights testing depending on content of EmptyContent

## 💁 Possible Solution

I _think_ what you want to do is have something like this be rendered:

```html
<div class="empty-content" role="row">
  <div role="cell">
    No data to show!
  </div>
</div>
```
And that way there's no issue as to what is top-most inside the EmptyContent

## 🔦 Context

Just trying to handle accessibility pass in our app and this looked like one that should be handled by the component rather than on individual apps.

## 🌍 Your Environment

.NET 8, FluentUI 3.2, Windows 11, Edge, Accessibility Insight for Web plugin

vnbaaij:
Also fix DataGrid rendering for manual grid. It would always render the `EmptyContent` before


src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -265,6 +267,10 @@ private void StartCollectingColumns()
     private void FinishCollectingColumns()
     {
         _collectingColumns = false;
+        if (_columns.Count == 0)
```
dvoituron, line 270: It's better to use `Any()`
And why not to use `_manualGrid = !_columns.Any()`

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -232,6 +233,7 @@ protected override async Task OnAfterRenderAsync(bool firstRender)
         {
             _jsModule = await JS.InvokeAsync<IJSObjectReference>("import", "./_content/Microsoft.Fast.Components.FluentUI/Components/DataGrid/FluentDataGrid.razor.js");
             _jsEventDisposable = await _jsModule.InvokeAsync<IJSObjectReference>("init", _gridReference);
+
```
dvoituron, line 236: Could you remove this extra empty line?

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -171,6 +171,7 @@ public partial class FluentDataGrid<TGridItem> : FluentComponentBase, IHandleEve
     private int? _lastRefreshedPaginationStateHash;
     private object? _lastAssignedItemsOrProvider;
     private CancellationTokenSource? _pendingDataLoadCancellationTokenSource;
+    private bool _manualGrid;
```
dvoituron, line 174: Do you think better to use a private property "auto-calulcated"?
`private bool ManualGrid => !_columns.Any();`

dvoituron (approved):

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
vnbaaij: Yes, this is better

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
vnbaaij: No, this doesn't work as the _columns object is not populated yet at creation time.

tlmii:
@vnbaaij I _think_ we need another small tweak to this change. The `empty-content` class is still on the row, but should probably actually be on the cell. All the alignment styles it includes don't really get applied to anything since it always just has the single cell div as a child.

PR #926, opened 2023-11-05
Dialog: Add autoheight class if Height set to 'auto'

### LINKED ISSUE #915: Setting height on Dialog gets overriden.
# 🐛 Bug Report

When setting the height on a dialog and having the alignment set to Center the height property is overridden with height:auto.

## 💻 Repro or Code Sample

```
          IDialogReference dialog = await DialogService.ShowDialogAsync(typeof(SearchDialog<T>), request, new DialogParameters()
          {
              Title = ModalTitle ?? "Search",
              Alignment = HorizontalAlignment.Center,
              Width = "90vw",
              Height = "90vh"
          });
```

## 🤔 Expected Behavior

Should be set to the height requested.

## 😯 Current Behavior

Is set to auto.

image

vnbaaij:
Resolve #915 by adding an autoheight class when Height set to auto. Also added a test for this.


src/Core/Components/Dialog/FluentDialog.razor.cs
```diff
@@ -24,6 +24,7 @@ public partial class FluentDialog : FluentComponentBase
         .AddClass("right", () => _parameters.DialogType == DialogType.Panel && _parameters.Alignment == HorizontalAlignment.Right)
         .AddClass("left", () => _parameters.DialogType == DialogType.Panel && _parameters.Alignment == HorizontalAlignment.Left)
         .AddClass("prevent-scroll", () => Instance is null ? (PreventScroll && !Hidden): _parameters.PreventScroll)
+        .AddClass("autoheight", () => !string.IsNullOrWhiteSpace(_parameters.Height) && _parameters.Height == "auto")
```
dvoituron, line 27: Probably `() => _parameters.Height == "auto"` is enough. Why to add an extra test on "Not Empty"?

src/Core/Components/Dialog/FluentDialog.razor.css
```diff
@@ -3,12 +3,13 @@
     position: relative;
     padding: 0px var(--dialog-padding);
     width: calc(var(--dialog-width) - 2 * var(--dialog-padding));
-    height: calc(var(--dialog-height) - 2 * var(--dialog-padding));
-    max-width: 80%;
 }
```
dvoituron, line 6: Why did you remove `max-width` ?

src/Core/Components/Dialog/FluentDialog.razor.css
```diff
@@ -3,12 +3,13 @@
...
 
-fluent-dialog:has(div[class~="fluent-dialog-body"]):not(:is(fluent-dialog[class~="right"], fluent-dialog[class~="left"]) )::part(control) {
+fluent-dialog.autoheight::part(control) {
```
dvoituron, line 8: When `Height` is a specified value. How the height is set, if you removed `height: calc(var(--dialog-height) - 2 * var(--dialog-padding));`?

Is it possible to keep the height in `fluent-dialog::part(control)` and to override with this style and an `!important` flag?

tests/Core/Dialog/FluentDialogServiceTests.razor
```diff
@@ -23,4 +23,26 @@
         // Assert
         cut.Verify();
     }
+
+    [Fact]
```
dvoituron, line 27: Probably interesting to add a second test with a specific Height (ex: `Height="240px").

src/Core/Components/Dialog/FluentDialog.razor.css
vnbaaij: Because on mobile I want a dialog or panel to take the whole width of the screen. Needs more changes probably

src/Core/Components/Dialog/FluentDialog.razor.css
vnbaaij: Height is then set as `--dialog-height` in the style of the fluent-dialog. There is actually no need to set the heigth in the  part(control) again

src/Core/Components/Dialog/FluentDialog.razor.cs
vnbaaij: You are right. Will change that

tests/Core/Dialog/FluentDialogServiceTests.razor
vnbaaij: Yes, will do

vnbaaij: The test above that one already has a height set

src/Core/Components/Dialog/FluentDialog.razor.css
dvoituron: In this case, we need to add a `@media` style to remove the `max-width` on mobile.
And I'm not even sure it's a good idea to accept 100% on mobile. Because a DialogBox should look like a window, even on mobile (in my opinion).

src/Core/Components/Dialog/FluentDialog.razor.css
dvoituron: Including the `2 * var(--dialog-padding)`? So, why to keep `width`?

dvoituron (approved):

PR #1016, opened 2023-11-23
replace max-width to width

twytec:
# Pull Request

## 📖 Description

```razor
<FluentStack Orientation="Orientation.Vertical" VerticalGap="10" Width="100%">
    <FluentToolbar Style="width: 100%;">
        <FluentButton>Foo</FluentButton>
    </FluentToolbar>
    <FluentListbox TOption="string"
                   Items="data"
                   Multiple="false" Height="100px" Width="100%" />
</FluentStack>

@code {
    List<string> data = new() { "Foo", "Bar" };
}
```

Screenshot 2023-11-23 234337

### General

- [ ] I have read the [CONTRIBUTING](https://github.com/Microsoft/fluentui-blazor/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://www.fast.design/docs/community/code-of-conduct/#our-standards) for this project.

### Component-specific

### DIFF
--- a/src/Core/Components/List/FluentListbox.razor.cs
+++ b/src/Core/Components/List/FluentListbox.razor.cs
@@ -14,7 +14,7 @@ public partial class FluentListbox<TOption> : ListComponentBase<TOption>
 
     /// <summary />
     protected virtual StyleBuilder BorderStyle => new StyleBuilder()
-        .AddStyle("max-width", Width, when: !string.IsNullOrEmpty(Width))
+        .AddStyle("width", Width, when: !string.IsNullOrEmpty(Width))
         .AddStyle("max-height", Height, when: !string.IsNullOrEmpty(Height))
         .AddStyle("overflow-y", "auto")
         .AddStyle("border", "calc(var(--stroke-width) * 1px) solid var(--neutral-stroke-rest)")


vnbaaij:
Hi,

We are missing a bit of information here. What problem are you trying to solve? Is the image from the before of after situation?

Also, moving forward we are going to be way more strict on PRs having unit tests associated with the changes. Please take a look at the [supplied docs](https://github.com/Microsoft/fluentui-blazor/blob/master/unit-tests.md) and provide tests for this change.

dvoituron:
I think you also need to correct the `Height` value:

```csharp
protected virtual StyleBuilder BorderStyle => new StyleBuilder()
    .AddStyle("width", Width, when: !string.IsNullOrEmpty(Width))                     // <==
    .AddStyle("height", Height, when: !string.IsNullOrEmpty(Height))                  // <==
    .AddStyle("overflow-y", "auto")
    .AddStyle("border", "calc(var(--stroke-width) * 1px) solid var(--neutral-stroke-rest)")
    .AddStyle("border-radius", "calc(var(--control-corner-radius) * 1px)");
```

twytec:
Hi,

sorry my english is ultimate bad :)

A width of 100% is specified in the FluentListbox. 
But the setting becomes max-width: 100%.
It is also not possible to define your own style. Since the style only works on the inner div.

```html
<div style="@BorderStyle.Build()"> <!--PARENT (max-width: 100%)-->
   ...
   <fluent-listbox @ref="Element" 
                        ...
                        style="@($"{StyleValue} border: none; width: 100%;")" <!--INNER DIV-->
   </fluent-listbox>
</div>
```

I only have one option:
```html
<FluentStack Style="align-items: stretch"></FluentStack>
```

I understand the importance of unit testing. But for that

twytec:
> I think you also need to correct the `Height` value:
> 
> ```cs
> protected virtual StyleBuilder BorderStyle => new StyleBuilder()
>     .AddStyle("width", Width, when: !string.IsNullOrEmpty(Width))                     // <==
>     .AddStyle("height", Height, when: !string.IsNullOrEmpty(Height))                  // <==
>     .AddStyle("overflow-y", "auto")
>     .AddStyle("border", "calc(var(--stroke-width) * 1px) solid var(--neutral-stroke-rest)")
>     .AddStyle("border-radius", "calc(var(--control-corner-radius) * 1px)");
> ```

I like the max-height `height` value

dvoituron:
> > I think you also need to correct the `Height` value:
> > ```cs
> > protected virtual StyleBuilder BorderStyle => new StyleBuilder()
> >     .AddStyle("width", Width, when: !string.IsNullOrEmpty(Width))                     // <==
> >     .AddStyle("height", Height, when: !string.IsNullOrEmpty(Height))                  // <==
> >     .AddStyle("overflow-y", "auto")
> >     .AddStyle("border", "calc(var(--stroke-width) * 1px) solid var(--neutral-stroke-rest)")
> >     .AddStyle("border-radius", "calc(var(--control-corner-radius) * 1px)");
> > ```
> 
> I like the max-height `height` value

Perhaps, but the problem is the same: without the correct `height` style, the ListBox won't be displayed correctly.
If you decide to set a width or a height values, the result must be this value.

twytec:
> Perhaps, but the problem is the same: without the correct `height` style, the ListBox won't be displayed correctly.
> If you decide to set a width or a height values, the result must be this value.


Yes, the problem is the same, but the correct height can be achieved by adding x items.
Achieving the correct width is more difficult.
max-height is not correct, but could stay there

dvoituron:
No, when using `max-height`, the result will not always be correct.
For example, if the number of elements is less than the height, the result will not be correct.

image

twytec:
Yes that is correct. Thanks

vnbaaij:
What is the status of this? Can we close th PR?

dvoituron:
The `max-height` value has not yet been changed to `height`. So it's not a valid PR at the moment, in my opinion.

MoodieG:
Hey, I am running into the same issue. I agree with @dvoituron, that `max-height` should also be changed to `height`. Looking forward for this fix to get in.

vnbaaij:
Replaced with PR #1083. Closing this one.

Issue #429, opened 2023-06-04
feat: add Localization to FluentPaginator


damianog:
# 🙋 Feature Request

Add string localization to FluentPaginator

## 🤔 Expected Behavior

Static string text should be displayed in local language

## 😯 Current Behavior

The pager display text in plain English

## 💁 Possible Solution

Implement IStringLocalizer

## 🔦 Context

I want to display the paginator text strings according to browser language.
Users ask for text translation.
I considered using a custom component but native is better.

## 💻 Examples

Current string "items", "Page", "of" are hardcoded:

![image](https://github.com/microsoft/fluentui-blazor/assets/2786682/97a86578-368c-4826-b948-b2aa550ea9d8)

Browsing in Italian language I should see:

![image](https://github.com/microsoft/fluentui-blazor/assets/2786682/9fe16e70-c866-4b69-b9f4-2e71d478ccb2)

vnbaaij:
Good suggestion. We would love to get a PR for this (and other languages)

damianog:
Hi @vnbaaij 

What is better for you? A single Resources folder inside the main project with all localized resource files or a separate project?

dvoituron:
It's likely that a single folder in the Core project would be best. To avoid a multitude of small Dlls (at least for the moment).

vnbaaij:
We are thinking to have a resx file per component (to keep it manageable) all in one Resources folder. We are working on this for the next version.

dvoituron:
I come back on my previous comment.

I don't think it's the library's role to manage translations.
It's a process that's far too complex depending on the number of languages to be maintained and on the specificity of each language: singular/plural, feminine/masculine, ...

On the other hand, I think it's preferable to give the developer a way to manage "localizable" content via RenderTemplate or string properties. For example: 'DialogParameters.DismissTitle.

vnbaaij:
This has been resolved by providing `SummaryTemplate` and `PaginationTextTemplate` parameters

Tolitech:
Hello, @vnbaaij .
How about the titles of the buttons Go to first, Go to previous, Go to next, and Go to last could also be passed as parameters?

Currently, I will need to rewrite the logic in the PaginationTextTemplate just to allow putting the button titles in the user's culture.

vnbaaij:
We will address localization in a later version.

PR #1018, opened 2023-11-25
FluentPaginator Localization

agriffard:
Related to #429

Enable Localization in WebAssembly Demo.Client.
CultureSelector Components to switch between languages. (adapted from this sample https://cosmin-vladutu.medium.com/localization-in-blazor-and-net-7-12cc959e6ecf).

Localization of FluentPaginator component in English and French.
Plural of item/items depending of the total count (even if plural is more complicated than that in some other languages).

Animation


dvoituron:
I don't think it's the library's role to manage translations.
It's a process that's far too complex depending on the number of languages to be maintained and on the specificity of each language: singular/plural, feminine/masculine, ...

On the other hand, I think it's preferable to give the developer a way to manage "localizable" content via RenderTemplate or string properties. For example: '`DialogParameters.DismissTitle`.

vnbaaij:
We've added a PaginationTextTemplate parameter with which is is now possible to change the text shown to your likening. It is not a true localization solution, but for now this is what we can and want to do about this, Closing this PR as a result

PR #1022, opened 2023-11-25
[FluentCheckbox] Add 3-states (check/uncheck/indeterminate)

dvoituron:
# [FluentCheckbox] Add 3-states (check/uncheck/indeterminate)

Added these new properties:
- **ThreeState**: Gets or sets a value indicating whether the Checkbox will allow three check states rather than two.
- **ShowIndeterminate**: Gets or sets a value indicating whether the user can display the indeterminate state by clicking the Checkbox. If this is not the case, the checkbox can be started in the indeterminate state, but the user cannot activate it with the mouse. Default is true.
- **CheckState**: Gets or sets the state of the CheckBox: true, false or null.

The current `Value` and the new `CheckState` properties are bindable.

## Result

The first example has `ThreeState = true` and the user can switch between Unchecked / Checked / Indeterminate.

The second example has `ThreeState = false` and the user can only switch between Unchecked / Checked.

The third example has `ThreeState = true` and `ShowIndeterminate = true`, so the first state can be Indeterminate (or updated by code), but the user can only switch between Unchecked / Checked.

TheeStates

## Unit Tests

Unit tests included.
image


src/Core/Components/Checkbox/FluentCheckbox.razor.cs
```diff
@@ -1,21 +1,189 @@
...
 using Microsoft.AspNetCore.Components;
+using Microsoft.JSInterop;
 
 namespace Microsoft.FluentUI.AspNetCore.Components;
 public partial class FluentCheckbox : FluentInputBase<bool>
 {
+    private const bool VALUE_FOR_INDETERMINATE = false;
+    private bool _intermediate = false;
+    private bool? _checkState = false;
+    private const string JAVASCRIPT_FILE = "./_content/Microsoft.FluentUI.AspNetCore.Components/Components/CheckBox/FluentCheckbox.razor.js";
```
vnbaaij, line 11: In the js path CheckBox -> Checkbox.
Leads to errors in SWA because of case sensitivity

vnbaaij (changes requested):

src/Core/Components/Checkbox/FluentCheckbox.razor.cs
dvoituron: I fix that now

dvoituron: Done

vnbaaij:
Great! Generated SWA works now.

vnbaaij (approved):

PR #1048, opened 2023-12-01
[FluentGrid] - Hiding elements

dvoituron:
# [FluentGrid] - Hiding elements

For faster, mobile-friendly development, use responsive display classes to show and hide elements according to device. Avoid creating entirely different versions of the same site, but instead hide elements reactively for each screen size.

To hide elements, simply use the **HiddenWhen** attribute by selecting one of the following values or combining them with the | operator (see [Enumeration flags](https://learn.microsoft.com/en-us/dotnet/api/system.flagsattribute))

To display an element only on a given range of screen sizes, you can combine a these values. e.g. `GridItemHidden.Md | GridItemHidden.Lg` will display the element for all screen sizes, except on medium and large devices.

Example:
```xml
<FluentGrid>
    <FluentGridItem xs="12"> ... </FluentGridItem>
    <FluentGridItem xs="12" sm="6" HiddenWhen="GridItemHidden.SmAndDown"> ... </FluentGridItem>
    <FluentGridItem xs="12" sm="6"> ... </FluentGridItem>
</FluentGrid>
```

image

## OnBreakPointEnterevent

> Previously, the `OnSizeEnter` event has been renamed `OnBreakPointEnter` (for better understanding).

A `FluentGrid.OnBreakPointEnter` event is triggered when a **compatible media query** is verified.

Example
```xml
<FluentGrid OnBreakPointEnter="@OnBreakPointEnterHandler"> ... </FluentGrid>
``` 

```csharp
void OnBreakPointEnterHandler(GridItemSize size)
{
    Console.WriteLine($"Page Size: {size}");
    DemoLogger.WriteLine($"Page Size: {size}");
}
```
GridResize

## Unit Tests

image


vnbaaij (approved):

mguinness:
Is it possible to use `OnSizeEnter.HasDelegate` to prevent calling [FluentGridInitialize](https://github.com/microsoft/fluentui-blazor/pull/1048/files#diff-4c697ef493623b9e8983852bd11e084043a90bb74db9e169bc811304a5492ac2R74)?  Avoids adding event listeners unnecessarily at the client where no event delegate was specified.

dvoituron:
The `OnSizeEnter` event has been renamed `OnBreakPointEnter` (for better understanding).

Issue #1055, opened 2023-12-01
Validation


boukenka:
# 🙋 Feature Request

I expected to find a field-level validation example in the documentation, not only a validation summary example.
I've tried the following to show validation errors. Is this the right method? 
Do you intend to implement the [feat: show validation error below form components like TextField or Checkbox](https://github.com/microsoft/fluentui-blazor/issues/838) ? 

```C#
<EditForm Model="@Data" OnValidSubmit="ValidHandlerAsync">
    <DataAnnotationsValidator />
     <FluentTextField Label="Last name:"
                      Placeholder="Enter your last name"
                      @bind-Value="@Data.LastName" Immediate="true" />
     <ValidationMessage For="@(() => Data.LastName)" />
</EditForm>

public class RegisterUser
{
    [Required(ErrorMessage = "Not empty!")]
    [MinLength(20, ErrorMessage = "Too short!")]
    public string? LastName { get; set; }
}
```

## 🤔 Expected Behavior

I thought the **FluentTextField** component would have methods/parameters dedicated to validation? And the same goes for the other components.

PS : Thank you for the nice video during the .NET Conference 2023.

vnbaaij:
Yes, we still intend to implement the feature you mention. But as said there, it might not appear until we have the v3 web components.

The code you posted already works today, so you can use that. You are using the right method. Might just need a bit more styling.

PS: Thanks. We liked doing the .NET Conf talk. Glad it showed!

PR #1079, opened 2023-12-06
Add FluentValidationMessage and Required

### LINKED ISSUE #838: feat: show validation error below form components like TextField or Checkbox
# 🙋 Feature Request

It would be great to have the validation error directly below the form components like TextField, Checkbox, TextArea etc.

## 🤔 Expected Behavior

In the Fluent 2 Design System this is described as Field, see: https://fluent2.microsoft.design/components/web/react/field/usage
Additionally having all the other capabilities other than the validation error would be awesome.

## 😯 Current Behavior

E.g. for a TextField the border is highlighted in red and the only option to show the validation error is via the FluentValidationSummary. 

## 💁 Possible Solution

From an end user perspective I wouldn't change much if anything. For e.g. a TextField it should stay the same but only the error message should be additionally shown. There might be an additional parameter to hide / show the message so it's customizable. The problem with a parameter is that a form can be pretty big and you actually only need to set it once for all components. You would either pick to always show the error message for all components or not at all.

Internally there might actually be a Field components which wraps all form components and does the job. MudBlazor has something like that, see: https://github.com/MudBlazor/MudBlazor/blob/dev/src/MudBlazor/Components/InputControl/MudInputControl.razor

A separate component would be helpful for advanced scenarios. With it, it would be possible to implement custom form components and use such a Field component to add validation error capabilities and all the accessibility features coming along with it.

vnbaaij:
Initial work to add FluentValidationMessage and a Reuired indicator after the label.
This does not yet contain the global flag we discussed. Fix #838 #1055


src/Core/Components/Label/FluentInputLabel.razor.cs
```diff
@@ -14,6 +14,9 @@ public partial class FluentInputLabel
     /// <summary />
     private IJSObjectReference? Module { get; set; }
 
+    [Parameter]
+    public FluentInputBase<TValue>? Input { get; set; }
```
dvoituron, line 18: I don't think that is a good idea to link a all input components to a `FluentInputLabel`.
You have `ForId` to do this (of course, only in HTML but what if the `ForId` is a listbox and the `Input` is a checkbox?).

Isn't it possible to do things differently?

src/Core/Components/Label/FluentInputLabel.razor
```diff
@@ -1,9 +1,14 @@
...
 {
-    <label for="@ForId" class="fluent-input-label" @attributes="@AdditionalAttributes">
-        @Label
-        @ChildContent
+    <label for="@(Input?.Id ?? ForId)" class="fluent-input-label" @attributes="@(Input?.AdditionalAttributes ?? AdditionalAttributes)" aria-label="@(Input?.AriaLabel ?? AriaLabel)">
+        @(Input?.Label ?? Label)
+        @(Input?.LabelTemplate ?? ChildContent)
+        @if (Input?.Required ?? Required)
+        {
+            <span aria-label="required" aria-hidden="true" style="padding-inline-start: 4px; color: var(--error);">*</span>
```
dvoituron, line 11: In a first time, could you put this in a **public static** variable.

```xml
public static MarkupString? RequiredLabel = (MarkupString)"<span aria-label=\"required\" aria-hidden=\"true\" style=\"padding-inline-start: 4px; color: var(--error);\">*</span>";
```

So, the dev could update/remove "globally" this markup.

tests/Core/Slider/FluentSliderTests.razor
```diff
@@ -6,7 +6,7 @@
     public void FluentSlider_HorizontalImplicit()
     {
         // Arrange && Act
-        var cut = Render(@<FluentSlider Min="0" Max="100" Step="10"></FluentSlider>);
+        var cut = Render(@<FluentSlider Min="0" Max="100" Step="10" Label="label"></FluentSlider>);
```
dvoituron, line 9: Could you avoid updating existing tests to check for new features?
It would be better to create new tests to validate the `Label` attribute.

tests/Core/Slider/FluentSliderTests.razor
```diff
@@ -38,7 +38,7 @@
     {
         double value = 3.5;
         // Arrange
-        var cut = Render(@<FluentSlider Min="-1.50" Max="7.0" Step="0.5" @bind-bind-Value="@value"></FluentSlider>);
+        var cut = Render(@<FluentSlider Min="-1.50" Max="7.0" Step="0.5" @bind-Value="@value" Label="label"></FluentSlider>);
```
dvoituron, line 41: Weird, before the tests approved `value="0"` and not 3.5. Any explanation?

src/Core/Components/Icons/CoreIcons.cs
```diff
@@ -10,6 +10,20 @@ namespace Microsoft.FluentUI.AspNetCore.Components;
 
 internal static partial class CoreIcons
 {
+    internal static partial class Filled
+    {
+        [global::System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage]
+        internal static partial class Size12
+        {
+            public class CheckmarkCircle : Icon { public CheckmarkCircle() : base("CheckmarkCircle", IconVariant.Filled, IconSize.Size12, "<path d=\"M1 6a5 5 0 1 1 10 0A5 5 0 0 1 1 6Zm7.35-.9a.5.5 0 1 0-.7-.7L5.5 6.54 4.35 5.4a.5.5 0 1 0-.7.7l1.5 1.5c.2.2.5.2.7 0l2.5-2.5Z\"/>") { } }
```
dvoituron, line 18: What are these icons for?

dvoituron (changes requested):
As I put it in the code, I think it's a problem to link FluentInputLabel to the component. For me, it's better to use a composition of components. Especially as the way of representing an Input Label may change from one component to another.

The `FluentInputLabel` could only use its Required property to display this asterisk or not.

src/Core/Components/Label/FluentInputLabel.razor.cs
vnbaaij: Look at it more as an addition of a 'shortcut'. If you want the InputLabel to be connected to the input, you can just use `this` for its value and then all the needed other parameters can be gotten from the Input value instead of specifying them all separately. But all the other parameters all still there so nothing is broken/changed from what was there before

tests/Core/Slider/FluentSliderTests.razor
vnbaaij: I think that has to do with the `@bind-bind-Value` that was there before

src/Core/Components/Label/FluentInputLabel.razor.cs
vnbaaij: As agreed, I remove the `Input` parameter an specify the other parameters where needed

tests/Core/Slider/FluentSliderTests.razor
vnbaaij: Tests are in order now

src/Core/Components/List/FluentCombobox.razor
```diff
@@ -10,7 +10,7 @@
 }
 
 <CascadingValue Value=@_internalListContext Name="ListContext" TValue="InternalListContext<TOption>" IsFixed=true>
-    <FluentInputLabel ForId="@Id" Label="@Label" AriaLabel="@AriaLabel" ChildContent="@LabelTemplate" />
+    <FluentInputLabel ForId="@Id" Label="@Label" AriaLabel="@AriaLabel" ChildContent="@LabelTemplate" Required=@Required />
```
dvoituron, line 13: To be coherent with other attributes, could you add double-quotes: `Required="@Required"`?

src/Core/Components/DateTime/FluentDatePicker.razor
```diff
@@ -1,7 +1,7 @@
 @namespace Microsoft.FluentUI.AspNetCore.Components
 @inherits FluentCalendarBase
 
-<FluentInputLabel ForId="@Id" Label="@Label" AriaLabel="@AriaLabel" ChildContent="@LabelTemplate" />
+<FluentInputLabel ForId="@Id" Label="@Label" AriaLabel="@AriaLabel" ChildContent="@LabelTemplate" Required=@Required TValue="DateTime?" />
```
dvoituron, line 4: To be coherent with other attributes, could you add double-quotes: Required="@Required"?

dvoituron, line 4: What is this `TValue`?

dvoituron (changes requested):

src/Core/Components/DateTime/FluentDatePicker.razor
vnbaaij: Left-over from previos Input paramer thing. Will remove

dvoituron (approved):

PR #1132, opened 2023-12-15
Some FluentTab changes

vnbaaij:
Add a Visible property and control the id generation for the tab panel.

Also, some doc updates (not Tab related)


src/Core/Components/Tabs/FluentTab.razor.cs
```diff
@@ -28,6 +28,18 @@ public partial class FluentTab : FluentComponentBase
...
+    /// Gets or sets the visability of a tab
+    /// </summary>
+    [Parameter]
+    public bool Visible { get; set; } = true;
+
+    /// <summary>
+    /// Callback to invoke when visibility changes
+    /// </summary>
+    [Parameter]
+    public EventCallback<bool> VisibleChanged { get; set; }
```
dvoituron, line 41: Why to have a `VisibilityChanged`, if this event is never called by the component?

dvoituron:
Did you check the behavior of the component with properties likes `ActiveTabId`  `ActiveTab` and `TabsOverflow`?

src/Core/Components/Tabs/FluentTab.razor.cs
vnbaaij: I was proactive and thought we would probably need it somewhere? 😂

vnbaaij:
Yes, I did check. But lets hold off onmerging for a bit and let me  test some more things

src/Core/Components/Tabs/FluentTab.razor.cs
dvoituron: Mmmm ? No way 😀

dvoituron (approved):

Issue #1124, opened 2023-12-15
rfc: Review Icon sizing mechanics by decoupling Icons size from svg viewport sizing, exposing Icons without their size and dropping the enum restricting the available size. 


EricOnGit:
# 💬 RFC

Let's break it down while keeping in mind that I'm fairly new to the library. So let me know if I missed something.
1- decoupling icons size and svg viewport size. 
The basic idea is simply to make any svg icon resizable to any value. As long as the viewport is properly defined, you can set the width to any value and "voilà".
Looking a the ToMarkup function in the Icon class, there seems to have been some thoughts in that direction since the styleWidth variable might be initialized to a different value if the size parameter is not null. But, I can't see how to provide such value since the icons constructors are not receiving the size. (the base class is, tho)

2- Exposing Icons without their size
If we agree on # 1, then it would make sense to remove the size portion from the path to an Icon and have it as a parameter to the constructor. That would facilitate icon discovery and open up any size for all icons.
From Icons.[IconVariant].[IconSize].[IconName] to Icons.[IconVariant].[IconName]**([IconSize])**. I understand that we might need to keep the previous path for compatibility.

3- Drop the IconSize enum restricting the available size. 
Well at this point why bother restricting the size? Any positive value would work. The enum becomes a nuisance that will only infuriate anyone that requires a 38px button...

In the end an Icon could be exposed this way (minus the IconSize cast):
`public static partial class Icons {
	public static partial class Filled {
		public class PresenceAvailable(uint size) : Icon("PresenceAvailable", IconVariant.Filled, (IconSize)size, "<path d=\"M5 10A5 5 0 1 0 5 0a5 5 0 0 0 0 10Zm2.1-5.9L4.85 6.35a.5.5 0 0 1-.7 0l-1-1a.5.5 0 0 1 .7-.7l.65.64 1.9-1.9a.5.5 0 0 1 .7.71Z\"/>") { }
	}
}`

End used like this:
`<FluentButton IconStart="@(new Icons.Filled.PresenceAvailable(38))" />`

4- Another options is to implement the same behavior exposed through WithColor in the Icon class but this time for the size. Something along the lines of `public virtual Icon WithSize(string size)...` allowing size using other suffix than px.

Usage syntax might feels off if # 2 is not possible:
`<FluentButton IconStart="@(new Icons.Filled.Size10.PresenceAvailable().WithSize("38rem"))" />`

dvoituron:
[...]

PR #1136, opened 2023-12-15
Implemented Icon.WithSize(int? size) enabling Icons to be resized...

EricOnGit:
… to any dimensions.

# Pull Request

## 📖 Description
This is the quick and easy implementation that was discussed in issue #1124.
It offers a new WithSize() function that enables the dev to override the width of the Icon to any size.

This is compatible with previous version since the OverriddenWidth value is initialized to null and then tested against at render time so it's only used if the value was changed.

## 👩‍💻 Reviewer Notes
Comments were added directly in the Icon.cs file to better explain the change.

## 📑 Test Plan
There are currently no test on Icons. But I did test the component in my current project using both a FluentIcon and a FluentButton without any adverse effect.

### General

## ⏭ Next Steps
none to my knowledge


EricOnGit:

src/Core/Components/Icons/Icon.cs
```diff
@@ -87,15 +87,28 @@ internal Icon InverseColor(bool accentContainer)
...
     }
 
+    internal virtual int? OverriddenWidth { get; private set; } = null; // init to null so the previous behavior is unchanged
+
+    /// <summary>
+    /// Sets the render size of the icon in px. Use null to revert to icon original size
+    /// </summary>
+    /// <param name="size"></param>
+    /// <returns></returns>
+    public virtual Icon WithSize(int? size)
```
dvoituron, line 102: I think it would be possible to have a `string` argument: `WithSize(string ? size)` as in the **ToMarkup** method. This could be interesting for using `%` or other unit types. What do you think?

dvoituron:
Thanks. Very cool improvement :-)

We did not have time to put in all the unit tests at the start of the project, but we try to add them with each component modification. So I'm going to add a few unit tests to Dev. This will allow you to add a new one for this method :-)

dvoituron (changes requested):

src/Core/Components/Icons/Icon.cs
EricOnGit: Certainly, it would minimally require changes in the FluentIcon component and that exact ToMarkup method where "px" is hardcoded. It might also require changes in the way Width and Size are delt with (mostly Width which is a protected in an unsealed class). And that's where I decided to play it safe due to my limited knowledge of the ripple effects. 

That said, if you're ok with pushing this a little further, I can spend some more time on this.
I would also be a bit more comfortable if we had some tests coverage here. Looks like we'll have to synch our changes to move forward😊

dvoituron: I added a PR #1136 to include Unit Tests

EricOnGit: I think you meant #1140

vnbaaij: Why not have both `WithSize(int? size)` and `WithSize(string? size)`?

And why are these parameters nullable? What would be the outcome of sending in a null value? I think we may expect people to supply a valid input value.

dvoituron: We could have both, with `WithSize(int? size) => WithSize(size is null ? null : $"{size}px");` 
The advantage of having a null value is to "reset" another specified size.

EricOnGit:
@dvoituron @vnbaaij I've added a new commit this morning that includes the changes to the generator so you can have a look at this code too.

vnbaaij:
@EricOnGit, please revert that last commit and create a new PR for it. I'm fairly certain we will mergi in the `WithSize` work but not certain yet of the rest. Having it in a seperate PR allows us to evealuate things easier.

vnbaaij:
Hold off a bit on creating the new PR please

vnbaaij:
Ok, let's add the sting variant and get this merged in.

EricOnGit:
> Ok, let's add the sting variant and get this merged in.

@vnbaaij there is a catch to bringing the string variant. 

The Width property of the Icon class is exposes as "protected internal". I'll admit I had to look it up to be sure. It turns out those modifiers are acting as OR. Old reference [here](https://stackoverflow.com/questions/585859/what-is-the-difference-between-protected-and-protected-internal) 

Which means that if we implement WithSize(string?) we will either have a messy meaning for Width (only the int part or something like "1.3rem"?) or have to expose Width as string which would break any derived class 😣

I thought of it for a moment and I see a couple options (none are perfect tho)

1. Introduce a Unit property that would support things like: rem, em, %. It feels like a can of worms at the bottom of a rabbit hole to me but it's an option nonetheless.

2. Introduce something like "bool UseParentWidth" or "bool FullWidth" to only support the use case of 100%. I think this is the most likely reason not to stick with pixels.

It's probably inevitable that this will have to break at some point but is this PR the right one? I rarely advise for caution but you mentioned earlier that you rather not introduce a breaking change.

PR #1155, opened 2023-12-19
Use GlobalState, use LocalizationDirection, less var's

vnbaaij:
As discussed

### DIFF
--- a/examples/Demo/Shared/Components/SiteSettingsPanel.razor
+++ b/examples/Demo/Shared/Components/SiteSettingsPanel.razor
@@ -3,7 +3,7 @@
 <div>
     <FluentDesignTheme @bind-Mode="@Mode"
                        @bind-OfficeColor="@OfficeColor"
-                       Direction="@(Direction ? "ltr" : "rtl")"
+                       Direction="@(Direction ? LocalizationDirection.ltr : LocalizationDirection.rtl)"
                        StorageName="theme" />
 
     <FluentStack Orientation="Orientation.Vertical" VerticalGap="0">
--- a/src/Core/Components/DesignSystemProvider/FluentDesignTheme.razor.cs
+++ b/src/Core/Components/DesignSystemProvider/FluentDesignTheme.razor.cs
@@ -8,11 +8,14 @@ public partial class FluentDesignTheme : ComponentBase
 {
     private const string JAVASCRIPT_FILE = "./_content/Microsoft.FluentUI.AspNetCore.Components/Components/DesignSystemProvider/FluentDesignTheme.razor.js";
     private DotNetObjectReference<FluentDesignTheme>? _dotNetHelper = null;
-    private string? _direction;
+    private LocalizationDirection _direction;
     private readonly JsonSerializerOptions JSON_OPTIONS = new JsonSerializerOptions
     {
         PropertyNameCaseInsensitive = true,
     };
+
+    [Inject]
+    private GlobalState GlobalState { get; set; } = default!;
 
     /// <summary />
     [Inject]
@@ -73,13 +76,15 @@ public partial class FluentDesignTheme : ComponentBase
     /// Gets or sets the body.dir value.
     /// </summary> 
     [Parameter]
-    public string? Direction
+    public LocalizationDirection Direction
     {
         get => _direction;
         set
         {
             _direction = value;
-            Module?.InvokeVoidAsync("UpdateDirection", value);
+           GlobalState.SetDirection(_direction);
+           
+            Module?.InvokeVoidAsync("UpdateDirection", value.ToAttributeValue());
         }
     }
 
@@ -119,6 +124,7 @@ public partial class FluentDesignTheme : ComponentBase
 
                 if (OnLuminanceChanged.HasDelegate)
                 {
+                    GlobalState.SetLuminance(value.Contains("dark") ? StandardLuminance.DarkMode : StandardLuminance.LightMode);
                     await OnLuminanceChanged.InvokeAsync(new LuminanceChangedEventArgs(mode, value.Contains("dark")));
                 }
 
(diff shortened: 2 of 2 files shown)


examples/Demo/Shared/Components/SiteSettingsPanel.razor
```diff
@@ -3,7 +3,7 @@
 <div>
     <FluentDesignTheme @bind-Mode="@Mode"
                        @bind-OfficeColor="@OfficeColor"
-                       Direction="@(Direction ? "ltr" : "rtl")"
+                       Direction="@(Direction ? LocalizationDirection.ltr : LocalizationDirection.rtl)"
```
dvoituron, line 6: Following HTML doc, Direction can be `ltr`, `rtl` or `auto` (and maybe empty?)
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir

And I think we need to refactor the enum using Pascal case.
```csharp
public enum LocalizationDirection
{
    Ltr,
    Rtl,
    Auto
}
```

src/Core/Components/DesignSystemProvider/FluentDesignTheme.razor.cs
```diff
@@ -156,13 +162,14 @@ protected async override Task OnAfterRenderAsync(bool firstRender)
             Module ??= await JSRuntime.InvokeAsync<IJSObjectReference>("import", JAVASCRIPT_FILE);
             _dotNetHelper = DotNetObjectReference.Create(this);
 
-            _direction = await Module.InvokeAsync<string?>("GetDirection");
+            if (!Enum.TryParse(await Module.InvokeAsync<string>("GetDirection"), true, out _direction))
```
dvoituron, line 165: Could you add `{ ... }` (naming conventions)

src/Core/Components/DesignSystemProvider/FluentDesignTheme.razor.cs
```diff
@@ -156,13 +162,14 @@ protected async override Task OnAfterRenderAsync(bool firstRender)
...
+                _direction = LocalizationDirection.ltr;
 
-            var themeJSON = await Module.InvokeAsync<string>("addThemeChangeEvent", _dotNetHelper, Id);
-            var theme = themeJSON == null ? null : JsonSerializer.Deserialize<DataLocalStorage>(themeJSON, JSON_OPTIONS);
+            string? themeJSON = await Module.InvokeAsync<string>("addThemeChangeEvent", _dotNetHelper, Id);
```
dvoituron, line 168: A reason why you changed `var` to `string?` and `DataLocalStorage?` ?

src/Core/Components/DesignSystemProvider/FluentDesignTheme.razor.cs
```diff
@@ -171,7 +178,7 @@ private async Task ApplyLocalStorageValues(DataLocalStorage? theme)
         // Mode (Dark / Light / System)
         if (!string.IsNullOrEmpty(theme?.Mode))
         {
-            if (!Enum.TryParse<DesignThemeModes>(theme.Mode, true, out var mode))
+            if (!Enum.TryParse(theme.Mode, true, out DesignThemeModes mode))
```
dvoituron, line 181: Same... why did you remove `var`? :-)

dvoituron (changes requested):

examples/Demo/Shared/Components/SiteSettingsPanel.razor
vnbaaij: Agree that is is part of the standard but we do not use 'auto' in any of our components (atm). How would we handle something like if 'auto' would be a valid option? 
```
    .AddStyle("left", $"{HorizontalPosition!.Value.ToString(CultureInfo.InvariantCulture)}%", () => HorizontalPosition.HasValue && GlobalState.Dir == LocalizationDirection.Ltr)
    .AddStyle("right", $"{HorizontalPosition!.Value.ToString(CultureInfo.InvariantCulture)}%", () => HorizontalPosition.HasValue && GlobalState.Dir == LocalizationDirection.Rtl)
```

I'll do the Pascal refactoring

src/Core/Components/DesignSystemProvider/FluentDesignTheme.razor.cs
```diff
@@ -73,13 +76,15 @@ public partial class FluentDesignTheme : ComponentBase
...
+    public LocalizationDirection Direction
     {
         get => _direction;
         set
         {
             _direction = value;
-            Module?.InvokeVoidAsync("UpdateDirection", value);
+           GlobalState.SetDirection(_direction);
+           
+            Module?.InvokeVoidAsync("UpdateDirection", value.ToAttributeValue());
```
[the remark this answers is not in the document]
vnbaaij: We do not have 'auto' as an available option to be sent here (with the siwitch in settings , for example, we choose either LTR or RTL.

src/Core/Components/DesignSystemProvider/FluentDesignTheme.razor.cs
```diff
@@ -8,12 +8,15 @@ public partial class FluentDesignTheme : ComponentBase
...
     private DotNetObjectReference<FluentDesignTheme>? _dotNetHelper = null;
-    private string? _direction;
+    private LocalizationDirection _direction;
     private readonly JsonSerializerOptions JSON_OPTIONS = new JsonSerializerOptions
     {
         PropertyNameCaseInsensitive = true,
     };
 
+    [Inject]
+    private GlobalState GlobalState { get; set; } = default!;
```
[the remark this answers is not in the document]
vnbaaij: I think renaming is a good thing to do but it would be a breaking change so not appropriate now. 
I like `GlobalDesign`! Let's reserve that one for when we do want to do the rename. I'll update the source documentation that this is something we are going to do in the next major version.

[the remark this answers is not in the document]
dvoituron: You could create a `GlobalDesign` class and set `Obsolete` all usages of `GlobalState` (but keeping the property). A lot of references?

examples/Demo/Shared/Components/SiteSettingsPanel.razor
dvoituron: I think `Auto` should be present in the Enum and that you should apply a "default value" for some components that need LTR or RTL. But you can't dictate the direction of the entire website.

src/Core/Components/DesignSystemProvider/FluentDesignTheme.razor.cs
vnbaaij: I don't like `var` :).
In VS it displays like this (for me based on my settings, only changed it for the first line):
image

If the type is already known an shown, then why not us it? VS actually does the substitution for you if you double check on the type.

src/Core/Components/DesignSystemProvider/FluentDesignTheme.razor.cs
vnbaaij: Same..

src/Core/Components/DesignSystemProvider/FluentDesignTheme.razor.cs
dvoituron: It shouldn't depend on preferences (like or dislike), but on naming convention 😀
Here's the one we use in many MS projects: https://dvoituron.com/naming-conventions/rules/

image

The code is clearer / aligned with `var` and it's more " scalable " when the objects (right) evolve.
But it's only a naming convention.

```csharp
var themeJSON = ...
var theme = ...
```

src/Core/Components/DesignSystemProvider/FluentDesignTheme.razor.cs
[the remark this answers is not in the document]
vnbaaij: As discussed. `UpdateDirection` can write a `null` value

examples/Demo/Shared/Components/SiteSettingsPanel.razor
vnbaaij: Added new LeftToRight an RightToLeft values and obsoleted th 'old' values

dvoituron (approved):

PR #1189, opened 2023-12-22
fix Combobox z-index stack order issue

### LINKED ISSUE #1188: fix: Combobox wrong z-index stack order
When the `Height` property is not specify in `FluentCombobox`, some elements are displayed on top of the combobox options list.

image

image

Basically the guitly is this logic:

`@if (!String.IsNullOrEmpty(Height))
{
    <style>
        @($"#{Id}::part(listbox) {{ max-height: {Height}; z-index: {ZIndex.ComboboxPopup} }}")
    </style>
}`

I can make a PR, I just wanted to know if this is intended.

dromerolovo:
This change resolves #1188

### DIFF
--- a/src/Core/Components/List/FluentCombobox.razor
+++ b/src/Core/Components/List/FluentCombobox.razor
@@ -3,7 +3,7 @@
 @typeparam TOption
 
 <style>
-    @($"#{Id}::part(listbox) {{ {(string.IsNullOrEmpty(Height) ? null : string.Concat("max-height:", Height))}; z-index: {ZIndex.ComboboxPopup} }}")
+    @($"#{Id}::part(listbox) {{ {(string.IsNullOrEmpty(Height) ? null : $"max-height: {Height};")} z-index: {ZIndex.ComboboxPopup} }}")
 </style>
 
 <CascadingValue Value=@_internalListContext Name="ListContext" TValue="InternalListContext<TOption>" IsFixed=true>
--- a/src/Core/Components/List/FluentCombobox.razor
+++ b/src/Core/Components/List/FluentCombobox.razor
@@ -2,12 +2,9 @@
 @inherits ListComponentBase<TOption>
 @typeparam TOption
 
-@if (!String.IsNullOrEmpty(Height))
-{
-    <style>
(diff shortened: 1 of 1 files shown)


vnbaaij (approved):

src/Core/Components/List/FluentCombobox.razor
```diff
@@ -2,12 +2,11 @@
...
 
-@if (!String.IsNullOrEmpty(Height))
-{
-    <style>
-        @($"#{Id}::part(listbox) {{ max-height: {Height}; z-index: {ZIndex.ComboboxPopup} }}")
-    </style>
-}
+
+<style>
+    @($"#{Id}::part(listbox) {{ max-height: {(string.IsNullOrEmpty(Height) ? null : Height)}; z-index: {ZIndex.ComboboxPopup} }}")
```
dvoituron, line 7: Could you modify the generation of this "string", because when "height" is not defined, the style still contains a "height" attribute.

Example
```html
<style>#combobox::part(listbox) { max-height: ; z-index: 9995 }</style>
```

dvoituron (changes requested):

src/Core/Components/List/FluentCombobox.razor
dromerolovo: Sure. I'll amend the changes

dromerolovo: Before:
Screenshot 2023-12-22 093152

After:
Screenshot 2023-12-22 093405

dvoituron: I'm going to be annoying but it's my job 😎. Can you add a space after the max-height? To be consistent with the rest of the generated code. 

And it's probably clearer to do ´$"max-height: {Height};"´.

vnbaaij:
@dvoituron I knew you were going to say that! 😂
I was like "we'll do that after merging"

src/Core/Components/List/FluentCombobox.razor
dromerolovo: No prob at all

Issue #280, opened 2023-01-01
fix: FluentMenu broken in published release build


juho-hanhimaki:
# 🐛 Bug Report

I have a basic FluentMenu.

<img width="346" alt="image" src="https://user-images.githubusercontent.com/12390182/210171194-6e79c0ff-b690-4fe0-960c-7692f88c7221.png">

On local development build everything works great. On a published release build clicking the FluentMenuItem causes an exception. There are no other warnings or exceptions visible in dev tools.

```
blazor.webassembly.js:1 
        
       Uncaught (in promise) Error: System.InvalidOperationException: There was an error parsing the event arguments. EventId: '8'.
 ---> System.NotSupportedException: DeserializeNoConstructor, JsonConstructorAttribute, Microsoft.Fast.Components.FluentUI.MenuChangeEventArgs Path: $ | LineNumber: 0 | BytePositionInLine: 1.
 ---> System.NotSupportedException: DeserializeNoConstructor, JsonConstructorAttribute, Microsoft.Fast.Components.FluentUI.MenuChangeEventArgs
   Exception_EndOfInnerExceptionStack
   at System.Text.Json.ThrowHelper.ThrowNotSupportedException(ReadStack& , Utf8JsonReader& , NotSupportedException )
   at System.Text.Json.ThrowHelper.ThrowNotSupportedException_DeserializeNoConstructor(Type , Utf8JsonReader& , ReadStack& )
   at System.Text.Json.Serialization.Converters.ObjectDefaultConverter`1[[Microsoft.Fast.Components.FluentUI.MenuChangeEventArgs, Microsoft.Fast.Components.FluentUI, Version=2.0.0.0, Culture=neutral, PublicKeyToken=null]].OnTryRead(Utf8JsonReader& , Type , JsonSerializerOptions , ReadStack& , MenuChangeEventArgs& )
   at System.Text.Json.Serialization.JsonConverter`1[[Microsoft.Fast.Components.FluentUI.MenuChangeEventArgs, Microsoft.Fast.Components.FluentUI, Version=2.0.0.0, Culture=neutral, PublicKeyToken=null]].TryRead(Utf8JsonReader& , Type , JsonSerializerOptions , ReadStack& , MenuChangeEventArgs& )
[...]

PR #1223, opened 2023-12-27
[DialogService] Add a fake instance of DialogEventArgs #280

dvoituron:
# [DialogService] Add a fake instance of DialogEventArgs #280

Add this line to prevent the _Trimming_ process from deleting this class?

```csharp
_ = new DialogEventArgs() { Id = "0", Reason = "Unknown" };
```

### DIFF
--- a/src/Core/Components/Dialog/Services/DialogService.cs
+++ b/src/Core/Components/Dialog/Services/DialogService.cs
@@ -1,14 +1,14 @@
-﻿using Microsoft.AspNetCore.Components;
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.AspNetCore.Components;
 
 namespace Microsoft.FluentUI.AspNetCore.Components;
 
 public partial class DialogService : IDialogService
 {
     /// <summary />
+    [DynamicDependency(DynamicallyAccessedMemberTypes.All, typeof(DialogEventArgs))]
     public DialogService()
     {
-        // Issue #280: to avoid the Trimming process to remove this class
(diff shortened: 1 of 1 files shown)


vnbaaij:
Should we not use a `DynamicDependencyAttribute` for this:
[DynamicDependency(DynamicallyAccessedMemberTypes.All, typeof(DialogEventArgs))]

See recommendations under https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming?pivots=dotnet-8-0

dvoituron:
> Should we not use a `DynamicDependencyAttribute` for this: [DynamicDependency(DynamicallyAccessedMemberTypes.All, typeof(DialogEventArgs))]
> 
> See recommendations under https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming?pivots=dotnet-8-0

I don't think so, but maybe I'm wrong. In my opinion, this feature is useful when calling a method that uses reflection. It adds a warning at compile time to inform the developer. Here, we're not using reflection with this class. However, adding this tiny class to the lib should correct this problem.

dvoituron:
@vnbaaij  You're probably right, this code is probably cleaner. I'll update it

```csharp
[DynamicDependency(DynamicallyAccessedMemberTypes.All, typeof(DialogEventArgs))]
public partial class DialogService : IDialogService
{
}
```

vnbaaij (approved):

PR #1242, opened 2023-12-31
Implemented OnDialogClosing callback

au5ton:
# Pull Request

### This change focuses on adding the `OnDialogClosing` DialogParameter, which enables consumers to implement their own "exit" animations for all FluentDialogs

In the past, the team has stated that implementing animations isn't a priority.

- https://github.com/microsoft/fluentui-blazor/discussions/330
- https://github.com/microsoft/fluentui-blazor/discussions/580

This is acceptable, however, this leaves implementing **entrance/exit animations for dialogs** up to the consumer of the library.

### Exit animation (pull request needed)

This pull request is necessary in order to perform this because there are many different ways that a FluentDialog may be closed, and there's currently no way to interrupt the lifecycle in order to play an animation before actually closing and removing the reference.

<details>
<summary>(Outdated example, don't use)</summary>

_app.css_
```css
fluent-dialog[class~="left"].transitioning::part(control) {
    transition: all linear 3s;
}
```

_Counter.razor_
```razor
@inject IJSRuntime JS

<FluentButton Appearance="Appearance.Accent" @onclick="OpenPanel">Open panel</FluentButton>

@code {
    private async Task OpenPanel()
    {
        var dialog = await DialogService.ShowPanelAsync<SimplePanel>(new()
            {
                Title = "Panel Title",
                Alignment = HorizontalAlignment.Left,
                PrimaryActionEnabled = false,
                SecondaryActionEnabled = false,
                Width = "300px",
                OnDialogClosing = EventCallback.Factory.Create(this, async () =>
                {
                    // Don't use eval, this is just as a proof-of-concept
                    // Here we show that a consumer could use this to animate dialogs
                    await JS.InvokeVoidAsync("eval", $@"
                    var flDial = document.getElementById('{dialog.Id}');
                    // dialog will be visually hidden immediately after being opened
                    flDial.classList.add('transitioning');
                    flDial.dialog.style.opacity = '0.0';
                    flDial.dialog.style.transform = 'translateX(-100%)';
                    ");
                    await Task.Delay(3000); // our transition time is defined as 3s in CSS   
                })
            });
        
    }
}
```

</details>

### Entrance animation (pull request not needed)

Without this change, implementing an **entrance** animation for a dialog could work as follows:

<details>
<summary>(Outdated example, don't use)</summary>

_app.css_
```css
fluent-dialog[class~="left"].transitioning::part(control) {
    transition: all linear 3s;
}
```

_Counter.razor_
```razor
@inject IJSRuntime JS

<FluentButton Appearance="Appearance.Accent" @onclick="OpenPanel">Open panel</FluentButton>

@code {
    private async Task OpenPanel()
    {
        var dialog = await DialogService.ShowPanelAsync<SimplePanel>(new()
            {
                Title = "Panel Title",
                Alignment = HorizontalAlignment.Left,
                PrimaryActionEnabled = false,
                SecondaryActionEnabled = false,
                Width = "300px",
            });
        // Don't use eval, this is just as a proof-of-concept
        await JS.InvokeVoidAsync("eval", $@"
        var flDial = document.getElementById('{dialog.Id}');
        // dialog will be visually hidden immediately after being opened
        flDial.dialog.style.opacity = '0.0';
        flDial.dialog.style.transform = 'translateX(-100%)';
        // necessary in order for animations to happen separately from the changes we just made
        setTimeout(() => {{
            flDial.classList.add('transitioning');
            // visually reset the dialog to its normal position and style
            flDial.dialog.style.opacity = '';
            flDial.dialog.style.transform = '';
        }}, 0);
        ");
    }
}
```

</details>

### General

### Component-specific

### DIFF
--- a/examples/Demo/Shared/Pages/Dialog/Examples/DialogServiceCallbackExample.razor
+++ b/examples/Demo/Shared/Pages/Dialog/Examples/DialogServiceCallbackExample.razor
@@ -42,6 +42,10 @@
             {
                 Title = $"Hello {simplePerson.Firstname}",
                 OnDialogResult = DialogService.CreateDialogCallback(this, HandleDialog),
+                // OnDialogClosing = EventCallback.Factory.Create(this, async () =>
+                // {
+                //     await Task.Delay(1000);    
+                // }),
                 PrimaryAction = "Yes",
                 PrimaryActionEnabled = false,
                 SecondaryAction = "No",
--- a/examples/Demo/Shared/Pages/Dialog/Examples/DialogServiceCallbackExample.razor
+++ b/examples/Demo/Shared/Pages/Dialog/Examples/DialogServiceCallbackExample.razor
@@ -42,10 +42,6 @@
             {
                 Title = $"Hello {simplePerson.Firstname}",
                 OnDialogResult = DialogService.CreateDialogCallback(this, HandleDialog),
-                // OnDialogClosing = EventCallback.Factory.Create(this, async () =>
-                // {
-                //     await Task.Delay(1000);    
-                // }),
                 PrimaryAction = "Yes",
                 PrimaryActionEnabled = false,
                 SecondaryAction = "No",
--- a/src/Core/Components/Dialog/FluentDialog.razor
+++ b/src/Core/Components/Dialog/FluentDialog.razor
@@ -13,7 +13,6 @@
                    aria-describedby=@(_parameters.AriaDescribedby ?? AriaDescribedby)
                    aria-labelledby=@(_parameters.AriaLabelledby ?? AriaLabelledby)
                    aria-label=@(_parameters.AriaLabel ?? AriaLabel)
-                   @ondialogdismiss=HandleDismissAsync
                    @attributes=@AdditionalAttributes>
 
         @* Default Header *@
--- a/src/Core/Components/Dialog/FluentDialog.razor
+++ b/src/Core/Components/Dialog/FluentDialog.razor
@@ -13,6 +13,7 @@
                    aria-describedby=@(_parameters.AriaDescribedby ?? AriaDescribedby)
                    aria-labelledby=@(_parameters.AriaLabelledby ?? AriaLabelledby)
                    aria-label=@(_parameters.AriaLabel ?? AriaLabel)
+                   @ondialogdismiss=HandleDismissAsync
                    @attributes=@AdditionalAttributes>
 
         @* Default Header *@
--- a/src/Core/Components/Dialog/FluentDialog.razor.cs
+++ b/src/Core/Components/Dialog/FluentDialog.razor.cs
@@ -150,6 +150,14 @@ public partial class FluentDialog : FluentComponentBase
         if (firstRender)
         {
             await Element.FocusAsync();
+
+            if (Instance is not null)
+            {
+                if (Instance.Parameters.OnDialogOpened.HasDelegate)
+                {
+                    await Instance.Parameters.OnDialogOpened.InvokeAsync(Instance);
+                }
+            }
         }
     }
 
@@ -226,7 +234,7 @@ public partial class FluentDialog : FluentComponentBase
         {
             if (Instance.Parameters.OnDialogClosing.HasDelegate)
             {
-                await Instance.Parameters.OnDialogClosing.InvokeAsync(dialogResult);
+                await Instance.Parameters.OnDialogClosing.InvokeAsync(Instance);
             }
         }
         DialogContext?.DialogContainer.DismissInstance(Id!, dialogResult);
--- a/src/Core/Components/Dialog/FluentDialog.razor.cs
+++ b/src/Core/Components/Dialog/FluentDialog.razor.cs
@@ -222,6 +222,13 @@ public partial class FluentDialog : FluentComponentBase
     /// </summary>
     public async Task CloseAsync(DialogResult dialogResult)
     {
+        if (Instance is not null)
+        {
+            if (Instance.Parameters.OnDialogClosing.HasDelegate)
+            {
+                await Instance.Parameters.OnDialogClosing.InvokeAsync(dialogResult);
+            }
+        }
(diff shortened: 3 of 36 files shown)


au5ton:

au5ton:
The goal of this pull request is to make it possible for consumers to implement "exit" animations for FluentDialogs. However, it could be very inconvenient to have to have to do this multiple times for every dialog called.

This may be better suited to be a new pair of EventCallbacks on `FluentDialogProvider` so that the animations can be provided throughout the entire library:
- `[Parameter] public EventCallback<IDialogReference> OnDialogOpened { get; set; }`
- `[Parameter] public EventCallback<IDialogReference> OnDialogClosing { get; set; }`

Feedback is welcomed. I am happy to make whatever changes may be necessary to get something accepted.

examples/Demo/Shared/Pages/Dialog/Examples/DialogServiceCallbackExample.razor
```diff
@@ -42,6 +42,10 @@
             {
                 Title = $"Hello {simplePerson.Firstname}",
                 OnDialogResult = DialogService.CreateDialogCallback(this, HandleDialog),
+                // OnDialogClosing = EventCallback.Factory.Create(this, async () =>
+                // {
+                //     await Task.Delay(1000);    
+                // }),
```
vnbaaij, line 48: Either include the code or remove it. Don't use commented out code.

vnbaaij (changes requested):
Let's also try to include the OnDialogOpening event callback

examples/Demo/Shared/Pages/Dialog/Examples/DialogServiceCallbackExample.razor
au5ton: Thank you, I addressed this in the latest commit 201fa727268fee72a92202cec2fc4eaa0d9393b2

au5ton:
> Let's also try to include the OnDialogOpening event callback

I added an "OnDialogOpened" event callback that executes after the FluentDialog is rendered for the first time. I also made both callbacks pass the `DialogInstance` so that the callback can access the ID and the DialogParameters, which are useful for getting the element references in JavaScript and changing the animation depending on the type of dialog (`DialogInstance.Parameters.DialogType`).

With it, I added example code to the "Dialog" page in `DialogServiceAnimationCallbackExample.razor`

In the example, the [Web Animations API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Animations_API) is used so the animations are safely interruptible and don't require CSS.

src/Core/Components/Dialog/FluentDialogProvider.razor
```diff
@@ -14,7 +14,7 @@
     {
         @foreach (DialogReference dialog in _internalDialogContext.References)
         {
-            <FluentDialog @key="@dialog" Id="@dialog.Id" Instance="@dialog.Instance" Data="@dialog.Instance.Content" @ondialogdismiss=OnDismissAsync />
+            <FluentDialog @key="@dialog" Id="@dialog.Id" Instance="@dialog.Instance" Data="@dialog.Instance.Content" />
```
vnbaaij, line 17: Is this change necessary for this PR spcifically?

src/Core/Components/Dialog/FluentDialog.razor.cs
```diff
@@ -236,6 +251,17 @@ public async Task CloseAsync(DialogResult dialogResult)
...
+    {
+        if (args is not null && args.Reason is not null && args.Reason == "dismiss" && !string.IsNullOrWhiteSpace(args.Id) && args.Id == this.Id)
+        {
+            if (this.Instance.Parameters.PreventDismissOnOverlayClick == false)
+            {
+                await CloseAsync(DialogResult.Cancel());
+            }
+        }
+    }
+
```
vnbaaij, line 264: Is this change necessary for this PR spcifically?

src/Core/Components/Dialog/FluentDialog.razor
```diff
@@ -13,6 +13,7 @@
                    aria-describedby=@(_parameters.AriaDescribedby ?? AriaDescribedby)
                    aria-labelledby=@(_parameters.AriaLabelledby ?? AriaLabelledby)
                    aria-label=@(_parameters.AriaLabel ?? AriaLabel)
+                   @ondialogdismiss=HandleDismissAsync
```
vnbaaij, line 16: Is this change necessary for this PR spcifically?

vnbaaij (changes requested):
I'm wondering if the three changes I commented on a needed to get the event callbacks added? Or are you trying to fix something else with that? If it is not necessary for this PR than we should create a new PR for those sepecific changes as it changes existing code (could potentially be a breaking change)

src/Core/Components/Dialog/FluentDialog.razor
au5ton: The original intention behind the change was because dismissing a dialog doesn't go down the same path as when it is closed by clicking the "X" or by clicking the Secondary action. By putting the dismiss under a similar path, checking for the "OnDialogClosing" callback would only need to exist in one place.

After looking into this further, any `@ondialogdismiss` added to `<FluentDialog />` is overwrites the one listed here because of the `@attributes=@AdditionalAttributes` bind. I will move it back in a follow up commit.

au5ton: I've addressed this in 41e1c1cd69a9993652e15e23241bcdfe08fb534d

au5ton:
After doing some further testing, the changes you pointed out were causing issues so I reverted them. As part of that, `OnDialogClosing` is now also invoked from FluentDialogProvider as it handles dismisses.

vnbaaij:
Code is looking good! Thanks and appreciated. I am going to checkout the PR (hopefully tomorrow) and do some final tests. If everything works, I'll approve (and merge) after that.

vnbaaij (approved):

PR #1267, opened 2024-01-05
Add Id parameters

vnbaaij:
- Card: Add Id parameter + tests 
- Stack: Add Id parameter + tests

### DIFF
--- a/examples/Demo/Shared/Components/SiteSettingsPanel.razor.cs
+++ b/examples/Demo/Shared/Components/SiteSettingsPanel.razor.cs
@@ -9,6 +9,7 @@ public partial class SiteSettingsPanel
     private string? _status;
     private bool _popVisible;
     private bool _ltr = true;
+    private FluentDesignTheme? _theme;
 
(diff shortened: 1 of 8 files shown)


examples/Demo/Shared/Components/SiteSettingsPanel.razor.cs
```diff
@@ -9,6 +9,7 @@ public partial class SiteSettingsPanel
     private string? _status;
     private bool _popVisible;
     private bool _ltr = true;
+    private FluentDesignTheme? _theme;
```
dvoituron, line 12: I don't know why you added a Ref, but ok.

dvoituron (approved):

examples/Demo/Shared/Components/SiteSettingsPanel.razor.cs
vnbaaij: It is to call the new reset functionality (delete cache +local storage)

PR #1288, opened 2024-01-09
Prevent loop when working with random color

### LINKED ISSUE #1282: fix: Looping behaviour after update to 4.3.0 / FluentDesignTheme
I've updated to 4.3.0 and I'm having a strange behaviour now in my settings panel where the user can set the theme via the `FluentDesignTheme` component.

It looks like the color select list is populated in an endless loop:

A sample repo is available here: https://github.com/woha/theme-demo

vnbaaij:
Remove random color selection from site settings
Add button to select random color on theme page
Fix #1282


src/Core/Utilities/OfficeColorUtilities.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Microsoft.FluentUI.AspNetCore.Components;
+
+public static class OfficeColorUtilities
+{
+    public static OfficeColor GetRandom(bool skipDefault = true)
```
dvoituron, line 5: To avoid computing all the Colors each time, you could have a code like this:
```csharp
public static readonly DesignThemeModes[] AllThemeModes = Enum.GetValues<DesignThemeModes>();

public static readonly OfficeColor[] AllColors = Enum.GetValues<OfficeColor>();

public static OfficeColor GetRandom(bool skipDefault = true)
{
    var values = AllColors.Skip(skipDefault ? 1 : 0);
    return values.ElementAt(new Random().Next(values.Count()));
}
```

examples/Demo/Shared/Pages/Design/Examples/DesignThemeDefault.razor
```diff
@@ -9,14 +8,16 @@
...
             <FluentSelect Label="Theme"
+                          Width="250px"
                           Items="@(Enum.GetValues<DesignThemeModes>())"
                           @bind-SelectedOption="@Mode" />
         </FluentGridItem>
 
         <FluentGridItem>
             <FluentSelect Label="Color"
-                          Items="@(Enum.GetValues<OfficeColor>())"
+                          Items="@(Enum.GetValues<OfficeColor>().Select(i => (OfficeColor?)i))"
```
dvoituron, line 18: (Optional) With my other comment, you could replace by:

```csharp
Items="@(OfficeColorUtilities.AllColors.Cast<OfficeColor?>())"
```

c0nd3v:
Please merge this already and issue a hotfix asap...

dvoituron (approved):

PR #1307, opened 2024-01-11
Fix #1305 by overriding SetParametersAsync for Combobox

### LINKED ISSUE #1305: fix: ComboBox component resetting bound object to null when custom content is entered
# 🐛 Bug Report

## 💻 Repro or Code Sample
- Use a ComboBox component with an IEnumerable<string> as the data source.
- Select an existing value from the data source (works as expected).
- Enter custom content in the ComboBox (UI displays correctly), notice that the bound object with using @bind-Value in the backend is reset to null.

## 🤔 Expected Behavior
The bound object should retain the custom content entered in the ComboBox when @bind-Value is used.

## 😯 Current Behavior
When custom content is entered in the ComboBox, the bound object with @bind-Value is reset to null .

## 💁 Possible Solution
The ComboBox component may need to be updated to handle custom content correctly.

## 🔦 Context
This issue affects the functionality of the ComboBox component when dealing with custom content and data binding.

## 🌍 Your Environment
* OS & Device: [e.g. Windows 11, PC]
* Browser [e.g. Edge ]
* .NET and FAST Version [e.g. .NET 8]

vnbaaij:
In v4.3 we merged in a change to `SetParametersAsync` in `ListComponentBase` to sync `Selectedoption` with `Value`. This interferes with how a Combobox works. By adding its own overridden `SetParametersAsync` to `FluentComnbobox` this works again as it should.

### DIFF
--- a/src/Core/Components/List/FluentCombobox.razor.cs
+++ b/src/Core/Components/List/FluentCombobox.razor.cs
@@ -41,6 +41,13 @@ public partial class FluentCombobox<TOption> : ListComponentBase<TOption> where 
         .AddStyle("min-width", Width, when: !string.IsNullOrEmpty(Width))
         .Build();
 
+    public override async Task SetParametersAsync(ParameterView parameters)
+    {
+        parameters.SetParameterProperties(this);
+
+        await base.SetParametersAsync(ParameterView.Empty);
+    }
+
     protected override async Task OnChangedHandlerAsync(ChangeEventArgs e)
     {
         if (e.Value is not null && Items is not null)


src/Core/Components/List/FluentCombobox.razor.cs
```diff
@@ -41,6 +41,13 @@ public partial class FluentCombobox<TOption> : ListComponentBase<TOption> where
         .AddStyle("min-width", Width, when: !string.IsNullOrEmpty(Width))
         .Build();
 
+    public override async Task SetParametersAsync(ParameterView parameters)
```
dvoituron, line 44: I think you could use this code to avoid to use "async/await" in this case.

```csharp
public override Task SetParametersAsync(ParameterView parameters)
{
     parameters.SetParameterProperties(this);
     return base.SetParametersAsync(ParameterView.Empty);
}
```

dvoituron (approved):

PR #1313, opened 2024-01-12
[FluentPanelSplitter] New Splitter component (Experimental)

dvoituron:
# Panel Splitter (Experimental)
> ⚠️ This component is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed. To use it, you need to add the following line to the top of your razor file `@{ #pragma warning disable FluentPanelSplitter }`

The **Panel Splitter** splits the page into sections and allows the user to control the page layout.

You can include as many **Pane** as you like in a container. By default, panels are resizable, but the `Resizable` property can be used to block this functionality. The initial `Size` is in percent or pixels and has a `Min` and `Max` value. A Panel can be fully collapsed by setting the `Collapsible` property to True.

These CSS variables are predefined with these values, but can be overwritten (using Style property

```css
    .fluent-panel-splitter {
        /* Resize Icon colors */
        --fluent-panel-splitter-color: var(--neutral-stroke-strong-rest);
        --fluent-panel-splitter-color-active: var(--neutral-stroke-strong-hover);

        /* Bar colors */
        --fluent-panel-splitter-background-color: var(--neutral-stroke-rest);
        --fluent-panel-splitter-background-color-active: var(--neutral-stroke-hover);
        --fluent-panel-splitter-hover-opacity: 0.8;

        /* Bar size */
        --fluent-panel-splitter-bar-size: 8px;
    }
```

> Why have we created this new component when FluentSplitter already exists?
> We want to generalize this component by making it easier to have several (>2) panels per container. We've also used CSS variables to simplify the customization of colors and sizes.
> Give us your feedback

## Example

PanelSplitter

```razor
<FluentPanelSplitter Height="150px">
    <FluentPanelSplitterPane Size="20%" Min="50px" Max="70%">
        Left Menu
    </FluentPanelSplitterPane>
    <FluentPanelSplitterPane Size="50%">
        <FluentPanelSplitter Orientation="Orientation.Vertical">
            <FluentPanelSplitterPane Collapsible="true">
                Main Content
            </FluentPanelSplitterPane>
            <FluentPanelSplitterPane Collapsible="true">
                Console log
            </FluentPanelSplitterPane>
        </FluentPanelSplitter>
    </FluentPanelSplitterPane>
    <FluentPanelSplitterPane Size="30%">
        Properties
    </FluentPanelSplitterPane>
</FluentPanelSplitter>
```


c0nd3v:
BarSize should be used when calculating the size of the panes. If I have two panes and set them both to 50%, they do not have the same size because BarSize is not taken into account.

dvoituron:
> BarSize should be used when calculating the size of the panes. If I have two panes and set their Size both to 50%, they do not have the same size because BarSize is not taken into account.

I've updated the code so that the proportions between the panels are correct. The problem wasn't removing the BarSize, but setting the correct `flex` value in the CSS: `flex: 0 0 auto;` => `flex: 0 1 auto;`.
In this way, the panel size is automatically calculated by Flex.

dvoituron:
⚠️ Following an internal discussion, we probably renamed this component to `FluentMultiSplitter`. To give a more distinctive name (vs FluentSplitter) and improve clarity of use.

vnbaaij (approved):

PR #1334, opened 2024-01-15
[FluentSortableList] New component

vnbaaij:
# Fluent Sortable List
An implementation of the [SortableJS library](https://sortablejs.github.io/Sortable/) for Blazor Fluent UI. Allows for reordering elements within a list using drag and drop. Based on [Burke Holland's article and code](https://devblogs.microsoft.com/dotnet/introducing-blazor-sortable/). re-used with permission.

## How to use it in your own project
1. Add SortableJS to your `index.html`/`_Layout.cshtml`/`App.razor` file. You can either download from the [SortableJS website](https://sortablejs.github.io/Sortable/) or use a CDN:
```
<script src="https://cdn.jsdelivr.net/npm/sortablejs@1.13.0/Sortable.min.js"></script>
```

1. Use the `FluentSortableList.razor` in your page/app.
1. Add the SortableList component to your page and define the SortableItemTemplate...

        <FluentSortableList Id="sortable" Items="items" OnUpdate="@SortList" Context="item">
            <ItemTemplate>@item.Name</ItemTemplate>
        </FluentSortableList>

- Items: The list of items to be displayed. Can be of any type.
- OnUpdate: The method to be called when the list is updated.
- Context: The name of the variable to be used in the template. (Optional. Use `@context.` when omitted)

The `ItemTemplate` can contain any markup or components that you want. Each item is automatically wrapped in a `div` with class "sortable-item". The following CSS classes can be used to split the drag functionality from the content:
- sortable-grab: the grabable part of the dragable item
- sortable-item-content: the content part of the draggable item 

By using these seperate classes, you can define and style areas for grabbing and displaying.. See the examples on how this can be used.

### Parameters
Only the component's own parameters are described here , not the parameters inherited from base classes

`ItemTemplate`: `RenderFragment<TItem>?` (default: null) - The template to be used to define each sortable item in the list. Use the @context parameter to access the item and its properties.

`Items`: `IEnumerable<TItem>` (default: null) - The list of items to be displayed in the list.

`Group`: string (default: null) - Used for dragging between lists. Set the group to the same value on both lists to enable.

`Clone`: bool (default: false) - Used when you want to clone elments into a list instead of moving them. Set `clone` to true.

`Drop`: bool (default: true) - Set to false to disable dropping from another list onto the current list.

`Sort`: bool (default: true) - Disable sorting within a list by setting to `false`.

`Handle`: bool (default: false) - Set to true to use the `.sortable-grab` CSS class name you set on an element in the `ItemTemplate` to function as the drag handle. In the default situation the whole item acts as a drragable.

`ItemFilter` : `Func<TItem, bool>?` (default: null) - Sepcify a lambda to indicate which item(s) from the `Items` parameter should be filtered out of the dragging capability

### Methods

`OnUpdate` (Optional): The method to be called when the list is updated. You can name this method whatever you like, but it expects a `FluentSortableListEventArgs` (containing `OldIndex` and `NewIndex`) parameter when the drag and drop occurs.

`OnRemove` (Optional): The method to be called when an item is removed from a list. This method is useful for handling the drop even between multiple lists. Like `OnUpdate`, it expects a `FluentSortableListEventArgs` (containing `OldIndex` and `NewIndex`) parameter 

You will need to handle all events yourself. **If you don't handle any events, no sort or move will happen** as Blazor needs to make the changes to the underlying data structures so it can re-render the list. 

Here is an example of how to reorder your list when the OnUpdate is fired...

    private void SortList((int oldIndex, int newIndex) indices)
    {
        var (oldIndex, newIndex) = indices;

        var items = this.items;
        var itemToMove = items[oldIndex];
        items.RemoveAt(oldIndex);

        if (newIndex < items.Count)
        {
            items.Insert(newIndex, itemToMove);
        }
        else
        {
            items.Add(itemToMove);
        }

        StateHasChanged();
    }
sortablelist


examples/Demo/Server/Pages/_Layout.cshtml
```diff
@@ -32,7 +32,7 @@
     </div>
 
     <script src="_framework/blazor.server.js"></script>
-
+    <script src="https://cdn.jsdelivr.net/npm/sortablejs@1.13.0/Sortable.min.js"></script>
```
dvoituron, line 35: Why is it not the same version in Server vs Client?

examples/Demo/Shared/Pages/Lab/IssueTester.razor
```diff
@@ -1 +1,18 @@
-﻿@page "/IssueTester"
\ No newline at end of file
+﻿@page "/IssueTester"
```
dvoituron, line 1: This page is probably a "temp" work? :-)

src/Core/Components/SortableList/FluentSortableList.razor.cs
```diff
@@ -0,0 +1,84 @@
...
+    [Parameter, AllowNull]
+    public List<TItem> Items { get; set; }
+
+    [Parameter]
+    public EventCallback<(int oldIndex, int newIndex)> OnUpdate { get; set; }
+
+    [Parameter]
+    public EventCallback<(int oldIndex, int newIndex)> OnRemove { get; set; }
+
+    [Parameter]
```
dvoituron, line 27: Could you add Comments to write the documentation ?

src/Core/Components/SortableList/FluentSortableList.razor.cs
```diff
@@ -0,0 +1,84 @@
...
+    protected string? ClassValue => new CssBuilder(Class)
+        .AddClass("fluent-sortable-list")
+        .Build();
+
+    protected string? StyleValue => new StyleBuilder(Style)
+        .Build();
+
+    private DotNetObjectReference<FluentSortableList<TItem>>? selfReference;
+
+    public FluentSortableList()
```
dvoituron, line 54: Can you place the members in a clearly defined order?

See **Can you place the members in a clearly defined order?** in 
https://dvoituron.com/naming-conventions/rules/

src/Core/Components/SortableList/FluentSortableList.razor.cs
```diff
@@ -0,0 +1,84 @@
...
+    public string Filter { get; set; } = string.Empty;
+
+    protected string? ClassValue => new CssBuilder(Class)
+        .AddClass("fluent-sortable-list")
+        .Build();
+
+    protected string? StyleValue => new StyleBuilder(Style)
+        .Build();
+
+    private DotNetObjectReference<FluentSortableList<TItem>>? selfReference;
```
dvoituron, line 52: The name of a private member should begin with a "_".

src/Core/Components/SortableList/FluentSortableList.razor.cs
```diff
@@ -0,0 +1,84 @@
...
+    private const string JAVASCRIPT_FILE = "./_content/Microsoft.FluentUI.AspNetCore.Components/Components/SortableList/FluentSortableList.razor.js";
+
+    [Inject]
+    private IJSRuntime JSRuntime { get; set; } = default!;
+
+    [Parameter]
+    public RenderFragment<TItem>? SortableItemTemplate { get; set; }
+
+    [Parameter, AllowNull]
+    public List<TItem> Items { get; set; }
```
dvoituron, line 19: It would be better to use an `IEnumerable<TItem>`.

src/Core/Components/SortableList/FluentSortableList.razor
```diff
@@ -0,0 +1,17 @@
...
+@inherits FluentComponentBase
+@using System.Collections.Generic
+@using System.Diagnostics.CodeAnalysis
+
+@typeparam TItem
+
+<div id="@Id" class="@ClassValue" style="@StyleValue">
+    @foreach (var item in Items)
+    {
+        @if (SortableItemTemplate is not null)
```
dvoituron, line 11: Could you do the test before the `foreach`?

src/Core/Components/SortableList/FluentSortableList.razor.js
```diff
@@ -0,0 +1,34 @@
+export function init(id, group, pull, put, sort, handle, filter, component) {
```
dvoituron, line 1: Is it possible to test if the `Sortable` class exists, to try to download the latest version of the JS file `Sortable.min.js`?

In any case, you need to test the existence of this class, because if your JS hasn't had time to be downloaded, it will trigger an error.

dvoituron (changes requested):

examples/Demo/Server/Pages/_Layout.cshtml
vnbaaij: Trying to fix an error. Corrected it now

examples/Demo/Shared/Pages/Lab/IssueTester.razor
vnbaaij: Yup, cleared it out

src/Core/Components/SortableList/FluentSortableList.razor.cs
vnbaaij: Is on the list

src/Core/Components/SortableList/FluentSortableList.razor.js
vnbaaij: Yes, will work on that

src/Core/Components/SortableList/FluentSortableList.razor
dvoituron: If `SortableItemTemplate` is required, you could add `[EditorRequired]`

src/Core/Components/SortableList/FluentSortableList.razor.cs
vnbaaij: Done

src/Core/Components/SortableList/FluentSortableList.razor
vnbaaij: Added EditorRequired and moved the check up

dvoituron (changes requested):

PR #1368, opened 2024-01-22
[DataGrid] - Add Loading and LoadingContent parameters

vnbaaij:
This PR adds `Loading` (bool, default = false) and `LoadingContent` (RenderFragment?) parameters to the FluentDataGrid. 

By default, when `Loading` is set to true, the loading content will be a progress ring with the text 'Loading...'  shown besides it. You can replace that with your own content through this parameter.

To change the state of the `Loading` parameter at a later stage, use the `SetLoadingState(bool loading)` method on the `FluentDataGrid`. To be able to call this method from your code, you need to add a `@ref` to your grid

The loading content is rendered in a `FluentDataGridRow` with one `FluentDataGridCell` inside of it. The row height will be set to 100%. If no height is set for the grid this will be a normal row height. It a height is set for the grid, the row will take up all remaining height. In the example below, no height was specified for the grid.

datagrid-1317

In this example te height of the grid is set to 100% of its container. Loading content takes full height (minus header height)
image

With this PR we are also chaging the way the EmptyContent is rendered to use the same approach (`FluentDataGridRow` with one `FluentDataGridCell` inside and following same height settings). As this is a change from earlier versions, please check your results!

### DIFF
--- a/src/Core/Components/DataGrid/FluentDataGrid.razor
+++ b/src/Core/Components/DataGrid/FluentDataGrid.razor
@@ -129,7 +129,7 @@
 
     private void RenderPlaceholderRow(RenderTreeBuilder __builder, PlaceholderContext placeholderContext)
     {
-        string? _rowsDataSize = Loading ? "height: 100%;" : $"height: {ItemSize}px";
+        string? _rowsDataSize = Loading ? "height: 100%;" : $"height: {ItemSize}px;";
 
         <FluentDataGridRow GridTemplateColumns=@GridTemplateColumns aria-rowindex="@(placeholderContext.Index + 1)" Style="@_rowsDataSize" TGridItem="TGridItem">
             @for (var colIndex = 0; colIndex < _columns.Count; colIndex++)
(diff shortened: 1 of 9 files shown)


src/Core/Components/DataGrid/FluentDataGrid.razor.css
```diff
@@ -2,18 +2,18 @@ fluent-data-grid {
     --col-gap: 1rem;
 }
 
-::deep .empty-content-row {
+::deep .empty-content-row,
```
dvoituron, line 5: You should have `fluent-data-grid` before each sub-classed, to avoid conflicts if the dev use another class with same name (.empty-content-cell, ...)

src/Core/Components/DataGrid/FluentDataGrid.razor
```diff
@@ -109,7 +129,7 @@
 
     private void RenderPlaceholderRow(RenderTreeBuilder __builder, PlaceholderContext placeholderContext)
     {
-        string? _rowsDataSize = $"height: {ItemSize}px";
+        string? _rowsDataSize = Loading ? "height: 100%;" : $"height: {ItemSize}px";
```
dvoituron, line 132: Missing `;` at the end of `{ItemSize}px`

dvoituron (changes requested):

src/Core/Components/DataGrid/FluentDataGrid.razor.css
vnbaaij: Nope, because it is an isolated css, this is already scoped to this component only

dvoituron (approved):

PR #1372, opened 2024-01-23
List related refactoring and maintenance

vnbaaij:
- Change components in such a way that it does not matter if the options come from an `IEnumerable<TOption>` or from manually specified `FluentOptions`. We find the selected option with JS in the aria-activedescendant attribute and then call `OnClickHandler` on on the corresponding `FluentOption<TOption>`

- Simplify option rendering by using a single `RenderOptions` method. Move check if Items exist to this method. If no, just outpud ChildContent. If yes, render the `FluentOption`s

- Make Option<T> with Icon work for all types in `RenderOptions`. In earlier code this only worked when T was a string

- AutoComplete: remove virtual keyword

- Cleanup IssueTester


src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -135,6 +135,14 @@ public partial class FluentDataGrid<TGridItem> : FluentComponentBase, IHandleEve
     /// </summary>
     [Parameter] public RenderFragment? EmptyContent { get; set; }
 
+    [Parameter] public bool Loading { get; set; }
```
dvoituron, line 138: Could you add a documentation?

src/Core/Components/List/FluentOption.razor.cs
```diff
@@ -25,14 +24,14 @@ public partial class FluentOption<TOption> : FluentComponentBase, IDisposable wh
     /// Gets or sets a value indicating whether the element is selected.
     /// </summary>
     [Parameter]
-    public bool Selected { get; set; }
+    public bool? Selected { get; set; }
```
dvoituron, line 27: Why `Disabled` and `Selected` are nullable? In my mind, a value can be selected or not (same for a disabled value).

src/Core/Components/List/ListComponentBase.razor
```diff
@@ -0,0 +1,55 @@
+﻿@using Microsoft.AspNetCore.Components.Rendering
+@using System.Reflection
+
+@namespace Microsoft.FluentUI.AspNetCore.Components
+@inherits FluentComponentBase
+@typeparam TOption
+
+@code
+{
+    private void RenderOptions(RenderTreeBuilder __builder)
```
dvoituron, line 10: Is it possible to move this method to the Code behind (a .cs file) ?
A future Code Analysis rule will refuse the code in razor file.

src/Core/Components/List/ListComponentBase.razor.cs
```diff
@@ -145,6 +161,9 @@ public abstract class ListComponentBase<TOption> : FluentComponentBase where TOp
     [Parameter]
     public virtual TOption? SelectedOption { get; set; }
 
+    [Parameter]
+    public virtual FluentOption<TOption>? SelectedFluentOption { get; set; }
```
dvoituron, line 165: What is the goal of this property ?

src/Core/Components/List/ListComponentBase.razor.cs
```diff
@@ -550,67 +571,20 @@ protected virtual async Task RaiseChangedEventsAsync()
...
 
     protected virtual async Task OnKeydownHandlerAsync(KeyboardEventArgs e)
     {
-        if (e is null || Multiple || Items is null)
+        if (e is null || Multiple)
         {
             return;
         }
 
+        string id = await Module!.InvokeAsync<string>("getAriaActiveDescendant", Id);
```
dvoituron, line 583: Could you test the value of Module: `if (e is null || Multiple || Module is null)`.
If the compiler says that Module may be null, it's best to reject this case explicitly (and avoid using `!`).

src/Core/Components/List/ListComponentBase.razor.cs
```diff
@@ -550,67 +571,20 @@ protected virtual async Task RaiseChangedEventsAsync()
...
-            case "Escape":
-            default:
-                break;
-        }
-
-        await OnSelectedItemChangedHandlerAsync(item);
-
+        FluentOption<TOption> item = _internalListContext.Options.First(i => i.Id == id);
 
+        await item.OnClickHandlerAsync();
```
dvoituron, line 587: Where is the code for using the arrow keys (Up, Down, Home, Escape), included in the previous code?

dvoituron (changes requested):

src/Core/Components/List/FluentOption.razor.cs
vnbaaij: If they are null, they will not be rendered by the web component as attributes so web component can apply its default rendering

src/Core/Components/List/ListComponentBase.razor.cs
vnbaaij: It is not needed anymore as it is already handled by the web component

src/Core/Components/List/ListComponentBase.razor
dvoituron: Yes, you could to that, returning a `RenderFragment` object. The single difficulty will be to create `FluentOption` by code :-)

src/Core/Components/List/FluentOption.razor.cs
dvoituron: I don't think so. 

Example: 
```razor
<h3 my-flag="@MyFlag">MyComponent</h3>

@code {
    [Parameter]
    public bool MyFlag { get; set; }
}
```

Will generate these code: `my-flag` will be generated only when the value if True.

```html
<MyComponent />                    ->   <h3>MyComponent</h3>
<MyComponent MyFlag="true" />      ->   <h3 my-flag>MyComponent</h3>
<MyComponent MyFlag="false" />     ->   <h3>MyComponent</h3>
```

dvoituron (changes requested):

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
vnbaaij: Done in PR #1375

src/Core/Components/List/FluentOption.razor.cs
vnbaaij: Right, bool is different that way (comared to string value). Will revert.

src/Core/Components/List/ListComponentBase.razor.cs
vnbaaij: Left over from coding efforts. Will remove

src/Core/Components/List/ListComponentBase.razor
vnbaaij: Will work on that for later release

dvoituron (approved):

PR #1382, opened 2024-01-25
[DatePicker] Fix popup styling to match similar elements

hksalessio:
# Pull Request

## 📖 Description

The styling of the `FluentDatePicker`'s calendar popup was a little off (no rounded borders, no space between input element and popup). This pull request fixes this issue by applying a styling similar to other popups (e.g. `FluentSelect`).

Before | After
---|---
image | image

### General

### Component-specific


src/Core/Components/DateTime/FluentDatePicker.razor
```diff
@@ -22,6 +22,11 @@
 
 @if (Opened)
 {
+    <style>
+        @($"#{Id} ~ fluent-anchored-region .fluent-datepicker-popup {{ padding: 10px; border-radius: calc(var(--control-corner-radius) * 1px); position: absolute; }}")
```
dvoituron, line 26: Why not include it in `FluentDatePicker.razor.css` so that it's common to all components? And so that it can be overloaded by other users if necessary.

And why did you set `position: absolute`?

dvoituron (changes requested):

src/Core/Components/DateTime/FluentDatePicker.razor
hksalessio: I have tried that but Blazor's CSS Isolation always adds this ID thing to the class, which makes it impossible to target the correct element. The only way I can think of evading this behavior would be to add a wrapper around the input element and the popup. If you are fine with this, I could change the code tomorrow. And absolute positioning was necessary to be able to use bottom and top respectively.

vnbaaij: The I'd things is what you use the ::deep selector for. It does the css to the component

hksalessio: I know, it's just that the ID is applied in the CSS but not the to the element. I guess that's due to the component having multiple child elements on the same level (the input element and the popup). If there is only one child (like a wrapper), the ID would is applied correctly.

hksalessio: I might have gotten this wrong. However, the ID is indeed not applied to the `fluent-text-field` element generated by the `FluentDatePicker`:
image
Could this be a bug with the web components?

vnbaaij: In the component CSS you can refer to (and should use) `::deep > .fluent-datepicker` (or something similair) target the text filed inside the component.

dvoituron: @hksalessio It's not a bug, the reason is that `label` is an HTML element but `fluent-text-field` is a Web Component and Blazor only adds scoped identifier to HTML element.

hksalessio: > In the component CSS you can refer to (and should use) `::deep > .fluent-datepicker` (or something similair) target the text filed inside the component.
> 
> You can find examples of this in the other components that have a scoped css file.

It appears, this still doesn't work. The CSS that is generated looks something like
```css
/* _content/Microsoft.FluentUI.AspNetCore.Components/Components/DateTime/FluentDatePicker.razor.rz.scp.css */
[b-i3i7ux5rp5] > .fluent-datepicker {
    ...
}
```
But when searching for 'b-i3i7ux5rp5' in the DOM, this string is not found. As a result, no styling is applied.

hksalessio: > @hksalessio It's not a bug, the reason is that `label` is an HTML element but `fluent-text-field` is a Web Component and Blazor only adds scoped identifier to HTML element.

How would I go about targeting the correct element then? Why is such a small problem, such an annoying task to solve? 😄

dvoituron: I don't have a quick solution to this issue.

We could add an "extra" global `div`... probably the best solution today.

Perhaps we should give some thought to the possibility of "forcing" the scoped CSS identifier using code like this.
```xml
<ItemGroup>
    <None Update="*/*.razor.css" CssScope="b-fui" />
</ItemGroup>
```

vnbaaij: One more remark...In React V9, the DAtePicker does not have a spacing between text field and calendar (but Select has as you correctly said). THink we should not deviate from the React design too much, so maybe do not apply the spacing in this PR.
(Still want to have the conrners etc..)

dvoituron (approved):

PR #1405, opened 2024-01-29
Improve navmenu-icon position demo

dromerolovo:
I've moved the icon to the header section so that it can be relatively positioned and not fixed with hard coded values, improving its position with a clear separation. 

Before: 
Screenshot 2024-01-29 155028

After:
Screenshot 2024-01-29 165009


vnbaaij (commented):
The label and icon need to be part of the `DemoNavMenu.razor` file for a reason. Has to do with media queries etc. Won't merge this but it will be fixed in a different way. Also, with your fix the most right icon is not aligned to the right marging anymore.

dromerolovo:
@vnbaaij Sounds good. I just made the simplest solution that I thought of.

PR #1409, opened 2024-01-30
2 small fixes 

vnbaaij:
- Fix how FluentInputFile disabled is set
- NavMenu: only add KeyCode on 'root' when Collapsible


src/Core/Components/InputFile/FluentInputFile.razor.cs
```diff
@@ -42,6 +42,14 @@ public FluentInputFile()
         .AddStyle("display", "none", () => !DragDropZoneVisible)
         .Build();
 
+    internal Dictionary<string, object?> Attributes
```
dvoituron, line 45: Why not to set `disabled="@Disabled"` directly in the **div** tag?

dvoituron (changes requested):

src/Core/Components/InputFile/FluentInputFile.razor.cs
vnbaaij: Done.

dvoituron (approved):

PR #1411, opened 2024-01-31
Add inputmode attribute support for FluentTextField

martinbu:
This pull adds an InputMode parameter for the FluentTextField. The same "workaround" is used to set the attribute on the native input control as it is already implemented for the autocomplete attribute.

Maybe this workaround can be adapted when https://github.com/microsoft/fast/issues/6082 is fixed.


martinbu:

src/Core/Components/TextField/FluentTextField.razor.cs
```diff
@@ -92,9 +98,17 @@ protected override async Task OnAfterRenderAsync(bool firstRender)
...
             {
                 Module ??= await JSRuntime.InvokeAsync<IJSObjectReference>("import", JAVASCRIPT_FILE);
-                await Module.InvokeVoidAsync("setAutocomplete", Id, AutoComplete);
+                await Module.InvokeVoidAsync("setControlAttribute", Id, "autocomplete", AutoComplete);
+            }
+
+            if (InputMode != null && !string.IsNullOrEmpty(Id))
+            {
+                Module ??= await JSRuntime.InvokeAsync<IJSObjectReference>("import", JAVASCRIPT_FILE);
+                var inputModeValue = InputMode.Value.ToString().ToLower();
```
dvoituron, line 107: You have an extension method `ToAttributeValue` to do that.

src/Core/Enums/InputMode.cs
```diff
@@ -0,0 +1,47 @@
...
+
+    /// <summary>
+    /// Numeric input keyboard, but only requires the digits 0–9. Devices may or may not show a minus key. 
+    /// </summary>
+    Numeric,
+
+    /// <summary>
+    /// A telephone keypad input, including the digits 0–9, the asterisk (*), and the pound (#) key. Inputs that *require* a telephone number should typically use <see cref="TextFieldType.Tel"/> instead. 
+    /// </summary>
+    Tel,
```
dvoituron, line 31: Could you use `Telephone` to avoid the abbreviation. It's not good practice to use an abbreviation that may be different for some people/cultures.

src/Core/Components/TextField/FluentTextField.razor.cs
```diff
@@ -76,6 +76,12 @@ public partial class FluentTextField : FluentInputBase<string?>
     [Parameter]
     public string? AutoComplete { get; set; }
 
+    /// <summary>
+    /// Gets or sets the input mode. See <see cref="AspNetCore.Components.InputMode"/>
```
dvoituron, line 80: Could you describe this property as follows?
> Gets or sets the type of data that can be entered by the user when editing the element or its content. This allows a browser to display an appropriate virtual keyboard. Not supported by Safari.

dvoituron (changes requested):

src/Core/Enums/InputMode.cs
vnbaaij: You can add `using System.ComponentModel;` to your Enum and then add `[Description("...")]` attribute to each element to set the value that needs to be used in code (ie in this case 'tel' for 'Telephone'.
The `.ToAttributeValue()` Denis mentioned does that for you.

vnbaaij:
Great addition. Thanks!

dvoituron (approved):

PR #1414, opened 2024-01-31
NavMenu / KeyCode optimizations

vnbaaij:
- [NavMenu] Abort KeyDown handlers if target id not correct
- [NavMenu] Bind Expander KeyDown handler to the item instead of the whole menu
Also:
- Fix smal MessageBarDemo issue


src/Core/Components/NavMenu/FluentNavMenu.razor
```diff
@@ -4,15 +4,13 @@
...
-{
-    <FluentKeyCode Anchor="@Id" OnKeyDown="@HandleExpandCollapseKeyDownAsync" />
-}
 <div id="@Id" @ref="@Element" @attributes="AdditionalAttributes" class="@ClassValue" style="@StyleValue" aria-label="@(Collapsible ? null : @Title)" role="menu" aria-expanded="@Expanded">
     <CascadingValue Value="this" IsFixed="true">
         @if (Collapsible)
         {
-            <div aria-label="@Title"
+            <FluentKeyCode Anchor="expander" OnKeyDown="@HandleExpandCollapseKeyDownAsync" />
+            <div id="@EXPAND_COLLAPSE_BUTTON_ID"
```
dvoituron, line 12: The div ´id` must be unique: you have to use `@Id`, for example adding a private property `ExpandCollapseButtonId => $"{Id}-expander"`

src/Core/Components/NavMenu/FluentNavMenu.razor.js
```diff
@@ -22,6 +22,18 @@ export function onUpdate() {
     
 }
 
+export function onDispose() {
```
dvoituron, line 25: When this `onDispose` method will be called?
I don't know how to detect when an HTML component will be "disposed".

src/Core/Components/NavMenu/FluentNavMenu.razor.js
```diff
@@ -22,6 +22,18 @@ export function onUpdate() {
...
+        }
+    }
+    for (let element of document.getElementsByClassName("fluent-nav-group")) {
+        detachEventHandlers(element);
+    }
+}
+
 function attachEventHandlers(element) {
     let navlink = element.getElementsByClassName("fluent-nav-link")[0];
     if (!navlink.href) {
```
dvoituron, line 39: Perhaps you should add a test to check whether the `navLink` element has been found (to handle all cases)

src/Core/Components/NavMenu/FluentNavMenu.razor.js
```diff
@@ -33,6 +45,17 @@ function attachEventHandlers(element) {
...
 }
 
+function detachEventHandlers(element) {
+    let navlink = element.getElementsByClassName("fluent-nav-link")[0];
+    if (!navlink.href) {
+        navlink.removeEventListener('click', toggleGroupExpandedAsync);
+    }
+    navlink.removeEventListener('keydown', handleExpanderKeyDownAsync);
+
+    let expandCollapseButton = element.getElementsByClassName("expand-collapse-button")[0];
```
dvoituron, line 55: Same, you should test if `expandCollapseButton` exists.

dvoituron (changes requested):

src/Core/Components/NavMenu/FluentNavMenu.razor.js
vnbaaij: It will get called as part of the PageScript flow (not happening right now, but to be prepared for it)

src/Core/Components/NavMenu/FluentNavMenu.razor
```diff
@@ -4,15 +4,13 @@
...
-{
-    <FluentKeyCode Anchor="@Id" OnKeyDown="@HandleExpandCollapseKeyDownAsync" />
-}
 <div id="@Id" @ref="@Element" @attributes="AdditionalAttributes" class="@ClassValue" style="@StyleValue" aria-label="@(Collapsible ? null : @Title)" role="menu" aria-expanded="@Expanded">
     <CascadingValue Value="this" IsFixed="true">
         @if (Collapsible)
         {
-            <div aria-label="@Title"
+            <FluentKeyCode Anchor="expander" OnKeyDown="@HandleExpandCollapseKeyDownAsync" />
+            <div id="${@Id}-expander"
```
dvoituron, line 12: The id will be `id="${h154sjjhd2}-expander"`. Is it normal to include `{` and `$` ?

dvoituron (changes requested):

src/Core/Components/NavMenu/FluentNavMenu.razor
vnbaaij: I always do that. Don't you? 😉

dvoituron (approved):

PR #1424, opened 2024-02-02
Not setting `current-value` for fluent-number-field

cupsos:
# Pull Request

## 📖 Description

In Blazor Server, holding down `Space`, `ArrowUp`, or `ArrowDown` leads to a loop in value changes.
This issue has been reproduced in both the dev branch and v4.3.1.

switch-loop
number-field-loop

Both fluent-switch and fluent-number-field have `value` and `current-value` properties.

The difference between them is that `current-value` represents the real-time value displayed in the UI.
This means that `current-value` and `value` can differ, especially with rapid inputs.

Since setting either `current-value` or `value` triggers a change in value, under certain conditions, this can cause a loop in value changes.

This PR resolves the issue by not setting `current-value`.

```mermaid
flowchart TD
subgraph fluent-ui web-component
current-value
value
onchange
end

subgraph blazor-server
SetCurrentValueAsync
StateHasChanged
end

value
--> onchange
-->|fire on value changed| SetCurrentValueAsync
--> StateHasChanged
-->|set value if diffrent| value & current-value
value -->|change each ohther| current-value
current-value -->|change each ohther| value
```

## 👩‍💻 Reviewer Notes

There are other components that set `current-value` and `value`, but since the bug could not be reproduced with them, they were not included in the commit.

### General

### Component-specific

### DIFF
--- a/src/Core/Components/NumberField/FluentNumberField.razor
+++ b/src/Core/Components/NumberField/FluentNumberField.razor
@@ -25,7 +25,6 @@
                      min="@(ReadOnly ? value : Min)"
                      id=@Id
                      value="@value"
-                     current-value="@value"
                      disabled="@Disabled"
                      name=@Name
                      required="@Required"


cupsos:

dvoituron:
I agree with you that this is a bug we need to fix.

But I don't think we need to remove `current-value` from the components. Because this attribute represents the value sent by the form. `current-value` must be set when the component is first rendered, to `on` for example, and must not change afterwards.

For example, using this simple HTML code using the Web Component (same for `fluent-switch`).

```html
<html>
  <head>
    <script type="module" src="https://unpkg.com/@fluentui/web-components"></script>
  </head>
  <form action="" method="get">
    <input name="myOne" checked value="1" current-value="2" type="checkbox">One
    <fluent-checkbox name="myTwo" value="3" current-value="4">Two</fluent-checkbox>
    <button type="submit">Save</button>
  </form>
</html>
```

image

cupsos:
Oh, I didn't know that. Yes, you're right.

https://github.com/microsoft/fluentui-blazor/blob/ed2a3c8d54ddd5bfe8efc84a03b379855072f959/src/Core/Components/Base/FluentInputBase.cs#L166-L173

Right now, can't make a fluent-switch with different `value` and `current-value` because `FluentInputBase.CurrentValue` is protected and the same as `Value`. Fixing the code that uses `CurrentValue` and changing the API is hard for me.

I think can remove `current-value` from fluent-number-field for now. Pressing up or down arrows in fluent-number-field is common, but it's unlikely people will keep pressing space in fluent-switch.

If it's okay, I'll undo the changes to fluent-switch and change the PR title.

vnbaaij:
I've done some more check and looked at it with @dvoituron. I removed the code that put both `value` and `current-value` in the `fluent-switch` web component (in our `FluentSwitch` component) in a separate PR

vnbaaij (approved):

dvoituron:
Could you update the Unit Tests?
image

Issue #1448, opened 2024-02-05
fix: two key strokes required to change item in FluentSelect


majkada:
# 🐛 Bug Report

While using a FluentSelect with a keyboard 2 key strokes (either arrow up or arrow down) are required to actually change an item.

## 💻 Repro or Code Sample

E.g. it can be reproduced on https://www.fluentui-blazor.net/Forms with the "Primary Classification" select.

## 🤔 Expected Behavior

Only one key stroke is required to change an item.

## 😯 Current Behavior

Exactly two key strokes are required.

* OS & Device: Windows on PC
* Browser Google Chrome
* FluentUI 4.4.0

vnbaaij:
The strange thing is this only happens in WASM and it is not necessary for every item in a select. Even stranger is that if you add an `fluent-select` web component, it does work as expected. On Server it works with one keypress regardless.

I'll leave this open, but as the impact is really low and the component works as it should , we won't prioritize finding a solution for it.

vnbaaij:
Ok, Adding that delay for the next release 4.4.1.

PR #1464, opened 2024-02-06
Users/vnbaaij/several fixes

### LINKED ISSUE #1450: fix: FluentSlider's two-way data binding on Value Property does not update appropriately
# 🐛 Bug Report

In my web app, I have a component that requires a slider reaches its max value, enabling a selection field, and then after a selection is made, a button triggers a command and this command also resets the value of the slider to 0.

In Fluent UI, the slider's value stays at the maximum value both as a property and visually, even though the bound data model has reset to 0.

I've prepared a demo of the problematic behaviors here: https://github.com/arazni/blazor-fluent-ui-slider-bug/tree/main

## 💻 Repro or Code Sample

Download and run the code here in dotnet 8: https://github.com/arazni/blazor-fluent-ui-slider-bug/tree/main

## 🤔 Expected Behavior

When I click a button that resets the two-way bound value of the Slider to 0, the Slider's Value property should change to 0 and the slider should visibly appear to reset to 0.

image

Click Reset.

image

## 😯 Current Behavior

When I click a button that resets the two-way bound value of the Slider to 0, the Slider's Value property stays at its current value and the slider's visually apparent value stays at its current value.

image

Click Reset.

image

Button becomes disabled (data bound is 0, but slider is still maxed).

## 💁 Possible Solution

Fix the value-binding to respect situations where the data model's property changes, and ensure that the slider visually changes as well.

## 🔦 Context

In my web app, I have a component that requires a slider reaches its max value, which enables a selection field. Once the selection field is set, a button enables that triggers a command and this command also resets the value of the slider to 0.

In the context of the web app, which is a character sheet for Blades in the Dark, when the character reaches maximum stress (the slider), they become traumatized. They select a trauma, press a button to add it, and their stress resets to 0.

## 🌍 Your Environment

* OS & Device: Windows 11 PC
* Browser at least: Microsoft Edge and Mozilla FireFox]
* .NET 8.0.1 and FluentUI Version 4.4.0

vnbaaij:
Fix #1450, #1448 and NotificationCenterPanel


.editorconfig
```diff
@@ -26,6 +26,16 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 spelling_exclusion_path = spelling.dic
+dotnet_style_coalesce_expression = true:warning
```
dvoituron, line 29: These config are already included in this file.

dvoituron (changes requested):

.editorconfig
vnbaaij: I know, but just opening the file in VS adds this

dvoituron (approved):

PR #1480, opened 2024-02-08
Use specific ids for rows and cells

### LINKED ISSUE #1478: fix: RowId error in FluentDataGrid
# 🐛 Bug Report

A page with many rows in FluentDataGrid is erroring.

## 💻 Repro or Code Sample

See this issue: https://github.com/dotnet/aspire/issues/2103.

The UI is attempting to display a data grid with 13,000 rows. Obviously not a good thing to do, and we'll investigate whether virtualization can be used.

However, the scenario creates an error caused by multiple rows having the same row ID and then erroring when added to a dictionary:

```
warn: Microsoft.AspNetCore.Components.Server.Circuits.RemoteRenderer[100]
      Unhandled exception rendering component: An item with the same key has already been added. Key: f44344300
      System.ArgumentException: An item with the same key has already been added. Key: f44344300
         at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
         at Microsoft.FluentUI.AspNetCore.Components.DataGrid.Infrastructure.InternalGridContext`1.Register(FluentDataGridRow`1 row)
         at Microsoft.AspNetCore.Components.ComponentBase.RunInitAndSetParametersAsync()
```

It looks like each row has an ID and that ID is randomly generated: 

https://github.com/microsoft/fluentui-blazor/blob/0536a0b78fd4e87dd58320a96ef0dfc6a00c997f/src/Core/Components/DataGrid/FluentDataGridRow.razor.cs#L10
https://github.com/microsoft/fluentui-blazor/blob/0536a0b78fd4e87dd58320a96ef0dfc6a00c997f/src/Core/Identifier.cs#L35

What that means is that every time there are more than two rows in the grid it is a lottery whether the same ID is generated and there is an error. The more rows there are, the more likely the error happens.

## 🤔 Expected Behavior

No error.

## 😯 Current Behavior

See error above.

## 💁 Possible Solution

I think you need to redesign the ID generation. A spot solution could be to check if the generated row ID is already in use and then generate another one.

Using an incrementing ID (with thread safety) seems like it would be better, e.g. the result of `Interlocked.Increment`. IDs could be unique within the context of a Blazor Circuit. If you reach max value, then loop around and start at the beginning again. It seems like it would be impossible to have IDs conflict within the context of a page.

## 🌍 Your Environment

* .NET and FAST Version [e.g. 1.8.0]

vnbaaij:
Fix #1478 by creating new sequential ids for rows/cells per grid

### DIFF
--- a/src/Core/Components/DataGrid/FluentDataGridRow.razor.cs
+++ b/src/Core/Components/DataGrid/FluentDataGridRow.razor.cs
@@ -7,7 +7,7 @@ namespace Microsoft.FluentUI.AspNetCore.Components;
 [CascadingTypeParameter(nameof(TGridItem))]
 public partial class FluentDataGridRow<TGridItem> : FluentComponentBase, IHandleEvent, IDisposable
 {
-    internal string RowId { get; } = Identifier.NewId();
+    internal string RowId { get; set; } = string.Empty;
     private readonly Dictionary<string, FluentDataGridCell<TGridItem>> cells = [];
 
     /// <summary>
@@ -61,6 +61,7 @@ public partial class FluentDataGridRow<TGridItem> : FluentComponentBase, IHandle
 
     protected override void OnInitialized()
     {
+        RowId = $"r{Owner.GetNextRowId()}";
         Owner.Register(this);
     }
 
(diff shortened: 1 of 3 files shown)


src/Core/Components/DataGrid/FluentDataGridRow.razor.cs
```diff
@@ -61,19 +61,22 @@ public partial class FluentDataGridRow<TGridItem> : FluentComponentBase, IHandle
...
     {
+
+        cell.CellId = $"c{Owner.GetNextCellId()}";
         cells.Add(cell.CellId, cell);
     }
 
     internal void Unregister(FluentDataGridCell<TGridItem> cell)
     {
-        cells.Remove(cell.CellId);
+        cells.Remove(cell.CellId!);
```
dvoituron, line 79: If VS says that this variable can be nullable. It's true in some cases 😊 Maybe a test to avoid the Remove in this case?

dvoituron (approved):

src/Core/Components/DataGrid/FluentDataGridRow.razor.cs
vnbaaij: I made it nullable during testing. I the end I reverted that so it isn't nullable anymore. Just forgot to remove the bang.

Issue #1387, opened 2024-01-26
fix: NavMenu Expanded won't update if prerendering is disabled


aco-mreble:
# 🐛 Bug Report

When using `@bind-Expanded="expanded"` or `ExpandedChanged="OnExpandedChanged"` and when prerendering is disabled `expanded` won't update/`OnExpandedChanged` won't be called when clicking on the Hamburger Menu.

## 💻 Repro or Code Sample
Create a new template with 
```shell
dotnet new fluentblazor -o FluentUiTest
```
go to `FluentUiTest\Components\Layout\NavMenu.razor` and add this line of code anywhere on the page:
```
<p>@expanded</p>
```
Run the application. Everything works as expected, the text in the `<p>` tag will update accordingly to true/false.

Now go to `App.razor` and replace the `<Routes />` line with this
```
<Routes @rendermode="new InteractiveServerRenderMode(prerender: false)"/>
```
Run the application again and see, that the text in the `<p>` tag won't change anymore.

## 🌍 Your Environment

* Blazor with interactive server side rendering enabled
* Edge
* .net 8.0.1 with fluent-ui 4.3.1

vnbaaij:
Never mind, I see what you mean in a proper Server site...

What is happening is that the FluentNavMenu script is handeling the expanding/collapsing with Javascript. That is why @expanded never gets updated. I'm asking internally if it is possible to hae  script that **only** runs in SSR mode and uses Blazr when interactivity is there,

vnbaaij:
Apparently there is no way to have a script that is run only when in SSR mode. So, for now we can either:
a) remove the script - We end app with a menu that does not expand/colapse (and same for groups inside the menu) in SSR mode
b) keep the script - in a couple of specific scenarios you can't get the expanded state of the menu. Everything else works

I vote for b 😀

Closing this as it is currently not possible to solve in another way

PR #1560, opened 2024-02-23
fix FluentNavMenu.razor.js for non SSR

c0g1t8:
# Pull Request

## 📖 Description

Fix is related to #1387 and specifically https://github.com/microsoft/fluentui-blazor/issues/1387#issuecomment-1921267970.

### 🎫 Issues

As stated in the comment, the collapse/expanding was being handled with the javacript added for SSR. The original code works correctly for all **fluentblazor** projects but not for **fluentblazorwasm** projects.

## 👩‍💻 Reviewer Notes

The fix provided is to prevent the addition of EventListeners when WASM rendered. It works with _.net8_ .

## 📑 Test Plan

The fix has been tested for:
-  _fluentblazorwasm_
- for all four _fluentblazor_ interative rendering modes.

Note: 

The EventListeners for collapse/expand disappear with interactive render modes Server, WebAssembly, and Auto. When stepping through the original _FluentNavMenu.razor.js_ with the debugger, you can see the eventlisteners being attached. Then suddenly disappear. This may be related to the fact that the _expander_ element also has a blazor based _onClick_ handler.

### General

### Component-specific


src/Core/Components/NavMenu/FluentNavMenu.razor.js
```diff
@@ -1,37 +1,47 @@
...
+        for (let element of document.getElementsByClassName("fluent-nav-group")) {
+            detachEventHandlers(element);
         }
     }
-    for (let element of document.getElementsByClassName("fluent-nav-group")) {
-        detachEventHandlers(element);
-    }
+}
+
+function isRunningInWASM() {
```
c0g1t8, line 42: This test is dependent on the implementation of the **DotNet** in .net 8.

dvoituron: Where does this command come from? Do you have a reference in the documentation?

What happens in "auto" mode where the component is initially displayed with interactive server-side rendering (interactive SSR) using the Blazor Server hosting model. The .NET WASM runtime and the entire application are downloaded to the client in the background and cached for use on subsequent visits.

dvoituron (changes requested):

c0g1t8:
@dvoituron 

> Where does this command come from? Do you have a reference in the documentation?

I'm calling the same **DotNet** object that is used to [Invoke a static .NET method](https://learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/call-dotnet-from-javascript?view=aspnetcore-8.0#invoke-a-static-net-method). It comes from the _"_framework/blazor.web.js"_ file.

I'm checking for the existence of a **Symbol** of '**wasm type**'. This PR is based on using something undocumented and possibly implementation specific to .net8. Rejecting it on this basis is reasonable. 🙂

> What happens in "auto" mode where the component is initially displayed with interactive server-side rendering (interactive SSR) using the Blazor Server hosting model. The .NET WASM runtime and the entire application are downloaded to the client in the background and cached for use on subsequent visits.

I see your point. Is it interactive before the switch to the WASM runtime? I need to test that again.

Retagging PR as WIP

c0g1t8:
@dvoituron 

I did some more testing of the PR using Auto interactivity. Interactivity is available before switching to WASM.

I created a new _fluentblazor_ project using Auto interactivity. I then, removed the dependency on the nuget package and added a dependency to a local copy of _Microsoft.FluentUI.AspNetCore.Components_. Edited the _FluentNavMenu.razor.js_ file and added a console log message to the _toggleMenuExpandedAsync_ function. Set the browser to throttle network and ran the project.

I started clicking the expander toggle as soon as it appeared. The console log showed that the javacript function was being called. After a few clicks, the console log did not show the javascript function being called but the menu was being toggled.

c0g1t8:
@dvoituron 

I found some documentation on the [ASP.NET Core Blazor startup](https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/startup?view=aspnetcore-8.0) process. A **Blazor** object is exposed in javascript as part of the startup process. This mean _isRunningInWASM_ can be rewritten with **Blazor.runtime**.

``` javascript
function isRunningInWASM() {
    return Object.getOwnPropertyNames(Blazor.runtime).length > 0
}
```

**Blazor.runtime** only has properties when running WASM.

During the startup process, the **onLoad** function in _FluentNavMenu.razor.js_ is called. In all server rendered scenarios, **Blazor.runtime** has no properties since WASM hasn't started yet. The only time **Blazor.runtime** has properties during startup is when hosted client side in WASM.

dvoituron:
> I'm checking for the existence of a **Symbol** of '**wasm type**'. This PR is based on using something undocumented and possibly implementation specific to .net8. Rejecting it on this basis is reasonable. 🙂

It's actually this part that disturbs me a little: using an element that is not documented and can therefore change at any time is very risky in my opinion. I don't think it's a good idea. It would be better to find another "safe" way of detecting whether you're in WASM or not. One solution might be to provide an `isRunningInWASM` property which returns this value, but which could be overridden by the developer in the event of a problem (or a change in the ASPNET team).

What's more, does this remain valid in all other modes of use, such as MAUI for example?

vnbaaij:
> @dvoituron
> 
> I found some documentation on the [ASP.NET Core Blazor startup](https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/startup?view=aspnetcore-8.0) process. A **Blazor** object is exposed in javascript as part of the startup process. This mean _isRunningInWASM_ can be rewritten with **Blazor.runtime**.
> 
> ```js
> function isRunningInWASM() {
>     return Object.getOwnPropertyNames(Blazor.runtime).length > 0
> }
> ```
> 
> **Blazor.runtime** only has properties when running WASM.
> 
> During the startup process, the **onLoad** function in _FluentNavMenu.razor.js_ is called. In all server rendered scenarios, **Blazor.runtime** has no properties since WASM hasn't started yet. The only time **Blazor.runtime** has properties during startup is when hosted client side in WASM.

This is the way to go indeed. The Blazor object is not going to change anytime soon (if at all). But interested to see if this also works with Blazor Hybrid/MAUI setup as @dvoituron asked.

vnbaaij (approved):

vnbaaij (approved):

c0g1t8:
I didn't test the Blazor Hybrid/MAUI scenarios. I'll try to run a test before the end of the week and respond here.

vnbaaij:
Thanks Gary!

c0g1t8:
@vnbaaij 

I finally got around to running a Blazor **Hybrid/MAUI** test. This fix does not cover this scenario. The javascript listeners are still being attached.

This makes sense given that MAUI generates native applications and WASM is not needed. A test to see if the component is running in **Blazor Hybrid/MAUI** is needed.

c0g1t8:
@vnbaaij 

I think I have a fix for this. I'll create a draft PR later today. 🙂

PR #1574, opened 2024-02-26
FluentMenu fixes

vnbaaij:
- Fix visual inconsistency with the FluentMenu by adding a border radius to the anchored region surrounding the menu
- Fix context menu cannot be closed without selecting a menu option by adding an overlay

### DIFF
--- a/src/Core/Components/Menu/FluentMenu.razor
+++ b/src/Core/Components/Menu/FluentMenu.razor
@@ -1,4 +1,4 @@
-@namespace Microsoft.FluentUI.AspNetCore.Components
+﻿@namespace Microsoft.FluentUI.AspNetCore.Components
 @inherits FluentComponentBase
 <CascadingValue Value="this" IsFixed="true">
     @if (Open)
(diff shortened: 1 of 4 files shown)


src/Core/Components/Menu/FluentMenu.razor
```diff
@@ -23,6 +23,9 @@
         }
         else
         {
+            if (_contextMenu) {
```
dvoituron, line 26: Why not to move the `FluentOverlay` from line 8 before `@if (Anchored)`, to have this possibility for all menus? In other words, why to add an Overlay only for `_contextMenu` = `Trigger == MouseButton.Right` ?

dvoituron (changes requested):

src/Core/Components/Menu/FluentMenu.razor
vnbaaij: That lead to errors on the demo page. Some menus don't show or are not clickable then because the Overlay gets in the way

dvoituron: Weird. OK. I trust in you :-)

dvoituron (approved):

PR #1577, opened 2024-02-26
Fix OnRowwFocus and sorting for DataGrid

### LINKED ISSUE #1540: fix: FluentDataGrid OnRowFocus and sorting not working when contained in FluentGrid
# 🐛 Bug Report

I tried to implement an OnRowClick event for the FluentDataGrid as already mentioned in #779 (I understand that this is not possible because of the underlying library). Then i stumbled across the same bug reported in #1343 but can now reproduce it.

## 💻 Repro or Code Sample

1. Create a new Blazor Server project with .NET 8
2. Add to csproj the PackageReference "Microsoft.FluentUI.AspNetCore.Components" in Version 4.4.1 (latest as of now)
3. Change content of Home.razor to

```
@page "/"

<PageTitle>People</PageTitle>

<FluentGrid>
    <FluentGridItem>
        <FluentDataGrid Items="@people" TGridItem="Person" OnRowFocus="HandleRowFocus">
            <PropertyColumn Property="@(p => p.PersonId)" Sortable="true" />
            <PropertyColumn Property="@(p => p.Name)" Sortable="true" />
            <PropertyColumn Property="@(p => p.BirthDate)" Format="yyyy-MM-dd" Sortable="true" />
        </FluentDataGrid>
    </FluentGridItem>
</FluentGrid>

@code {
    record Person(int PersonId, string Name, DateOnly BirthDate);

    IQueryable<Person> people = new[]
    {
        new Person(10895, "Jean Martin", new DateOnly(1985, 3, 16)),
        new Person(10944, "António Langa", new DateOnly(1991, 12, 1)),
        new Person(11203, "Julie Smith", new DateOnly(1958, 10, 10)),
        new Person(11205, "Nur Sari", new DateOnly(1922, 4, 27)),
        new Person(11898, "Jose Hernandez", new DateOnly(2011, 5, 3)),
        new Person(12130, "Kenji Sato", new DateOnly(2004, 1, 9)),
    }.AsQueryable();

    private void HandleRowFocus(FluentDataGridRow<Person> row)
    {
        Console.WriteLine($"Row focused: {row.Item?.Name}");
    }
}
```

## 🤔 Expected Behavior

The FluentDataGrid should
- execute the HandleRowFocus method if i click on the row
- and also sort the rows if i click on one of the headers

## 😯 Current Behavior

The FluentDataGrid does
- execute the HandleRowFocus method if i click on the row ✅ 
- but does not sort the rows if i click on one of the headers ❌ 

## 💁 Possible Solution

If I remove the containing FluentGrid and FluentGridItem, everything works as expected.

## 🔦 Context

Usually you have a FluentGrid on every page to do the styling and eventually it contains a FluentDataGrid to display a large amount of data. So I think, this is a pretty common scenario.

## 🌍 Your Environment

Windows
.NET 8.0.2
Blazor Server
Google Chrome latest
FluentUI Blazor 4.4.1

Thank you very much 😄

vnbaaij:
Fix #1540 by only invoking RowFocus/CellFocus for default rows/cells (so not for (sticky) header, ColumnHeader and RowHeader)

### DIFF
--- a/src/Core/Components/NumberField/FluentNumberField.razor
+++ b/src/Core/Components/NumberField/FluentNumberField.razor
@@ -1,4 +1,4 @@
-@namespace Microsoft.FluentUI.AspNetCore.Components
+﻿@namespace Microsoft.FluentUI.AspNetCore.Components
 @inherits FluentInputBase<TValue>
 @typeparam TValue where TValue : new()
 @using System.Globalization;
@@ -24,7 +24,7 @@
                      max="@(ReadOnly ? value : Max)"
                      min="@(ReadOnly ? value : Min)"
                      id=@Id
-                     value="@value"
+                     current-value="@value"
                      disabled="@Disabled"
                      name=@Name
                      required="@Required"
--- a/examples/Demo/Shared/Microsoft.FluentUI.AspNetCore.Components.xml
+++ b/examples/Demo/Shared/Microsoft.FluentUI.AspNetCore.Components.xml
@@ -4706,7 +4706,7 @@
         <member name="P:Microsoft.FluentUI.AspNetCore.Components.FluentMenu.Trigger">
             <summary>
             Gets or sets the automatic trigger. See <seealso cref="T:Microsoft.FluentUI.AspNetCore.Components.MouseButton"/>
-            Possible values are None (default), Left, Middle, Right, Back, Forward 
+            Possible values are None (default), Left, Middle, Right, Back, Forward
             </summary>
         </member>
         <member name="P:Microsoft.FluentUI.AspNetCore.Components.FluentMenu.Open">
--- a/examples/Demo/Shared/Pages/Menu/Examples/MenuNested.razor
+++ b/examples/Demo/Shared/Pages/Menu/Examples/MenuNested.razor
@@ -14,7 +14,7 @@
             <FluentMenuItem>Menu item 2.3</FluentMenuItem>
         </MenuItems>
     </FluentMenuItem>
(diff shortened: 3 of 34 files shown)


src/Core/Components/NumberField/FluentNumberField.razor
```diff
@@ -24,7 +24,7 @@
                      max="@(ReadOnly ? value : Max)"
                      min="@(ReadOnly ? value : Min)"
                      id=@Id
-                     value="@value"
+                     current-value="@value"
```
dvoituron, line 27: Are you sure that we need to use `current-value` ?
I can't find this attribute in [this documentation](https://github.com/microsoft/fluentui/blob/master/packages/web-components/src/number-field/number-field.vscode.definition.json).

Have you checked if using a **EditForm** with `mode=post` sends the value correctly?

vnbaaij: This is what I seein the FAST Explorer and on https://fluent-components.azurewebsites.net/?path=/docs/components-number-field--number-field:
image

vnbaaij: The web components in general , the value attribute is only used to set the initial value of a component.
- `<fluent-number-field value="5" />` renders the component with 5 as a starting value
- `<fluent-number-field current-value="7" />` renders the component with 7 as a starting value

So both work for initial value but only current-value works for (re)setting the component with a @bind-Value from the Blazor side.

dvoituron (approved):

PR #1595, opened 2024-02-27
[FluentCard] Add MinimalStyle property

dvoituron:
# [FluentCard] Add MinimalStyle property

By default the **FluentCard** component is a visual container and design system provider. Cards are snapshots of content that are typically used in a group to present collections of related information. Thus, by design, the **FluentCard** component includes all the styles required for a good representation of the other components included in it. This adds a large number of styles such as `justify-content; align-items; gap; width;` and **over 80 CSS variables**. The code generated can therefore become very important, especially if you use this component several times, for example in a list or during a foreach.

To avoid this, you can use the **MinimalStyle** property. This will remove all the default styles from the **FluentCard** component. This will allow you to use the **FluentCard** component as a simple container without any style, except the styles to draw the box-shadow (and some minimal styles).

## Example

```xml
<FluentCard MinimalStyle="true">
    <p>Just some content in a card (with a button that does not do anything).</p>

    <FluentButton Appearance="Appearance.Accent">Hello</FluentButton>
    <p>No Width or Height has been specified.</p>
</FluentCard>
```
image

## Code generated

```html
<div class="fluent-card-minimal-style">
  <p>Just some content in a card (with a button that does not do anything).</p>
  <fluent-button type="button" appearance="accent" class="accent">Hello</fluent-button>
  <p>No Width or Height has been specified.</p>
</div>
```

## Default `FluentCard` code generated

image

### DIFF
--- a/src/Core/Components/Card/FluentCard.razor.css
+++ b/src/Core/Components/Card/FluentCard.razor.css
@@ -5,6 +5,7 @@ fluent-card ::deep {
 .fluent-card-minimal-style ::deep {
     height: var(--card-height, 100%);
     width: var(--card-width, 100%);
+    contain: content;
     background: var(--fill-color);
     color: var(--neutral-foreground-rest);
     border: calc(var(--stroke-width)* 1px) solid var(--neutral-stroke-layer-rest);
--- a/src/Core/Components/Card/FluentCard.razor.css
+++ b/src/Core/Components/Card/FluentCard.razor.css
@@ -1,3 +1,14 @@
 fluent-card ::deep {
     padding: calc(var(--design-unit) * 5px);
 }
+
+.fluent-card-minimal-style ::deep {
+    height: var(--card-height, 100%);
+    width: var(--card-width, 100%);
+    background: var(--fill-color);
+    color: var(--neutral-foreground-rest);
+    border: calc(var(--stroke-width)* 1px) solid var(--neutral-stroke-layer-rest);
+    border-radius: calc(var(--layer-corner-radius)* 1px);
+    box-shadow: var(--elevation-shadow-card-rest);
+    padding: calc(var(--design-unit)* 5px);
+}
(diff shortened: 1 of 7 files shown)


src/Core/Components/Card/FluentCard.razor.css
vnbaaij, line 1: Looking at the web components `card.styles.ts` file the following CSS is programmatically being set:
```
export const cardStyles: (
  context: ElementDefinitionContext,
  definition: FoundationElementDefinition,
) => ElementStyles = (context: ElementDefinitionContext, definition: FoundationElementDefinition) =>
  css`
    ${display('block')} :host {
      display: block;
      contain: content;
      height: var(--card-height, 100%);
      width: var(--card-width, 100%);
      box-sizing: border-box;
      background: ${fillColor};
      color: ${neutralForegroundRest};
      border: calc(${strokeWidth} * 1px) solid ${neutralStrokeLayerRest};
      border-radius: calc(${layerCornerRadius} * 1px);
      box-shadow: ${elevationShadowCardRest};
    }

    :host {
      content-visibility: auto;
    }
```

Maybe we also need to set the `display: block;` and `contain: content;`?

dvoituron: I've seen that... In my initial code, I added these line, but `contain` is new (see [here](https://developer.mozilla.org/en-US/docs/Web/CSS/contain) and not yet full supported by browsers.

And these two attributes are these to "apply some rules to sub-elements". I don't see why we should impose the format of the sub-elements. In our case, I think it's more about adding a simple border with a shading effect.

What do you think?

vnbaaij: I think `contain` is usefull because it allows devs to position elements in the card absolutely (as `contain: content;` creates a new containing block).  It will do so only if the browser supports it of course and otherrwise it will just be ignored.
Agree on the `display: block`. Lets not do that one.

dvoituron: Make sense. I will add this style.

vnbaaij (approved):

PR #1599, opened 2024-02-27
refactor: Migrate Demo Search to use FluentAutocomplete

Hona:
# Pull Request

## 📖 Description

- Moved from FluentSearch & a manual FluentMenu -> FluentAutocomplete
- Added Properties to FluentAutocomplete
    - ShowOverlayOnEmptyResults (default True to be non breaking).
    - ValueText - Allows setting the text inside the input
    - ValueTextChanged - Allows binding to the text changes inside the input
- Doesn't show the overlay if Items is null or empty

image
Figure: Default state

image
Figure: Keyboard navigable menu

image
Figure: Clears & navigates on enter press OR click

### General

### Component-specific


src/Core/Components/List/FluentAutocomplete.razor.cs
```diff
@@ -39,6 +37,15 @@ public FluentAutocomplete()
     [Parameter]
     public string? Placeholder { get; set; }
 
+    /// <summary>
+    /// Gets or sets the text field value.
+    /// </summary>
+    [Parameter]
+    public string ValueText { get; set; } = string.Empty;
+
+    [Parameter]
```
dvoituron, line 46: Could you add a comment?

src/Core/Components/List/FluentAutocomplete.razor.cs
```diff
@@ -216,13 +230,15 @@ protected async Task InputHandlerAsync(ChangeEventArgs e)
...
             Items = Items ?? Array.Empty<TOption>(),
-            Text = _valueText,
+            Text = ValueText,
         };
 
         await OnOptionsSearch.InvokeAsync(args);
 
-        Items = args.Items.Take(MaximumOptionsSearch);
-        SelectableItem = Items.FirstOrDefault();
+        Items = args.Items?.Take(MaximumOptionsSearch);
```
dvoituron, line 238: With this code, `args.Items` cannot be null:

```csharp
?? Array.Empty<TOption>()
```

dvoituron (changes requested):
I found this new bindable property `ValueText` very interesting. Thanks.

Could you add new Unit Tests to validate this change?

Could you update your PR description to be more explicit? Adding this new property, adding a screenshot/gif of the visual result, ...

examples/Demo/Shared/Shared/DemoSearch.razor
```diff
@@ -1,24 +1,17 @@
...
-            <FluentMenuItem OnClick="@(() => HandleSearchClicked(item))">
-                <span slot="start">
-                    <FluentIcon Value="@(item.Icon)" Class="search-result-icon" Color="Color.Neutral" Slot="start"/>
-                </span>
-                @item.Title
-            </FluentMenuItem>
-        }
-    </FluentMenu>
-}
+﻿<FluentAutocomplete TOption="NavItem"
```
dvoituron, line 1: You must "force" the width of the component to avoid the visual effect of "displacement".

peek

src/Core/Components/List/FluentAutocomplete.razor.cs
Hona: If the user supplied callback for OnOptionsSearch sets e.Items to null, then it will blow up (I ran into that exception while playing with this PR)

dvoituron: Oh... OK, if the user updates this property from `OnOptionsSearch.InvokeAsync`.
Make sense.

src/Core/Components/List/FluentAutocomplete.razor
```diff
@@ -14,11 +14,11 @@
                          AutoComplete="@AutoComplete"
                          Appearance="@Appearance"
                          Disabled="@Disabled"
-                         Placeholder="@(this.SelectedOptions?.Any() == false ? Placeholder : string.Empty)"
+                         Placeholder="@(SelectedOptions?.Any() is not true ? Placeholder : string.Empty)"
```
dvoituron, line 17: After checking Unit Tests, the result is not the same using `== false` and `is not true` when SelectedOptions is null (I think). In fact, the actual Unit Tests don't contain a `Placeholder` not defined, and with this change, the Unit Tests contains a string empty `Placeholder`.

It's probably better to render a `null` value using a code like this (to verify)
`Placeholder="@(SelectedOptions?.Any() is not true ? Placeholder : null)"`

Hona: I'll double check that logic in the unit test. When there were null VS .count == 0 selected items, it would show/hide the placeholder. I think the valid logic is null OR 0 items should show the placeholder. I'll update the test

examples/Demo/Shared/Shared/DemoSearch.razor
Hona: done

src/Core/Components/List/FluentAutocomplete.razor.cs
Hona: done

src/Core/Components/List/FluentAutocomplete.razor
Hona: Reverted the breaking change - looks like the placeholder shows fine now - probably related to the selected options changes

Hona:
I've got some unit tests to write - however the rest should be good to go

Hona:
Sorry - been a bit busy. Going to find some time to write these tests :)

dvoituron (approved):

Issue #1370, opened 2024-01-22
fix: Unable to expand menu if initially unexpanded in BlazorWasm in FluentNavMenu


c0g1t8:
# 🐛 Bug Report

Create a Blazor WASM application from fluent templates (v4.3.1). Configure the menu to initially start in unexpanded mode. Clicking the hamburger does not expand the menu in desktop mode.

 

## 💻 Repro or Code Sample

Change line 13 in **NavMenu.razor** file to:

```c#
    private bool expanded = false;
```
Run application. Behaviour can be seen in desktop mode. 

## 🤔 Expected Behavior

Ability to toggle between expanded and unexpanded mode.

## 😯 Current Behavior

Toggling menu does not work.

## 🌍 Your Environment

* OS & Device: Windows on PC
* Browser: Microsoft Edge, Google Chrome
* .NET and FAST Version: net8 and v4.3.1

c0g1t8:
Looking at the source of _FluentNavMenu.razor.js_

![image](https://github.com/microsoft/fluentui-blazor/assets/23250332/eedcdda3-5327-4d72-bfb1-f9d425671f59)

In line 6, **origStyle** is set when the menu is originally initialized. **origStyle** is unexpanded. **toggleMenuExpandedAsync** effectively changes nothing. 

I believe the fix would entail setting **origStyle** to the expanded style initially.

vnbaaij:
The functionality in the `FluentNavMenu.razor.js` script is only supposed to be called in an SSR scenario. Currently it gets called even when interactivity is available. I'm in touch with the Bazor team to investigate.

vnbaaij:
As said in #1387, there is nothing that can be done at the moment to change fact that menu uses the SSR script in a non SSR senario,
However, there have been some other chages to the scrip and component for v4.3.2 that makes you described scenario work so closing this one.
![issue-#1370](https://github.com/microsoft/fluentui-blazor/assets/1761079/e7cd4a45-249d-450f-85e7-4f44b4c77948)

PR #1633, opened 2024-03-05
fixes FluentPageScript with Blazor Hybrid/MAUI and WebAssembly Standalone

c0g1t8:
# Pull Request

## 📖 Description

Prevents listeners from being added when running in BlazorWebView
 
### 🎫 Issues

Event listeners are being added to collapse/expand button for FluentNavMenu when used in Blazor Hybrid/MAUI.

**Note**: See [comment](https://github.com/microsoft/fluentui-blazor/pull/1633#issuecomment-2002552015) `below for root cause`.

## 👩‍💻 Reviewer Notes

This is an addtional fix related to #1560 that [did not cover Blazor Hybrid/MAUI](https://github.com/microsoft/fluentui-blazor/pull/1560#issuecomment-1974307950).

The fix is based on determining if the component is running in BlazorWebView.

``` javascript
function isRunningInWebView() {
    return navigator.userAgentData.brands.find(e => e.brand == 'Microsoft Edge WebView2') != null;
}
```

## 📑 Test Plan

I ran all the tests that I used for #1560. I created a new Blazor Hybrid/MAUI application created from the default template and updated it with the Fluent components to match the defaults created for a server rendered webassembly application to verify the desired behaviour.

### General

### Component-specific


c0g1t8:
@dvoituron 

I was debating submitting this PR but decided to draft one so the behaviour would consistent across the different hosting/rendering options. I agree that this definitely 'smells'.

I've also discovered additional quirks related UI state not being preserved when transitioning from static rendered to Blazor rendered.
- If you collapse the menu as soon as the UI appears, it will suddenly expand when Blazor rendering/interactivity is available. 
- There's actually an additional transition when the interactivity option is WebAssembly. Once for the server static rendering, again when WebAssembly starts the page, and finally when completely running in WebAssembly.

This also violates this [note in the documentation](https://learn.microsoft.com/en-us/aspnet/core/blazor/components/render-modes?view=aspnetcore-8.0#apply-a-render-mode-to-a-component-definition) regarding not coupling to render mode.

All of this to provide the collapsing and expanding functionality of the FluentNavMenu when statically rendered.

As for the _source of problem_, I'm still trying to figure that out. :smile:. I've learned a lot about the rendering process as I've looked at this issue. I've been unable to find any detailed documentation on the transition from static rendered to interactive rendered.

I've also considered but not tried:
- Removing the event listener added in JavaScript in the **OnInitializedAsync**. This is likely going to exhibit the quirky behaviour I mentioned above as well.
- Moving the desired behvaiour into Javascript only so that the collapse and expand behaviour is only handled in JavaScript. This would be my next candidate to explore more thoroughly.

c0g1t8:
I've been doing some further research looking for a root cause as per Denis's comments.

I found https://github.com/microsoft/fluentui-blazor/blob/3f9a21f4f26e37a400a30ad462efb89a5e7960ac/src/Core/Components/NavMenu/FluentNavGroup.razor#L39

**Data** is a property inherited from **Microsoft.FluentUI.AspNetCore.Components.FluentComponentBase**. It is being used to prevent the script from being run https://github.com/microsoft/fluentui-blazor/blob/3f9a21f4f26e37a400a30ad462efb89a5e7960ac/src/Core/Components/NavMenu/FluentNavMenu.razor#L3-L6

An alternative to making any code changes to the **FluentUI-Blazor** library would be to explicity do this in the **NavMenu** component on a per application basis for **Blazor Hybrid/MAUI** and **WebAssembly hosted** applications.

@dvoituron @vnbaaij Thoughts??

Note: This still does not address the UI state transition issues. See https://github.com/microsoft/fluentui-blazor/pull/1633#issuecomment-1979332854

vnbaaij:
I'm using the `Data` parameter to prevent the script from being included multiple times when the menu hase a couple of layers. It does not prevent the script form being run.

c0g1t8:
> I'm using the `Data` parameter to prevent the script from being included multiple times when the menu hase a couple of layers. It does not prevent the script form being run.

That makes sense. I forgot to include the following screenshot above😞.

image

I made this change to the NavMenu component in the **Blazor Hybrid/MAUI** and **WebAssembly hosted** test applications and it effectively gave me the results I wanted. The javascript listeners were not added. Both these applications defaulted to using the Blazor provided event handler.

It's basically a temporary work-around for 4.4.1. I'm currently using it in my production **WebAssembly hosted** application where I originally ran across this issue. 😄

I'm still looking into a better long term fix that would also address UI state change issue.

dvoituron:
> An alternative to making any code changes to the FluentUI-Blazor library would be to explicity do this in the NavMenu component on a per application basis for Blazor Hybrid/MAUI and WebAssembly hosted applications.

Indeed, I prefer to have an explicit property (for example by adding an option to `AddFluentUIComponents`) and if it's possible to define a default value using the detected environment (MAUI, Web, ...).

c0g1t8:
> > An alternative to making any code changes to the FluentUI-Blazor library would be to explicity do this in the NavMenu component on a per application basis for Blazor Hybrid/MAUI and WebAssembly hosted applications.
> 
> Indeed, I prefer to have an explicit property (for example by adding an option to `AddFluentUIComponents`) and if it's possible to define a default value using the detected environment (MAUI, Web, ...).

@dvoituron I think the explicit declaration is the way to go.  

As I reflect on how this started and on @vnbaaij [comment](https://github.com/microsoft/fluentui-blazor/issues/1370#issuecomment-1917962347), [Enhanced navigation](https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?view=aspnetcore-8.0#enhanced-navigation-and-form-handling) should not have applied. I think this is a Blazor bug.

When a Blazor Web App is first pre-rendered/statically rendered, the [javascript for static SSR](https://learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/static-server-rendering?view=aspnetcore-8.0) is sent to the browser where it's executed on the **enhancedload** event. When the application is ready, another render occurs but this time there is no **enhancedload** event.

In Blazor Hybrid/MAUI and WebAssembly hosted applications, no pre-rerendering is done but the **enhancedload** event is still being triggered on first render. Preventing the **enhancedload** event from occuring would make the component behaviour consistent no matter the hosting environment.

If this is truly a bug, by explicitly declaring it would make it easier to deprecate in the future when it's fixed. 

Assuming that using an explicit property as the approach, how should it be implemented? 

I see two options:
1. Add an option called '**UseStaticSsr**' with a default value of **true** for **AddFluentUIComponents**. **UseStaticSsr** would need to be explicitly set to false for Blazor Hybrid/MAUI and WebAssembly hosted applications. Add logic that conditionally checks for this option before including **fluent-page-script** in the _FluentPageScript.razor_ file. 
2. Add a property to **FluentNavMenu** named **UseStaticSsr** with a default value of **true**. Then conditionally check this in _FluentNavMenu.razor_.

Note:
- Currently only the **FluentNavMenu** component is using the **FluentPageScript** component that adds the javascript for static SSR. Adding an option to **AddFluentUIComponents** could be considered overkill. 
- As for detecting the environment, the only method I've found so far is in javascript so far. 

I can see pros and cons for either option. Thoughts?

vnbaaij:
> @dvoituron I think the explicit declaration is the way to go.
> 
> As I reflect on how this started and on @vnbaaij [comment](https://github.com/microsoft/fluentui-blazor/issues/1370#issuecomment-1917962347), [Enhanced navigation](https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?view=aspnetcore-8.0#enhanced-navigation-and-form-handling) should not have applied. I think this is a Blazor bug.
> 
> When a Blazor Web App is first pre-rendered/statically rendered, the [javascript for static SSR](https://learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/static-server-rendering?view=aspnetcore-8.0) is sent to the browser where it's executed on the **enhancedload** event. When the application is ready, another render occurs but this time there is no **enhancedload** event.
> 
> In Blazor Hybrid/MAUI and WebAssembly hosted applications, no pre-rerendering is done but the **enhancedload** event is still being triggered on first render. Preventing the **enhancedload** event from occuring would make the component behaviour consistent no matter the hosting environment.

Yes, that is the ultimate goal. 

 
> If this is truly a bug, by explicitly declaring it would make it easier to deprecate in the future when it's fixed.
> 
> Assuming that using an explicit property as the approach, how should it be implemented?
> 
> I see two options:
> 
> 1. Add an option called '**UseStaticSsr**' with a default value of **true** for **AddFluentUIComponents**. **UseStaticSsr** would need to be explicitly set to false for Blazor Hybrid/MAUI and WebAssembly hosted applications. Add logic that conditionally checks for this option before including **fluent-page-script** in the _FluentPageScript.razor_ file.

I think this is the way to go. 

> 2. Add a property to **FluentNavMenu** named **UseStaticSsr** with a default value of **true**. Then conditionally check this in _FluentNavMenu.razor_.

This seems like a too specific solution for me.

 
> Note:
> 
> * Currently only the **FluentNavMenu** component is using the **FluentPageScript** component that adds the javascript for static SSR. Adding an option to **AddFluentUIComponents** could be considered overkill.

The NavMenu is the only place I've added it _on purpose_. I anticipate that we would like to have more components that offer enhanced functionality (other than just displaying) when used in a SSR scenario. Because of the issues this is giving, I have not yet looked at which components we could use this. If we do, I expect we will be using the `FluentPageScript` again.

> * As for detecting the environment, the only method I've found so far is in javascript so far.

I believe extra/new/better functionality is on the roadmap. I'll sk the Blazor team what is the est way to do this now.

c0g1t8:
### Background

I thought the root cause of the issue was that the enhancedload event was the cause of this issue so I filed an [issue](https://github.com/dotnet/aspnetcore/issues/54463). After some feedback from @MackinnonBuck and attempting to create a minimal repro, I discovered a difference between how **PageScript** and **FluentPageScript** was implemented. The latter was a pretty faithful implementation in Typescript. The primary difference is was when the handler was added to the **enhancedload** event.

In **PageScript** we have in [BlazorPageScript.lib.module.js](https://github.com/MackinnonBuck/blazor-page-script/blob/main/src/wwwroot/BlazorPageScript.lib.module.js):

```javascript
export function afterWebStarted(blazor) {
    customElements.define('page-script', class extends HTMLElement {
        static observedAttributes = ['src'];


        // We use attributeChangedCallback instead of connectedCallback
        // because a page-script element might get reused between enhanced
        // navigations.
        attributeChangedCallback(name, oldValue, newValue) {
            if (name !== 'src') {
                return;
            }


            this.src = newValue;
            unregisterPageScriptElement(oldValue);
            registerPageScriptElement(newValue);
        }


        disconnectedCallback() {
            unregisterPageScriptElement(this.src);
        }
    });


    blazor.addEventListener('enhancedload', onEnhancedLoad);
}
```

In FluentPageScript, this initialization is done in the [index.ts](https://github.com/microsoft/fluentui-blazor/blob/dev/src/Core.Assets/src/index.ts) file.

https://github.com/microsoft/fluentui-blazor/blob/072a9388e9bdb6dfe83573e14c45a786c13f04ad/src/Core.Assets/src/index.ts#L297-L299

and

https://github.com/microsoft/fluentui-blazor/blob/072a9388e9bdb6dfe83573e14c45a786c13f04ad/src/Core.Assets/src/index.ts#L304-L310

### Why is happening?

customElement **fluent-page-script** is always defined in [index.ts](https://github.com/microsoft/fluentui-blazor/blob/dev/src/Core.Assets/src/index.ts). This means that _attributeChangedCallback_ in [FluentPageScript.ts](https://github.com/microsoft/fluentui-blazor/blob/dev/src/Core.Assets/src/FluentPageScript.ts) is called in for any fluent-page-script tag present.  This in turns calls _registerPageScriptElement_ that calls _initializePageScriptModule_ which then calls the _onLoad_ in [FluentNavMenu.razor.js](https://github.com/microsoft/fluentui-blazor/blob/dev/src/Core/Components/NavMenu/FluentNavMenu.razor.js).

_onLoad_ is triggered by _attributeChangedCallback_  **NOT** by the **enhancedload** event. And because the customElement **fluent-page-script** is always defined. The Javascript handlers were always added to the toggle.

### Fix

The fix is to move line 307 to after line  297.

image

This fix also addresses the root cause of #1560 as well.

c0g1t8:
@vnbaaij 

>The NavMenu is the only place I've added it on purpose. I anticipate that we would like to have more components that offer enhanced functionality (other than just displaying) when used in a SSR scenario. Because of the issues this is giving, I have not yet looked at which components we could use this. If we do, I expect we will be using the FluentPageScript again.

vnbaaij (approved):

vnbaaij:
Thanks Gary for perservering in finding the root cause!

Issue #1621, opened 2024-03-02
fix: High CPU usage in FluentDataGrid


JamesNK:
# 🐛 Bug Report

I stress tested Aspire and had a grid with many rows that updated once a second. I noticed very high CPU usage. I profiled and found uncached usage of `Expression.Compile()`.

Profile:
![image](https://github.com/microsoft/fluentui-blazor/assets/303201/8f716eee-d20c-463b-a88b-0fd107cffacd)

Likely offender:
https://github.com/microsoft/fluentui-blazor/blob/4516f9567354f6d721ac9f78c5ec5e055b7868ef/src/Core/Components/DataGrid/Columns/TemplateColumn.cs#L26-L27

Which is called from `RenderRow`:
https://github.com/microsoft/fluentui-blazor/blob/4516f9567354f6d721ac9f78c5ec5e055b7868ef/src/Core/Components/DataGrid/FluentDataGrid.razor#L89-L106

## 💻 Repro or Code Sample

Create a big grid with tooltips. Have it update often.

## 🤔 Expected Behavior

Renders quickly.

## 😯 Current Behavior

Very high CPU usage.

## 💁 Possible Solution

Cache compiled expression.

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* .NET and Fluent UI Blazor library Version [e.g. 8.0.2 and 4.4.1]

JamesNK:
Can you reopen this? I think the current solution will create a dictionary for every combination of row type and property type. It would probably be better to have one dictionary for all expressions, and then cast the returned type. It's not as strongly typed, but will return memory usage.

Also, there are some other places where there is Expression.Compile that might need work to improve caching:

https://github.com/microsoft/fluentui-blazor/blob/902bcc8448063bdfcd84ec1fbc9188b7633e9bc3/src/Core/Components/DataGrid/Columns/PropertyColumn.cs#L53-L102

Some research is probably needed into the best way to cache expressions. This seems like a common problem so there will be people who have solved it before.

vnbaaij:
Expression and Compile logic has been removed from the TooltipText parameter and handling.

PR #1635, opened 2024-03-05
Remove Expression from TooltipText

vnbaaij:
Both `FluentDataGrid`'s `PropertyColumn` and `TemplateColumn` have a `TooltipText` parameter. Based on other code in the grid, this parameter was using an `Expression<Func>>` for its type. THis is unnecessary and lead to performance issues/CPU spikes (see #1621). 
With this PR the `Expression` part is removed.


dvoituron (approved):
I like it... simpler :-)

Issue #1638, opened 2024-03-05
fix: Autocomplete / Combobox hangs if many items are bound


martinbu:
If many items are bound to a FluentAutocomplete or FluentCombobox all items gets rendered. In the newest dev branch the items are even "rendered" before the combobox is opened.
This causes - depending on the items count - a lag of multiple seconds when opening the page or in a previous version when opening the combobox (list).

Is there a possibility to only render those items that are visible like the virtualization in the FluentDataGrid?

An example that reproduces this issue with the FluentCombobox can be found [here](https://github.com/martinbu/fluentui-blazor/commit/e2cd1ac0eca210d51e811fa3dc7dd4a91ea068c0).

vnbaaij:
No, we can't do virtualuzation here as the (Fluent) options does not support that.

How many items are we talking about?

martinbu:
At the moment I have a list with about 800 items. But because these controls have the possibilities to search/filter the bound items count could be even much much longer for other use cases...

martinbu:
That is true, but if no search text is specified I want to "preserve" the possibility to scroll through the complete list. A user that does not know that he/she can search with the keyboard will then only see e.g. the first 10 items which can be confusing.

Additionally, the FluentAutocomplete has at the moment no possibility to "look like" a e.g. FluentCombobox with a simple one item selected view. On the other hand I cannot use a FluentCombobox because it has insufficient search capabilities like stated here #1639

martinbu:
When I look at the two Components FluentAutoComplete and FluentCombobox I think it is much easier to add the functionalities
1. List Virtualization
2. Single Select option --> Autocomplete should look like a Combobox in single select mode

The virtualization part could be implemented with [ASP.NET Core Razor component virtualization](https://learn.microsoft.com/en-us/aspnet/core/blazor/components/virtualization?view=aspnetcore-8.0) because the options are rendered with a foreach loop in the FluentAutocomplete.razor page.

For the "look like a combobox" part, an addtional option could be added which changes the appearance.

What do you think?

dvoituron:
The FluentCombobox is a functionally similar representation of the same HTML component. The data to be searched must be present in the HTML page already loaded. This is the main reason why we've added the FluentAutocomplete, which allows you to launch a search on the "server" to obtain some results.
The constraint is that you have to start a search to display data.

[...]

PR #1647, opened 2024-03-06
Virtualization for FluentAutocomplete

martinbu:
This PR implements the virtualization feature for FluentAutocomplete discussed here #1638 

For me the implementation works fine and greatly improves the initial rendering performance and the reactivity when a search is done on that list. The improvement can be checked by changing the Virtualize Parameter in the Autocomplete with Many Items Demo example.

Additionally, if the MaximumOptionsSearch is set to a value smaller than 0, it is disabled. With this change the OnOptionsSearch implementer can decide on their own how many items should be displayed. E.g. when a search text is active the displayed items can be reduced to n results. But when e.g. no Search text is active all items can be shown by default.
This is a breaking change:
Before: MaximumOptionsSearch < 0 --> empty list is rendered
After: MaximumOptionsSearch < 0 --> all items are rendered

### DIFF
--- a/src/Core/Components/List/FluentAutocomplete.razor.cs
+++ b/src/Core/Components/List/FluentAutocomplete.razor.cs
@@ -112,7 +112,6 @@ public partial class FluentAutocomplete<TOption> : ListComponentBase<TOption> wh
 
     /// <summary>
     /// Gets or sets the number of maximum options (items) returned by <see cref="OnOptionsSearch"/>.
-    /// If the value is smaller than 0, all search results are shown.
     /// Default value is 9.
     /// </summary>
     [Parameter]
@@ -188,8 +187,7 @@ public partial class FluentAutocomplete<TOption> : ListComponentBase<TOption> wh
     /// If you use <see cref="Virtualize"/>, you should supply a value for <see cref="ItemSize"/> and must
     /// ensure that every row renders with the same constant height.
     ///
-    /// Generally it's preferable not to use <see cref="Virtualize"/> if the amount of data being rendered
-    /// is small or if you are using pagination.
+    /// Generally it's preferable not to use <see cref="Virtualize"/> if the amount of data being rendered is small.
     /// </summary>
     [Parameter] public bool Virtualize { get; set; }
 
@@ -267,14 +265,7 @@ public partial class FluentAutocomplete<TOption> : ListComponentBase<TOption> wh
 
         await OnOptionsSearch.InvokeAsync(args);
 
-        if (MaximumOptionsSearch < 0)
-        {
-            Items = args.Items;
-        }
-        else
-        {
-            Items = args.Items?.Take(MaximumOptionsSearch);
-        }
+        Items = args.Items?.Take(MaximumOptionsSearch);
 
         SelectableItem = Items != null
             ? Items.FirstOrDefault()
--- a/src/Core/Components/List/FluentAutocomplete.razor.cs
+++ b/src/Core/Components/List/FluentAutocomplete.razor.cs
@@ -110,6 +110,7 @@ public partial class FluentAutocomplete<TOption> : ListComponentBase<TOption> wh
 
     /// <summary>
     /// Gets or sets the number of maximum options (items) returned by <see cref="OnOptionsSearch"/>.
+    /// If the value is smaller than 0, all search results are shown.
     /// Default value is 9.
     /// </summary>
     [Parameter]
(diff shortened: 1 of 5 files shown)


src/Core/Components/List/FluentAutocomplete.razor.cs
```diff
@@ -104,6 +106,7 @@ public override bool Multiple
 
     /// <summary>
     /// Gets or sets the number of maximum options (items) returned by <see cref="OnOptionsSearch"/>.
+    /// If the value is smaller than 0, all search results are shown.
```
vnbaaij, line 109: I don't like this 'magic parameter' where you use a negative number to force a certain behavior. Think it is better to introduce a separate (new) parameter for that.

src/Core/Components/List/FluentAutocomplete.razor.cs
```diff
@@ -171,6 +174,26 @@ public override bool Multiple
...
+    /// <summary>
+    /// If true, the options list will be rendered with virtualization. This is normally used in conjunction with
+    /// scrolling and causes the option list to fetch and render only the data around the current scroll viewport.
+    /// This can greatly improve the performance when scrolling through large data sets.
+    ///
+    /// If you use <see cref="Virtualize"/>, you should supply a value for <see cref="ItemSize"/> and must
+    /// ensure that every row renders with the same constant height.
+    ///
+    /// Generally it's preferable not to use <see cref="Virtualize"/> if the amount of data being rendered
+    /// is small or if you are using pagination.
```
vnbaaij, line 186: We don't offer support for pagination in the search result list so please change this line.

People might think it is supported otherwise

vnbaaij (changes requested):

dvoituron (dismissed):

src/Core/Components/List/FluentAutocomplete.razor.cs
vnbaaij: Would it not be slower? Component needs to be invoked, do some calculations and then come to the conclusion nothing needs to be done.

dvoituron: To validate, but I don't think so. I think that only `LoadFilteredItemsAsync` will be call to get items.

martinbu: I introduced the _Virtualize_ Parameter because FluentDataGrid also has one. And in the comment of the FluentDataGrid's parameter it is [mentioned](https://github.com/microsoft/fluentui-blazor/blob/dev/src/Core/Components/DataGrid/FluentDataGrid.razor.cs#L43) that:
"Generally it's preferable not to use <see cref="Virtualize"/> if the amount of data being rendered is small or if you are using pagination."

Therefore, I thought that it is better to let the user decide because there may be performance drawbacks if using Virtualization with only a few items.

vnbaaij: Yeah, I like that better too. Make it a consious choice made by the developer. I propose to not make it always use `Virlualize` (ie leave it as you have it now).

Remark about the `magic parameter` still standss though...

martinbu: Which solution is prefered by you, regarding the MaximumOptionsSearch parameter
1. New Parameter for disabling this option
2. Remove -1 logic and just use e.g. int.Max

I think both are better than the actual implemented behaviour in the PR because then we do not have a breaking change.

vnbaaij: Can't we just leave as is an in case of `Virtualize` being used ignore the `MaximumOptionSearch` value?

martinbu: Regarding always Virtualize or not. If the Virtualize parameter is introduced, we can be sure that no different behaviour is introduced for projects that use the AutoComplete component. So everyone have to switch manually if needed.

Different behaviour: This can for example be the case if items with a different height are in place. The Virtualize component sets the item height to a fixed height of 50px by default. I toke over this default value to the AutoComplete ItemsHeight parameter.

vnbaaij: > Regarding always Virtualize or not. If the Virtualize parameter is introduced, we can be sure that no different behaviour is introduced for projects that use the AutoComplete component. So everyone have to switch manually if needed.

Agreed. Lets not make it do Virtualize always

> Different behaviour: This can for example be the case if items with a different height are in place. The Virtualize component sets the item height to a fixed height of 50px by default. I took over this default value to the AutoComplete ItemsHeight parameter.

Default is good, I think. Believe it is at 32 for the DataGrid.

vnbaaij: Martin, I'm preparing 4.5 release right now. Can we get this finished and merged in today?

vnbaaij: You mean default stays at 9 and dev can set it to int.Max if they want everything? If os, yes that seems like a good approach.

martinbu: Exactly.

vnbaaij (approved):

PR #1657, opened 2024-03-08
IMessageService enhancements to support the 'AllowDismiss' property

StevenRasmussen:
# Pull Request

## 📖 Description

Building off of the recently released version which added the ability to prevent dismissing of the message bar notification in the `MessageBar` component, this PR adds overloads to the `IMessageService` which can be used for adding message bar components to a `MessageBarProvider` instance.

### General

### Component-specific


src/Core/Components/MessageBar/Services/IMessageService.cs
```diff
@@ -16,25 +16,43 @@ public interface IMessageService : IDisposable
     Message ShowMessageBar(Action<MessageOptions> options);
 
     /// <summary />
-    Message ShowMessageBar(string message);
+    Message ShowMessageBar(string title);
 
     /// <summary />
-    Message ShowMessageBar(string message, MessageIntent intent);
+    Message ShowMessageBar(string title, bool allowDismiss);
```
dvoituron, line 22: I don't think we'd like to make it so easy to remove the close icon.

In FluentUI, it's always preferable for the users to be able to close a message if they wish.

However, adding the option in MessageOptions (only) is still a good idea, I think.

src/Core/Components/MessageBar/Services/IMessageService.cs
```diff
@@ -16,25 +16,43 @@ public interface IMessageService : IDisposable
...
 
     /// <summary />
-    Message ShowMessageBar(string message, MessageIntent intent, string section);
+    Message ShowMessageBar(string title, MessageIntent intent);
```
dvoituron, line 25: Why rename `message` to `title`?

dvoituron (changes requested):

src/Core/Components/MessageBar/Services/IMessageService.cs
StevenRasmussen: Ok.. I will remove the overloads.

src/Core/Components/MessageBar/Services/IMessageService.cs
StevenRasmussen: The implementation (`MessageService`) had the parameter named `title` so I was merely updating it for consistency.  If the better name is `message` then I will update everything to be `message.

StevenRasmussen: It appears that the underlying `Options` object has a `Title` property on it that the `message` parameter is being mapped to... perhaps that's why the parameter was named `title` on the implementation side.  Either way, I think the interface and the implementation should have the same name.  What would you like that to be?

vnbaaij: I want to follow the design guidelines as much as possible:
image

We are already streching that with th option to not have the close button

Going by what we are doing in the example 'Add on top',
image

I would say that that is showing the Title (number 3 from the design image). So I think `MessageService.cs` is correct and `IMessageService.cs` needs to be changed. Agree?

dvoituron: Yes. Correct

src/Core/Components/MessageBar/Services/IMessageService.cs
StevenRasmussen: Removed

src/Core/Components/MessageBar/Services/IMessageService.cs
StevenRasmussen: Both the `IMessageService` and the `MessageService` implementation use `title` now.

src/Core/Components/MessageBar/Services/IMessageService.cs
dvoituron: I think the arguments are still in the `MessageService` class, whereas they are no longer in `IMessageService`.

StevenRasmussen: Good catch!  Removed.

dvoituron (approved):

Issue #1662, opened 2024-03-08
rfc: add the ability for the Wizard component to automatically validate any Blazor form components


StevenRasmussen:
# 💬 RFC

It would be nice if the Wizard component was able to automatically validate any Blazor form components before allowing to move to the next page.

## 🔦 Context

I am wanting to use the wizard control to walk a user through completing some forms.  Currently this is a bit awkward to do and it would be ideal if the Wizard component would automatically detect that a form was present and validate the form before proceeding to the next page. 

I have some thoughts on how to accomplish this and have already put together a POC that I think will work but I wanted to get feedback on whether something like this would be accepted into the project before putting forth the effort of creating a PR.

vnbaaij:
Think that could be definitely bean interesting addition. Certainly not against it. As long as it can be added as a non breaking feature...

dvoituron:
This is an interesting feature. There are several scenarios to consider:
1. By default, don't activate it (as it is now)
2. Validate before moving on to the next page (**Next** button)
2. Validate at end of Wizard only (**Finish** button). Because it's probably also a case where the user completes several steps, can go back and validate at the end of the process.

AlthalusDGr8:
Having validation at the end of the process might prove a bit of a challenge on how best to indicate which field in which step has issues.

AlthalusDGr8:
PS the way I am currently handling this is as follows in case someone needs a quick solution:
Razor:
```
 <FluentWizardStep Label="Details" OnChange="@FurtherValidationAndProceed">
     <EditForm FormName="reg-new" EditContext="editContext">
         <DataAnnotationsValidator />
         <FluentValidationSummary />
         
         <FluentTextField Label="Tenant Name" Required="true" @bind-Value="RegistrationApplication.TenantName" />
         <FluentValidationMessage For="@(() => RegistrationApplication.TenantName)" />

         <FluentTextField Label="Admin Email" Required="true" @bind-Value="RegistrationApplication.RegistrationEmailAddress" Name="Model.EmailAddress" TextFieldType="TextFieldType.Email" />
         <FluentValidationMessage For="@(() => RegistrationApplication.RegistrationEmailAddress)" />

     </EditForm>
 </FluentWizardStep>
 ```
Code Section:
```
private async Task FurtherValidationAndProceed(FluentWizardStepChangeEventArgs e)
{        
    // first do model checks, if model is invalid then do not proceed
    if (!editContext.Validate())
    {
[...]

PR #1663, opened 2024-03-08
Added the ability for the Wizard component to automatically validate an EditForm

StevenRasmussen:
In reference to #1662.  I have not added any unit tests yet.  I wanted to seek feedback before trying to add those.  Running one of the demo projects and navigating to the Wizard page will show you how it works.

# Pull Request

## 📖 Description

Adds the ability to the Wizard to automatically validate an EditForm when trying to navigate to the 'Next' page or complete the wizard.

### General

### Component-specific

- [x] I have validated [Unit Tests](https://github.com/Microsoft/fluentui-blazor/blob/master/unit-tests.md) for an existing component


vnbaaij:
Looks good (only quick first look taken)

examples/Demo/Shared/Pages/Wizard/Examples/WizardEditForms.razor
```diff
@@ -0,0 +1,123 @@
...
+              StepTitleHiddenWhen="@GridItemHidden.XsAndDown"
+              Height="auto"
+              Style="min-height: 300px;"
+              OnFinish="@OnFinishedAsync">
+    <Steps>
+        <FluentWizardStep Label="Personal Info"
+                          OnChange="@OnStepChange">
+            <EditForm Model="_formData1" FormName="personalInfo" method="post">
+                <DataAnnotationsValidator />
+                <FluentWizardStepFormValidator />
```
dvoituron, line 19: What is the goal of the `FluentWizardStepFormValidator`?
Apart from registering the EditContext with the WizardStep, which is more a "technical" reason.

In other words, what will be the result if I use `EditForm` without `FluentWizardStepFormValidator`? The `DataAnnotationsValidator` will not be enough?

Is it the single possibility?

dvoituron (changes requested):

examples/Demo/Shared/Pages/Wizard/Examples/WizardEditForms.razor
StevenRasmussen: The `FluentWizardStepFormValidator` is what enables the `WizardPage` to become aware of the edit context so that it can call the `EditContext.Validate()` method before moving to the next page.  If you leave this out then the default behavior is maintained.  If you are aware of a better way to inform the `WizardPage` that it contains an `EditForm` with an `EditContext` then let me know.

dvoituron: I don't know 😊
I'd like to know if this is the only solution.

It would be easier not to have to add `DataAnnotationsValidator` and also `FluentWizardStepFormValidator`. But the question is indeed how to retrieve EditContext 😇

I can check this weekend

dvoituron: I'd like to avoid adding something too quickly, without knowing if it's possible to make it simpler.

StevenRasmussen: Agreed.  The only other way I know of is to instantiate your own `EditContext` higher up the chain and then pass it down to the different components via a parameter (either explicit or cascading).  But I'll let you come back after the weekend to confirm if that's the case.  Also, the `<DataAnnotationsValidator>` is not required per se by this change, but is rather the default/best practice for validating a Blazor `EditForm` when using data annotations.  For more info see here: [https://learn.microsoft.com/en-us/aspnet/core/blazor/forms/?view=aspnetcore-8.0](https://learn.microsoft.com/en-us/aspnet/core/blazor/forms/?view=aspnetcore-8.0)

I personally have created my own higher-level component called `<FluentForm>` which encapsulates an `EditForm` and includes a lot of the defaults for editing forms and I include the `<DataAnnotationsValidator>` as well.  I would also place the `FluentWizardStepFormValidator` in it so that I don't need to add it everywhere I use an `EditForm`.

dvoituron: It might be another (and good) idea to add a `FluentEditForm`. It might simplify this and other (future or existing) features. @vnbaaij  what is your feeling about EditForm?

Give me a few days to do some tests. Tomorrow is my son's birthday, but I'll try to find some time on Sunday :-)

src/Core/Components/FluentEditForm/FluentEditForm.razor.cs
StevenRasmussen, line 1: I tried for some time to have this just inherit from the `EditForm` directly so that we would not have to duplicate the properties and pass them down to the child `EditForm` component but was unsuccessful. If we can make it work with inheritance that would be ideal.  Definitely open to ideas on how to do that.

dvoituron: I haven't had time to do much code analysis yet, but why not do this?

1. Remove the file `FluentWizardStepFormValidator.cs` 
1. In `FluentWizardStep.razor`, wrap the code in `<CascadingValue Value="this" IsFixed="true">`.
1. Replace the **FluentEditForm** component with this code.

```csharp
using Microsoft.AspNetCore.Components.Forms;
using Microsoft.AspNetCore.Components.Rendering;

namespace Microsoft.FluentUI.AspNetCore.Components;
public class FluentEditForm : EditForm
{
    [CascadingParameter]
    public FluentWizardStep? WizardStep { get; set; }

    protected override void BuildRenderTree(RenderTreeBuilder builder)
    {
        base.BuildRenderTree(builder);

        if (WizardStep is not null && EditContext is not null)
        {
            WizardStep.RegisterEditFormAndContext(this, EditContext);
        }        
    }
}
```

StevenRasmussen: Brilliant!  After adding the `FluentEditForm` component, I hadn't realized that we could implement the same logic that the `FluentWizardStepFormValidator` was performing in the `FluentEditForm`.  That being said, I think it might still be useful to keep the `FluentWizardStepFormValidator` component for those that would rather continue to use the OOB `EditForm`.  I don't see any downside and it allows some flexibility for those that want to continue to use the `EditForm` component.

vnbaaij: Although good for this outpost, we don't want to have this component added. This one is for the wizard, next week devs want one for tabs, etc. With the adapted EditForm we have one consistent approach. Sound good?

dvoituron (changes requested):

src/Core/Components/FluentEditForm/FluentEditForm.razor.cs
StevenRasmussen: Done

src/Core/Components/Wizard/FluentWizardStep.razor.cs
```diff
@@ -134,4 +135,51 @@ protected override void OnInitialized()
         Index = FluentWizard.AddStep(this);
         base.OnInitialized();
     }
+
+    private Dictionary<EditForm, EditContext> _editForms = new Dictionary<EditForm, EditContext>();
```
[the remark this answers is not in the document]
StevenRasmussen: Done

src/Core/Components/Wizard/FluentWizardStep.razor.cs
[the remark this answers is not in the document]
StevenRasmussen: I added some unit test and verified the code coverage. It appears that all of my changes are now covered.  Let me know if there's anything else I need to do on this PR. Thanks!

dvoituron (approved):
@StevenRasmussen Many thanks. Approved.

PR #1705, opened 2024-03-16
[FluentProfileMenu] New component

dvoituron:
# [FluentProfileMenu] New component

A "User Profile Menu" is a component commonly found on websites or applications. It typically appears in the top-right corner of a web page and provides options related to the user's account. Within this menu, users can access features such as viewing their account details, adjusting settings, and logging out. It serves as a convenient hub for managing account-related actions.

peek

```xml
<FluentProfileMenu  Image="@DataSource.SamplePicture"
                    Status="@PresenceStatus.Available"
                    HeaderLabel="Microsoft"
                    Initials="BG"
                    FullName="Bill Gates"
                    EMail="bill.gates@microsoft.com"
                    Style="min-width: 330px;" />
```

## Unit Tests

image


src/Core/Components/ProfileMenu/FluentProfileMenu.razor.cs
```diff
@@ -0,0 +1,117 @@
...
+    /// Gets or sets the name to display.
+    /// </summary>
+    [Parameter]
+    public string? FullName { get; set; }
+
+    /// <summary>
+    /// Gets or sets the company name to display.
+    /// </summary>
+    [Parameter]
+    public string? CompanyName { get; set; }
```
vnbaaij, line 68: Should this perhaps be a bit more general like `Details`?

src/Core/Components/List/FluentPersona.razor
```diff
@@ -1,8 +1,8 @@
 ﻿@namespace Microsoft.FluentUI.AspNetCore.Components
 @inherits FluentComponentBase
 
-<div id="@Id" class="@ClassValue" style="@StyleValue">
-    <div class="initials">
+<div id="@Id" class="@ClassValue" style="@StyleValue" @attributes="@AdditionalAttributes">
+    <div class="initials" @onclick="@OnClick" style="@(OnClick.HasDelegate ? $"cursor: pointer;" : null)">
         <FluentPresenceBadge Status="@Status" Size="@StatusSize" StatusTitle="@StatusTitle">
```
vnbaaij, line 6: Should there be a check on not rendering the PresenceBadge when `Status` is `null`?

vnbaaij (changes requested):

src/Core/Components/List/FluentPersona.razor
vnbaaij: Ive added this commen in #1710  as well. Think it belongs there more.

src/Core/Components/ProfileMenu/FluentProfileMenu.razor.cs
dvoituron: Use an object? Yes, that makes sense. It could be more flexible.

vnbaaij: I was thinking more like 'Address Line 1' (general) instead of 'Street' (specific) (where it would still be a string)  but making it an object would offer even more flexibility, yes.

dvoituron: I renamed to these new attribute names.
- `HeaderLabel`: Gets or sets the header label (e.g Company Name) to display on the top-left.
- `HeaderButton`: Gets or sets the Header Button label (e.g. Sign out) on the top-right.
- `FooterLabel`: Gets or sets the footer label to display on the bottom-left.
- `FooterLink`: Gets or sets the Footer hyperlink label (e.g. View account) on the bottom-right.

vnbaaij (approved):

PR #1710, opened 2024-03-17
[FluentPersona] Manage the empty Name

dvoituron:
# [FluentPersona] Manage the empty Name

Manage the empty `Name=""` to display a **Persona** without other properties.

- Adding `@attributes`. 
- Adding `OnClick` event.
- Adding `String.IsNullOrEmpty(Name)` to not generate the code if content of name is empty.

## Examples

image

```xml
<FluentPersona Name="Lydia Bauer"
               ImageSize="50px"
               Status="PresenceStatus.Busy"
               StatusSize="PresenceBadgeSize.Small">
</FluentPersona>

<FluentPersona Initials="LB" ImageSize="50px" />

<FluentPersona ImageSize="50px" />
```

## Unit Tests

Unit tests updated and verified.

### DIFF
--- a/src/Core/Components/List/FluentPersona.razor
+++ b/src/Core/Components/List/FluentPersona.razor
@@ -1,8 +1,8 @@
 ﻿@namespace Microsoft.FluentUI.AspNetCore.Components
 @inherits FluentComponentBase
 
-<div id="@Id" class="@ClassValue" style="@StyleValue">
-    <div class="initials">
+<div id="@Id" class="@ClassValue" style="@StyleValue" @attributes="@AdditionalAttributes">
+    <div class="initials" @onclick="@OnClick" style="@(OnClick.HasDelegate ? $"cursor: pointer;" : null)">
         <FluentPresenceBadge Status="@Status" Size="@StatusSize" StatusTitle="@StatusTitle">
             <div style="display: table-cell; vertical-align: middle; @GetImageMinSizeStyle()">
                 @if (string.IsNullOrEmpty(Image))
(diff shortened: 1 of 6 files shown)


src/Core/Components/List/FluentPersona.razor
```diff
@@ -1,8 +1,8 @@
 ﻿@namespace Microsoft.FluentUI.AspNetCore.Components
 @inherits FluentComponentBase
 
-<div id="@Id" class="@ClassValue" style="@StyleValue">
-    <div class="initials">
+<div id="@Id" class="@ClassValue" style="@StyleValue" @attributes="@AdditionalAttributes">
+    <div class="initials" @onclick="@OnClick" style="@(OnClick.HasDelegate ? $"cursor: pointer;" : null)">
         <FluentPresenceBadge Status="@Status" Size="@StatusSize" StatusTitle="@StatusTitle">
```
vnbaaij, line 6: Should there be a check on not rendering PresenceBadge when `Status` is `null`

vnbaaij (changes requested):

src/Core/Components/List/FluentPersona.razor
dvoituron: That could be done... but no code will be changed because the **Initials** are included in the DIV generated. So, I cannot remove the **Presence Status** code :-(

```html
<div class="fluent-persona">
  <div class="initials">
    <div class="fluent-presence-badge">
	  <div style="display: table-cell; vertical-align: middle; width: 50px; min-width: 50px; height: 50px; min-height: 50px;">
	    LB
	  </div>
	</div>
  </div>
</div>
```

vnbaaij (approved):

PR #1729, opened 2024-03-21
feat(MessageService): added the methods 'ShowMessageBar' and 'ShowMessageBarAsync' with the additional parameter 'timeout' which is used  to set the 'Timeout' porperty of the 'Message' class

lettuceDestroyer:
# Pull Request

### General

- [x ] I have read the [CONTRIBUTING](https://github.com/Microsoft/fluentui-blazor/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://www.fast.design/docs/community/code-of-conduct/#our-standards) for this project.

### Component-specific

- [x ] I have modified an existing component

### DIFF
--- a/src/Core/Components/MessageBar/Services/IMessageService.cs
+++ b/src/Core/Components/MessageBar/Services/IMessageService.cs
@@ -25,6 +25,9 @@ public interface IMessageService : IDisposable
     Message ShowMessageBar(string title, MessageIntent intent, string section);
 
     /// <summary />
+    Message ShowMessageBar(string title, MessageIntent intent, string section, int timeout);
+
+    /// <summary />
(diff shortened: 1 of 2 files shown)


vnbaaij:
Hi, 
Can you explain what the big advantage of this is compared to using the method variant that takes the `MessageOptions` as a parameter?

dvoituron:
In fact, there's already a `MessageOptions` parameter that lets you configure the **MessageBar**.
Why should this `Timeout` attribute be more important than the other options? 
I don't think all the options should be in the parameter. 

If the need is to allow this `Timeout` to be modified **globally** in the application, I think it's more useful to add a property to the global configuration `LibraryConfiguration`.

Issue #1735, opened 2024-03-25
feat: add capability in FluentPersona for the name to appear before or after initials.


subratb:
# 🙋 Feature Request

FluentPersona should allow customization for the name to appear either before or after the initials.

## 🤔 Expected Behavior

![image](https://github.com/microsoft/fluentui-blazor/assets/2043577/9c2d9478-12e4-4419-8f4e-45edb7fe1119)

or

![image](https://github.com/microsoft/fluentui-blazor/assets/2043577/a48a8056-6c73-4f91-acc7-21102f7166ba)

## 😯 Current Behavior

![image](https://github.com/microsoft/fluentui-blazor/assets/2043577/cac4afde-62ee-4746-bbb3-f4d19aa93706)

## 💁 Possible Solution

Add ImageAlign parameter to FluentPersona that takes left or right as values.

```
<FluentPersona Name="Anonymous"
               Status="PresenceStatus.Offline"
               StatusSize="PresenceBadgeSize.Small"
               StatusTitle="Anonymous user is offline"
               Image="@(new Icons.Regular.Size48.Person().ToDataUri(size: "25px", color: "white"))"
               ImageSize="50px"
               ImageAlign="right">
</FluentPersona>
```

## 🔦 Context

Trying to place the image after the name. Right now, FluentPersona is inside another container where it appears on the right without any name supplied to it. The parent container displays the name before FluentPersona.

## 💻 Examples

![image](https://github.com/microsoft/fluentui-blazor/assets/2043577/a48a8056-6c73-4f91-acc7-21102f7166ba)

PR #1750, opened 2024-03-27
feat: add capability in FluentPersona for the name to appear before or after initials.

subratb:
# Pull Request

## 📖 Description

feat: add capability in FluentPersona for the name to appear before or after initials.

### 🎫 Issues

 #1735 

## 👩‍💻 Reviewer Notes

Unit test was updated to test the change. Demo application FluentPersona section was also updated with example.

## 📑 Test Plan

A new unit test was added.

### General


vnbaaij:
Hi, 

Thanks for contributing. One thing though...for this functionality I think it is better to use 'Start' and  'End' instead of 'Left' and 'Right' because we also support RTL layouts with these components. 
In the config you have added now, if you set the initials to left and use RTL, they would actualy be on the right.

src/Core/Components/List/FluentPersona.razor.cs
```diff
@@ -49,6 +49,12 @@ public partial class FluentPersona : FluentComponentBase
     [Parameter]
     public string? ImageSize { get; set; }
 
+    /// <summary>
+    /// Gets or sets the position of the image.
+    /// </summary>
+    [Parameter]
+    public PersonaImagePosition? ImagePosition { get; set; }
```
dvoituron, line 56: I think it would be more logical to follow the "same philosophy" used by React:
by calling the new attribute `TextPosition`.
 
https://react.fluentui.dev/?path=/docs/components-persona--default#text-alignment

src/Core/Enums/PersonaImagePosition.cs
```diff
@@ -0,0 +1,17 @@
...
+
+/// <summary>
+/// Describes the position at which a <see cref="FluentPersona"/> image is shown.
+/// </summary>
+public enum PersonaImagePosition
+{
+    /// <summary>
+    /// The tooltip is positioned to the left of the anchor element.
+    /// </summary>
+    Left,
```
dvoituron, line 11: As in my previous comment, and to follow Vincent's logic, it's probably better to use the names `Before` and `After`.

src/Core/Components/List/FluentPersona.razor.cs
dvoituron, line 56: I wouldn't put a nullable property, but with a default value of "Right/After".

src/Core/Components/List/FluentPersona.razor
```diff
@@ -2,23 +2,26 @@
...
-                    @(Initials ?? GetDefaultInitials())
-                }
-                else
-                {
-                    <img src="@Image" alt="@Name" title="@Name" style="@GetImageMaxSizeStyle()" />
-                }
-            </div>
-        </FluentPresenceBadge>
-    </div>
+    @if (!ImagePosition.HasValue || ImagePosition == PersonaImagePosition.Left)
```
dvoituron, line 5: To avoid duplicating code in this file (depending on the value of `ImagePosition`), you could use CSS like this, based on a new `position` attribute to be defined on the `div` "main".

```xml
<div id="@Id" class="@ClassValue" position="@ImagePosition.ToAttributeValue()" ...
```

```css
.fluent-persona[position='before'] .name {
    order: 1;
    margin-inline-end: 16px;
}

.fluent-persona[position='before'] .close {
    order: 2;
}

.fluent-persona[position='before'] .initials {
    order: 3;
}
``` 

What do you think? Possible (to be verified if no side effects are detected).

dvoituron (changes requested):

src/Core/Enums/PersonaImagePosition.cs
vnbaaij: We use start/end throughout the library (icon slots, alignment, ...) so I would stick with that. Don't recall we use before/after anywhere

src/Core/Components/List/FluentPersona.razor.cs
vnbaaij: End 😉

subratb:

src/Core/Components/List/FluentPersona.razor
subratb: Done. It is cleaner. We might have to revisit this if the goal is feature-parity with fluentui react.

vnbaaij (approved):

tests/Core/Microsoft.FluentUI.AspNetCore.Components.Tests.csproj
```diff
@@ -62,4 +62,11 @@
     
   </ItemGroup>
 
+  <ItemGroup>
```
dvoituron, line 65: Can you remove these new lines. Not required in this csproj.

subratb: Sent #1769.

PR #1769, opened 2024-03-30
feat: add capability in FluentPersona for the name to appear before or after initials. Clean up test project.

subratb:
# Pull Request

## 📖 Description

Test project had specific file reference, which was removed.
### 🎫 Issues

#1750 

### General

- [X] I have added tests for my changes.
- [X] I have tested my changes.
- [X] I have updated the project documentation to reflect my changes.
- [X] I have read the [CONTRIBUTING](https://github.com/Microsoft/fluentui-blazor/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://www.fast.design/docs/community/code-of-conduct/#our-standards) for this project.

PR #1775, opened 2024-04-01
[DataGrid] Add MarkupString support in FluentDataGrid

### LINKED ISSUE #1764: feat: add MarkupString support in FluentDataGrid
# 🙋 Feature Request

Add support for [`MarkupString`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.components.markupstring?view=aspnetcore-8.0)s to be rendered within `FluentDataGrid` cells when the content is populated through `PropertyColumn`s.

## 🤔 Expected Behavior

`MarkupString` objects can be used in `FluentDataGrid`s so that additional HTML can be dynamically rendered within cells.

## 😯 Current Behavior

Currently, trying to pass in `MarkupString` objects with HTML tags results in the HTML being rendered as plaintext.

## 💁 Possible Solution

It looks like the `PropertyColumn` currently passes a string to the render tree builder that is created from the `_cellTextFunc`. This function is defined to return a string, so even setting the `TProp` type param as `MarkupString` does not do anything.

One potential solution could be a switch parameter added to `PropertyColumn`, and if `true`, use a different cell text function that outputs a `MarkupString`. Another solution could be to add special logic that checks if `TProp` is `MarkupString`, and if so, do the same.

## 🔦 Context

I'm querying data that has text occasionally containing links (i.e., "https://..."). This data is displayed in a `FluentDataGrid`, but as it is now, the links are plaintext and are not hyperlinked. I have logic in place to identify these snippets and wrap them with HTML `a` tags and cast them to `MarkupString`s, but `PropertyColumn` converts these back to `string`s before any of the markup parsing can take place.

## 💻 Examples

```html
<fluent-data-grid-cell cell-id="c10" cell-type="default" grid-column="4" class="col-justify-start multiline-text" title="..." aria-label="..." b-w6qdxfylwy="" tabindex="-1" role="gridcell" style="grid-column: 4;">
  #shadow-root (open)
  "Most Azure services and client SDKs include a retry mechanism. However, these differ because each service has different characteristics and requirements, and so each retry mechanism is tuned to a specific service

  Reference: <a href="https://learn.microsoft.com/en-gb/azure/architecture/best-practices/retry-service-specific#azure-storage-retry-guidelines" target="_blank">https://learn.microsoft.com/en-gb/azure/architecture/best-practices/retry-service-specific#azure-storage-retry-guidelines</a>"
  <!---->
</fluent-data-grid-cell>
```

image

vnbaaij:
Fix #1764 
Add support for [MarkupString](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.components.markupstring?view=aspnetcore-8.0)s to be rendered within FluentDataGrid cells when the content is populated through PropertyColumns.

I'm implementing a solution for it like this:

```
 /// <inheritdoc />
 protected internal override void CellContent(RenderTreeBuilder builder, TGridItem item)
 => CellContent(builder, item, false);

 protected internal override void CellContent(RenderTreeBuilder builder, TGridItem item, bool allowRawHtml)
 => builder.AddContent(0, new MarkupString(_cellTextFunc?.Invoke(item) ?? string.Empty));
```
So a new overloaded method and a new `bool AllowRawHtml` parameter (defaults to false).

When using this new variant on a `TemplateColumn` it is just being ignored as it is not possible to pass the parameter to the ChildContent it renders.  

With this new implementation there are no breaking changes.

Also process change from #1769


vnbaaij:
Yes, `TemplateColumn` offers a solution. Closing this

Issue #1762, opened 2024-03-29
feat: add internal scope to DateTimeExtensions, EnumExtensions classes


adospace:
# 🙋 Feature Request

DateTimeExtensions, EnumExtensions  should be internal as BooleanExtensions for example

## 🤔 Expected Behavior

It's common to have extension methods for the DateTime, i.e. ToDateOnly(). 
Adding reference to Fluent UI package should not interfere with other locally referenced projects that define extension methods for DateTime or Enum

## 😯 Current Behavior

Referencing Fluent UI package raises exceptions like: 
The call is ambiguous between the following methods or properties: 'our extension methid' and 'Microsoft.FluentUI.AspNetCore.Components.DateTimeExtensions.ToDateTime(System.DateOnly?)'

## 💁 Possible Solution

Make the DateTimeExtensions, EnumExtensions internal classes (i.e. used only internally and not exposed)

## 🔦 Context

Upgrading to the latest version causes conflicting errors caused by the ToDateOnly() and other extension methods imported from the Fluent UI package

dvoituron:
These methods are not used only in the Lib, but are available for devs. Like explain on this page, your can use it to simplify the Binding of DateOnly or TimeOnly types. https://www.fluentui-blazor.net/DateTime

Probably your problem is due to the namespace included in `_import.razor` file. Correct ?

We could change the namespace of these extension methods to another one, to avoid conflict.

adospace:
Hi, thanks for the prompt reply!

Yes, moving them to another namespace like Microsoft.FluentUI.AspNetCore.Components.Extensions would be a good solution.
That said, I still think that such kind of extension methods should be kept internal, but it's just my personal opinion.

Btw, the quality of the framework is outstanding, great work!

vnbaaij:
Hi,

The work to do this is done (in the PR mentioned above). We can not merge it yet because it is a (small) breaking change. We want to highlight that in a blog post, etc. We only do those for major and minor (ie 4.x) releases and not for patch releases (ie 4.6.x).

adospace:
the issue can be closed for me, thank you very much!

PR #1776, opened 2024-04-01
[Extensions] Add a namespace for extension methods to prevent possible naming clashes

dvoituron:
# [Extensions] Update of the namespace for all extension methods

#1762

Most projects add `@using Microsoft.FluentUI.AspNetCore.Components` to `_Imports.razor`.
To avoid conflicts with existing methods in other libraries, we've decided to place extension 
methods in a **Extensions** sub-namespace.

This can be a **minor breaking-change**: requiring only an `using`.

For existing projects:
1. Add `@using Microsoft.FluentUI.AspNetCore.Components.Extensions` to `_Imports.razor` file.
2. Add `using Microsoft.FluentUI.AspNetCore.Components.Extensions` to the top of some C# files.

In the case of a conflict, you can remove these lines to distinguish the methods of **FluentUI.AspNetCore.Components** 
from those of other libraries.


vnbaaij (commented):
I'm ok with the changes, but as it is a (minor) breaking change, I dont want to release this in a patch release. Think it is better to hold it for the next minor release as we give those more publicity and exposure.

dvoituron:
> I'm ok with the changes, but as it is a (minor) breaking change, I dont want to release this in a patch release. Think it is better to hold it for the next minor release as we give those more publicity and exposure.

Yes, we can wait for v4.7. But we could merge it, in `dev`, a few days before to have time to receive optional comments.

vnbaaij (approved):

PR #1786, opened 2024-04-03
[CounterBadge] Add ShowWhen, Dot, and VerticalPosition. 

vnbaaij:
To make the counterbadge more versatile, some rework has been done. Previously when the `Count` value was 0, the whole component would not get rendered. This made it hard to use the component because you needed a `if..else` construct to render the component that the FluentCounterBadge wraps. In most cases that lead to repeated rendering logic.

With the changes made, the `FluentCounterBadge` will always render its container and `ChildContent`. The rendering of the actual badge is still dependant on the `Count'. By default the badge will not be shown if the `Count` is zero.

New parameters:
- `ShowWhen`: a `Func<int,bool>` which can be used to specify the condition that determines if the badge is shown. This replaces `ShowZero` which has now been obsoleted. Default value is `Count => Count > 0`. To get the `ShowZero` functionality, you could  specify `ShowWhen="@(Count => Count >= 0)"`

image

- `Dot` (`bool`, default false). When set to true, a dot is rendered instead of a full badge. `Count` is ignored in this case. `BadgeContent`

image

- `VerticalPosition` (int, default 60) This just replaces the `BottomPosition` parameter (which has been obsoleted) to better align with the `HorizontalPosition` parameter name.

- `BadgeContent` has been obsoleted and is replaced by `BadgeTemplate` to be more aligned with how we use this in other parts of the library. `BadgeTemplate` is now a real template and makes the `Count` available for use throught the `@context` parameter.

image

Tests have been veriefied and updated.


src/Core/Components/CounterBadge/FluentCounterBadge.razor.cs
```diff
@@ -83,27 +96,43 @@ public partial class FluentCounterBadge : FluentComponentBase, IDisposable
...
+    public Func<int, bool> ShowWhen { get; set; } = Count => Count > 0;
+
     /// <summary>
     /// If an plus sign should be displayed when the <see cref="Count"/> is greater than <see cref="Max"/>.
     /// Defaults to true.
     ///</summary>
     [Parameter]
     public bool ShowOverflow { get; set; } = true;
 
+    public FluentCounterBadge()
```
dvoituron, line 131: Why this empty constructor?
If required, could you move it before the properties?

src/Core/Components/CounterBadge/FluentCounterBadge.razor.css
```diff
@@ -20,6 +21,12 @@
     cursor: default;
 }
 
-    .counterbadge:empty {
+    .counterbadge:not(.dot):empty {
         display: none;
     }
+
+.dot {
```
dvoituron, line 28: We shouldn't use "generic / basic" class names, to avoid conflicts with other existing libraries.
In this case, `dot` is probably already in use.
We should use a `fluentui` prefix for all classes.

In this case, it would be preferable to use a `dot-only` attribute (for example):
```xml
<div class="@ClassValue" style="@StyleValue" dot-only></div>
```

Same for `counterbadge` :-)

src/Core/Components/CounterBadge/FluentCounterBadge.razor.cs
```diff
@@ -35,8 +36,8 @@ public partial class FluentCounterBadge : FluentComponentBase, IDisposable
     /// Gets or sets the number displayed inside the badge.
     /// Can be enriched with a plus sign with <see cref="ShowOverflow"/>
     /// </summary>
-    [Parameter, EditorRequired]
-    public int? Count { get; set; }
+    [Parameter]
+    public int Count { get; set; } = 0;
```
dvoituron, line 40: Why to remove the `nullable` value? To avoid breaking-change, is it possible to test this nullable value?

dvoituron (changes requested):
Awesome new features.

src/Core/Components/CounterBadge/FluentCounterBadge.razor.css
vnbaaij: Good catch. Obviously code from the early days. Will make it work with attribute

src/Core/Components/CounterBadge/FluentCounterBadge.razor.cs
vnbaaij: Changed that back

dvoituron (approved):

PR #1790, opened 2024-04-03
[AppBar] Add Count parameter and facilitate OnClick without navigation

vnbaaij:
Add a `Count` parameter to the `FluentAppBarItem` to make it possible to show a CounterBadge on an item (like yu have in Teams with number of unread chats, etc)

AppBarCount

Also, you can now use OnClick to trigger an action with doing a navigation step. Just supply a `null` value to the AppBarItem's `Href`. See the AppBarClick demo for examples. 

#1786 needs to be merged before this


dvoituron (changes requested):
Nothing to say :-) Just to block this PR before the **CounterBadge** PR.

There are no Unit Tests to change? Maybe a good time to add these tests :-)

dvoituron (approved):

PR #1800, opened 2024-04-04
[KeyCode, AnchorRegion, Popover] Add key navigation in AnchorRegion / Popup

dvoituron:
# [KeyCode, AnchorRegion, Popover] Add key navigation in AnchorRegion / Popup

## 1. FluentKeyCode
- Updating the `FluentKeyCode` component to unregister the `keydown` event when the component is deleted.

## 2. FluentAnchorRegion
- Adding `AutoFocus` attribute to set whether this region should receive the focus when the component is loaded.
- Adding `FocusToNextElementAsync` and `FocusToOriginalElementAsync` methods to move the focus to the 
next element included in this anchor region only and to the initial element that triggered the anchor region.

## 3. FluentPopover

- Adding `AutoFocus` attribute to set whether the element should receive the focus when the component is loaded.
If this is the case, the user cannot navigate to other elements of the page while the Popup is open.
Default is true.
   
- Adding `CloseKeys` attribute to set the keys that can be used to close the popover (ESC).

## Demo (using only keyboard)

peek


src/Core/Components/AnchoredRegion/FluentAnchoredRegion.razor.cs
```diff
@@ -140,4 +149,49 @@ public partial class FluentAnchoredRegion : FluentComponentBase
...
+    protected async override Task OnAfterRenderAsync(bool firstRender)
+    {
+        if (firstRender)
+        {
+            Module ??= await JSRuntime.InvokeAsync<IJSObjectReference>("import", JAVASCRIPT_FILE);
+
+            // Go to the first Focusable element (after a delay of 100ms, to allow the DOM to render the Web Components)
+            if (AutoFocus)
+            {
+                await Module.InvokeVoidAsync("goToNextFocusableElement", Element, false, 100);
```
vnbaaij, line 169: 100ms seems a bit long to me. Can we make it shorter?

vnbaaij (approved):

src/Core/Components/AnchoredRegion/FluentAnchoredRegion.razor.cs
dvoituron: We can, but I'm afraid the Web Component won't be created. 
And it's only used on the first call (after, the time is not set), when the user performs a manual action, so it's not perceptible to the eye.

PR #1826, opened 2024-04-10
[DataGrid] Add remove sort capability

vnbaaij:
Previosly when a sort was executed on a DataGrid, (by clicking on a column header button or pressing enter when the column header button is focused), there was no option to undo that. 

With this PR it is now possible to right click on the sorted colum header button to clear the sort and restore it to the default/start situation. **updated: or use Ctrl+Enter to remove the sorting when the column header button is focused** 

When a grid has not sort specified, it just goes back to its initial state. When thare was a defaut sort specified, it will be restored:

grid-sort-remove

Based on discussion in #283

### DIFF
--- a/WHATSNEW.md
+++ b/WHATSNEW.md
@@ -1,3 +1,8 @@
+## v4.6.2
+
+### Components
+[DataGrid] Add remove sort capability on columns ([#1826](https://github.com/microsoft/fluentui-blazor/pull/1826))
+
 ## V4.6.1
 
 ### Demo site and documentation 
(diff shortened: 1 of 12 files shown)


dvoituron:
We can have a method to "remove" the sort order, but :
1. It's not very intuitive to use Right-Click, and it's certainly not compatible with Accessibility features.
2. Why would you want such an option? It doesn't exist in Windows Explorer, for example.

dvoituron (approved):

vnbaaij:
> We can have a method to "remove" the sort order, but :
> 
> 1. It's not very intuitive to use Right-Click, and it's certainly not compatible with Accessibility features.

I'll look into the a11y piece. Good point.

> 2. Why would you want such an option? It doesn't exist in Windows Explorer, for example.

It is fairly common to have a way to "unsort" a column in a DataGrid (MudBlazor, Radzen and DevExpress have it too)

tommyvct:
> 1. It's not very intuitive to use Right-Click, and it's certainly not compatible with Accessibility features.

How about click for the 3rd time to remove the sort?

vnbaaij:
> How about click for the 3rd time to remove the sort?

No. that is not an option currently. Internally the button just looks at a boolean sortByAscending variable. Would need a lot more additional coding.

tommyvct:
I see, that could introduce a lot of work and chaos...

How about have another button, like the filter button to remove the sort?

PR #1839, opened 2024-04-11
[Demo & docs] Clears demo site cache on startup if new version

c0g1t8:
# Pull Request

## 📖 Description

Clears the browser cache when the demo site is started. 

### 🎫 Issues

Old versions of the markdown are retained between sessions. So when a new version of the site is released, new markdown content will not be downloaded if an older version is cached.

For example, the WhatsNew page continues to render previous version.

## 👩‍💻 Reviewer Notes

I've made the CacheStorageAccessor a singleton and provided a flag to check if the cache has been previously cleared.

I would have preferred being to _tag_ the cache but was not sure if the extra effort was necessary. That way, the cache would only be cleared if there is a new version or build.  Thoughts?

### General

### DIFF
--- a/examples/Demo/Shared/Infrastructure/AppVersionService.cs
+++ b/examples/Demo/Shared/Infrastructure/AppVersionService.cs
@@ -1,0 +1,35 @@
+// ------------------------------------------------------------------------
+// MIT License - Copyright (c) Microsoft Corporation. All rights reserved.
+// ------------------------------------------------------------------------
+
+using System.Reflection;
+
+namespace FluentUI.Demo.Shared.Infrastructure;
+internal class AppVersionService : IAppVersionService
+{
+    public string Version
+    {
+        get => GetVersionFromAssembly();
+    }
+
+    static public string GetVersionFromAssembly()
+    {
+        string strVersion = default!;
+        var versionAttribute = Assembly.GetExecutingAssembly().GetCustomAttribute<AssemblyInformationalVersionAttribute>();
+        if (versionAttribute != null)
+        {
+            var version = versionAttribute.InformationalVersion;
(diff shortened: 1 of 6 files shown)


dvoituron (approved):

vnbaaij:
In the `CacheStorageAccessor.js` script, I 'put' the responses with a cache duration (7 days) and chechk the validity on the 'get' action. Wouldn't this change make that part a bit useless then? As everything gets cleared out when the site is restarted anyway.
Maybe we should just shorthen the cache duration (to 2 or 3 days)?

c0g1t8:
> In the `CacheStorageAccessor.js` script, I 'put' the responses with a cache duration (7 days) and chechk the validity on the 'get' action. Wouldn't this change make that part a bit useless then? As everything gets cleared out when the site is restarted anyway. Maybe we should just shorthen the cache duration (to 2 or 3 days)?

Shortening the cache time will still result in the user getting old information for several days after a new release. I'm thinking in the case of breaking changes.  

The best solution would be to _tag_ the cache with the version and build information (_Version: 4.6.1+ff23edc6_). Check to see if there is a match and only flush the cache if they dont' match.

The plan would be to implement a new service that returns the current build information and inject it into `CacheStorageAccessor`. Thoughts?

vnbaaij:
Yes, that sounds like a good plan. Tagging each item is the way to go, I think

c0g1t8:
@vnbaaij 

I put a file in the cache ('appVersion') with the version corresponding to the files cached. I used the same version string as that shown in the footer of the demo site. When this version is updated, all files cached are considered stale and removed.

c0g1t8:
Changing status back to draft. I accidentally reverted something. 😒

vnbaaij (approved):

PR #1846, opened 2024-04-12
[Input] NotifyFieldChanged is called twice for all FluentInputBase derived components

### LINKED ISSUE #1830: fix: NotifyFieldChanged is called Twice in FluentInputBase.ChangeHandlerAsync
# 🐛 Bug Report

When `FluentInputBase.ChangeHandlerAsync` is called and the input is valid, it calls `SetCurrentValueAsync` [here](https://github.com/microsoft/fluentui-blazor/blob/dev/src/Core/Components/Base/FluentInputBaseHandlers.cs#L34), which calls the `EditContext.NotifyFieldChanged` [here](https://github.com/microsoft/fluentui-blazor/blob/dev/src/Core/Components/Base/FluentInputBase.cs#L162).  It then carries on to call `NotifyFieldChanged` again [here](https://github.com/microsoft/fluentui-blazor/blob/dev/src/Core/Components/Base/FluentInputBaseHandlers.cs#L48).

This results in two notification calls in the same change. 

## 💻 Repro or Code Sample

Using a FlientTextField as such:

```
<FluentTextField Name="name" @bind-Value="Input.Name" Immediate="true" ImmediateDelay="750" Label="Name" Required Maxlength="64" />
```

After typing and the update fires, I get two calls to listeners listening to the `Name` property change.

## 🤔 Expected Behavior

Ideally, only one call should happen per control state change.

## 😯 Current Behavior

Currently two calls are made per control state change.

## 💁 Possible Solution

Reporting here to ensure I am not misunderstanding something fundamental here.

## 🔦 Context

I am learning this library and everything else has been working wonderfully. :)  Thank you for your efforts out there. 🙏

## 🌍 Your Environment

* OS & Device: Windows 10
* Browser Edge/Chrome
* .NET and Fluent UI Blazor library Version `Microsoft.FluentUI.AspNetCore.Components, Version=4.5.0.24067, Culture=neutral, PublicKeyToken=null`

vnbaaij:
Fix #1830 by adding a check to see if NotifyFieldChanged has already been called.

```
protected virtual async Task ChangeHandlerAsync(ChangeEventArgs e)
{
    var _notifyCalled = false; //<--👈added
    var isValid = TryParseValueFromString(e.Value?.ToString(), out TValue? result, out var validationErrorMessage);

    if (isValid)
    {
        await SetCurrentValueAsync(result ?? default);
        _notifyCalled = true; //<--👈added
    }
    else
    {
        if (FieldBound && CascadedEditContext != null)
        {
            _parsingValidationMessages ??= new ValidationMessageStore(CascadedEditContext);

            _parsingValidationMessages.Clear();
            _parsingValidationMessages.Add(FieldIdentifier, validationErrorMessage ?? "Unknown parsing error");
        }
    }
    if (FieldBound && !_notifyCalled)  //<--👈changed
    {
        CascadedEditContext?.NotifyFieldChanged(FieldIdentifier);
    }
}
```

### DIFF
--- a/examples/Demo/Shared/Microsoft.FluentUI.AspNetCore.Components.xml
+++ b/examples/Demo/Shared/Microsoft.FluentUI.AspNetCore.Components.xml
@@ -6288,7 +6288,7 @@
             <summary>
             Optionally supplies a template for rendering the pagination summary.
             The following values can be included:
-            {your State parameter name}.CurrentPageIndex (zero-basedd, so +1 for the current page number)
+            {your State parameter name}.CurrentPageIndex (zero-based, so +1 for the current page number)
             {your State parameter name}.LastPageIndex (zero-based, so +1 for the total number of pages)
             </summary>
         </member>
--- a/src/Core/Components/Base/FluentInputBaseHandlers.cs
+++ b/src/Core/Components/Base/FluentInputBaseHandlers.cs
@@ -27,11 +27,13 @@ public partial class FluentInputBase<TValue>
     /// <returns></returns>
     protected virtual async Task ChangeHandlerAsync(ChangeEventArgs e)
     {
+        var _notifyCalled = false;
         var isValid = TryParseValueFromString(e.Value?.ToString(), out TValue? result, out var validationErrorMessage);
 
         if (isValid)
         {
             await SetCurrentValueAsync(result ?? default);
+            _notifyCalled = true;
         }
         else
         {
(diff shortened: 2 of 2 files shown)


dvoituron (approved):

vnbaaij:
@kamazheng that is a whole differnet issue and has nothing to do with this NotifyFiledChanged.
Your issue has been addressed already in #1965

Issue #1847, opened 2024-04-12
Add option to disable FluentMenuButton items


MarvinKlein1508:
# 🙋 Feature Request

It should be possible to disable certain items in `FluentMenuButton `.

PR #1848, opened 2024-04-12
[MenuButton] Add ChildContent so items can be supplied manually

vnbaaij:
Based on issue #1847 I've added a `ChildContent` parameter to MenuButton. This allows for manually supplying the `FluentMenuItem` children and thereby use all it's parameters (like Disabled)

An example has been added 

image

### DIFF
--- a/src/Core/Components/MenuButton/FluentMenuButton.razor
+++ b/src/Core/Components/MenuButton/FluentMenuButton.razor
@@ -9,16 +9,10 @@
     </FluentButton>
     <FluentOverlay @bind-Visible="@_visible" Transparent="true" FullScreen="true" />
     <FluentMenu @ref="Menu" Anchor="@_buttonId" aria-labelledby="button" Style="@MenuStyleValue" @bind-Open=@_visible @onmenuchange=OnMenuChangeAsync>
-        @if (Items is not null)
+        @foreach (KeyValuePair<string, string> item in Items)
         {
-            @foreach (KeyValuePair<string, string> item in Items)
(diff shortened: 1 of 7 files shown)


src/Core/Components/MenuButton/FluentMenuButton.razor
```diff
@@ -9,9 +9,16 @@
     </FluentButton>
     <FluentOverlay @bind-Visible="@_visible" Transparent="true" FullScreen="true" />
     <FluentMenu @ref="Menu" Anchor="@_buttonId" aria-labelledby="button" Style="@MenuStyleValue" @bind-Open=@_visible @onmenuchange=OnMenuChangeAsync>
-        @foreach (KeyValuePair<string, string> item in Items)
+        @if (Items is not null)
```
dvoituron, line 12: Maybe interesting to remove this test and to allow both: `Items` and next `ChildContent`.
Just to add this information in the doc :-)

```razor
@foreach (KeyValuePair<string, string> item in Items)
{
    <FluentMenuItem Id="@item.Key">@item.Value</FluentMenuItem>
}

@ChildContent
```

dvoituron (changes requested):

src/Core/Components/MenuButton/FluentMenuButton.razor
vnbaaij: Yes, that is even better. Made that change and added documentation.

dvoituron (approved):

PR #1866, opened 2024-04-15
Fix Aspire 3364 by adding IKeyCodeListner and handler

vnbaaij:
Fix https://github.com/dotnet/aspire/issues/3364.  Add DataGrid header resizing by using the keyboard.

- `+` - increase column width by 10px
- `-` - decrease column width by 10 px
- `Shift+r` reset to original values
(these should be the same keys as used elsewhere in Aspire Dashboard.

grid-resize-keyboard


src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -215,6 +216,12 @@ public FluentDataGrid()
         columnsFirstCollectedSubscriber.SubscribeOrMove(_internalGridContext.ColumnsFirstCollected);
     }
 
+    /// <inheritdoc />
+    protected override void OnInitialized()
+    {
+        KeyCodeService.RegisterListener(OnKeyDownAsync);
```
dvoituron, line 222: If you register the service using a method (not `this`) you don't need to add the interface `IKeyCodeListener`.
The interface should be used only when you register the "full" component `this`

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -558,4 +565,40 @@ private async Task HandleOnRowFocusAsync(DataGridRowFocusEventArgs args)
...
+        if (args.ShiftKey == true && args.Key == KeyCode.KeyR)
+        {
+            ResetColumnWidths();
+        }
+
+        if (args.Key == KeyCode.NumpadSubtract || args.Key == KeyCode.Minus2)
+        {
+            SetColumnWidth(-10);
+        }
+        if (args.Key == KeyCode.NumpadAdd || (args.ShiftKey == true && (args.Key == KeyCode.Equal || args.Key == KeyCode.Equal2)))
```
dvoituron, line 580: You could use `if (args.Value = "+")` to simplify this test (same for "Minus")

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -558,4 +565,40 @@ private async Task HandleOnRowFocusAsync(DataGridRowFocusEventArgs args)
...
+            SetColumnWidth(10);
+        }
+        return Task.CompletedTask;
+    }
+
+    private void SetColumnWidth(float widthChange)
+    {
+        if (_gridReference is not null)
+        {
+            _ = Module?.InvokeVoidAsync("resizeColumn", _gridReference, widthChange).AsTask();
```
dvoituron, line 592: Why are you using `AsTask`. I prefer to use `async / await`, transforming the `SetColumnWidth` to an async method (suffix Async)

dvoituron (changes requested):
I have 2 questions:
1. Why this "resize" feature works only for the "Typical usage" data grid and not for other in the Demo page?
2. It seems strange that when you reduce one column (including the "extra strong" column), all the others retain their current size.
e.g. I pressed `-` multiple-times on the "Name of Country" column...
peek

vnbaaij:
1) This only works for resizable DataGrid's
2) This is fixed now

vnbaaij:
Resize bugs fixed: 
grid-resize-keyboard 2

dvoituron (approved):

Issue #1836, opened 2024-04-11
fix: Cross Dependent sliders do not reset thumb after range is updated.


russellSmithMS:
# 🐛 Bug Report

I have a Blazor app that I am using to display electrical usage over time.  On my page, I have 4 sliders.  2 represent the years associated with the data and the other 2 represent the months.  These are used for min/max.  When I drag the start year, I want the min option for end year to be no smaller than the selected start year.  Same goes for Start/end month.  
The underlying selected (bound) values are correct, the display of the slider bar is what does not line up.  

## 💻 Repro or Code Sample

This is a page I created to repro this issue.
```razor
@page "/"
@rendermode InteractiveServer
<PageTitle>Home</PageTitle>

<FluentGrid Style="width:100%">
    <FluentGridItem xs="12">
        <h1>Related Sliders</h1>
    </FluentGridItem>
    <FluentGridItem xs="2" Justify="JustifyContent.FlexEnd">From Month</FluentGridItem>
    <FluentGridItem xs="4">
        <FluentSlider Step="1" Min="1" Max="EndMonth" @bind-Value="StartMonth">
            @foreach (var sy in GetCollection(1, EndMonth))
            {
                <FluentSliderLabel Position="@sy">@sy</FluentSliderLabel>
            }
        </FluentSlider>
    </FluentGridItem> 
    <FluentGridItem xs="2" Justify="JustifyContent.FlexEnd">To Month</FluentGridItem>
    <FluentGridItem xs="4">
        <FluentSlider Step="1" Min="StartMonth" Max="12" @bind-Value="EndMonth">
            @foreach (var sy in GetCollection(StartMonth, 12))
            {
                <FluentSliderLabel Position="@sy">@sy</FluentSliderLabel>
            }
        </FluentSlider>
    </FluentGridItem>
</FluentGrid>
@code {
    private int StartMonth = 1;
    private int EndMonth = 12;

    public IEnumerable<int> GetCollection(int start, int end) => Enumerable.Range(start, end - start + 1);
}
```
This is what the output looks like
![image](https://github.com/microsoft/fluentui-blazor/assets/113932296/24f3a41c-c408-4413-9c40-2a81ea892cb1)

When I slide the To Month to 6 The from month's max value changes to 6 and the steps reflect properly.
![image](https://github.com/microsoft/fluentui-blazor/assets/113932296/90186451-addd-458e-b666-38c850d9bf80)

[...]

PR #1873, opened 2024-04-16
[FluentSlider] Fixes thumb redraw issues (Fix for #1836)

c0g1t8:
# Pull Request

## 📖 Description

[FluentSlider] Fixes thumb redraw issues (Fix for #1836)

### 🎫 Issues

- thumb not being redrawn when min or max is changed as per #1836 
- unable to programmtically change thumb value

## 👩‍💻 Reviewer Notes

Part of this fix is dependent on the implementation of the underlying [fast slider](https://github.com/microsoft/fast/blob/master/packages/web-components/fast-foundation/src/slider/slider.ts) component. It calls the `setThumbPositionForOrientation` function which is marked as private.

https://github.com/microsoft/fast/blob/5b2b801208a86c1d342feff49961f5c490ad247e/packages/web-components/fast-foundation/src/slider/slider.ts#L338-L355

### General

### Component-specific

### DIFF
--- a/src/Core/Components/Slider/FluentSlider.razor.cs
+++ b/src/Core/Components/Slider/FluentSlider.razor.cs
@@ -3,12 +3,22 @@ using System.Globalization;
 
 using Microsoft.AspNetCore.Components;
 using Microsoft.FluentUI.AspNetCore.Components.Utilities;
+using Microsoft.JSInterop;
 
 namespace Microsoft.FluentUI.AspNetCore.Components;
 
-public partial class FluentSlider<TValue> : FluentInputBase<TValue>
+public partial class FluentSlider<TValue> : FluentInputBase<TValue>, IAsyncDisposable
     where TValue : System.Numerics.INumber<TValue>
 {
+    private const string JAVASCRIPT_FILE = "./_content/Microsoft.FluentUI.AspNetCore.Components/Components/Slider/FluentSlider.razor.js";
+
+    /// <summary />
+    [Inject]
+    private IJSRuntime JSRuntime { get; set; } = default!;
+
+    /// <summary />
+    private IJSObjectReference? Module { get; set; }
+
     /// <summary>
     /// Gets or sets the slider's minimal value.
     /// </summary>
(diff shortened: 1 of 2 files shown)


vnbaaij (approved):

src/Core/Components/Slider/FluentSlider.razor.cs
```diff
@@ -62,6 +72,15 @@ protected override void OnParametersSet()
         ArgumentNullException.ThrowIfNull(Step, nameof(Step));
     }
 
+    protected override async Task OnAfterRenderAsync(bool firstRender)
+    {
+        if (firstRender)
+        {
+            Module ??= await JSRuntime.InvokeAsync<IJSObjectReference>("import", JAVASCRIPT_FILE);
+        }
+        await Module!.InvokeVoidAsync("updateSlider", Element);
```
dvoituron, line 81: Please NEVER use the `'!` to force the compiler to say that this object is not null. Technically, it is possible for `Module` to be `null`. So, prefer to use the test in this case: `if (Module is not null) ...`.

I suppose it's deliberate to execute this JavaScript method outside `firstRender'?

c0g1t8: @dvoituron

I got busy and  didn't have a chance to respond to this earlier. 

> Please NEVER use the '! to force the compiler to say that this object is not null. Technically, it is possible for Module to be null. So, prefer to use the test in this case: if (Module is not null) ....

It's my understanding that even though the compiler flags the possibility that `Module` may be null, it would require
1. `firstRender` is `false` on the firstpass which can't happen based on Blazor lifecycle.
2. Or, `InvokeAsync` returns null which it can't based on it's definition. If this call failed, an exception would be thrown.

I know this PR has been merged but can make this change for the sake of defensive programming 😄

> I suppose it's deliberate to execute this JavaScript method outside `firstRender'?

Yes, I did this in case `StateHasChanged` is called in the future for some reason.

geowong2022: Hi @c0g1t8 @dvoituron , I hit this error today in Fluent v3.70,

Error: System.AggregateException: One or more errors occurred. (Value cannot be null. (Parameter 'jsObjectReference'))
 ---> System.ArgumentNullException: Value cannot be null. (Parameter 'jsObjectReference')
   at Microsoft.Fast.Components.FluentUI.FluentSlider`1.OnAfterRenderAsync(Boolean firstRender) in /_/src/Core/Components/Slider/FluentSlider.razor.cs:line 81
   --- End of inner exception stack trace ---

Any idea?

c0g1t8: @geowong2022 

> I hit this error today in Fluent v3.70
 
Were you able to reproduce this error with the current release?  

BTW - If you can provide a repro, I'll take a look.

PR #1899, opened 2024-04-19
[FluentGrid] Add AdaptiveRendering property

dvoituron:
# [FluentGrid] Add AdaptiveRendering property

**AdaptiveRendering** allows you not to generate the HTML code of the **FluentGridItem** that respects the `HiddenWhen`. In other words, when `AdaptiveRendering=false` (default), the content of the **FluentGridItem** is simply hidden by CSS styles, whereas if `AdaptiveRendering=true`, the content of the **FluentGridItem** is not rendered by Blazor. In certain cases where this content is important, it may be more efficient not to generate the HTML code instead of simply hiding it.

```xml
<FluentGrid AdaptiveRendering="true">
   ...
</FluentGrid> 
```

peek_1

### DIFF
--- a/examples/Demo/Shared/Pages/Grid/GridPage.razor
+++ b/examples/Demo/Shared/Pages/Grid/GridPage.razor
@@ -55,6 +55,12 @@
     e.g. <code>GridItemHidden.Md | GridItemHidden.Lg</code> will display the element for all screen sizes, except on medium and large devices.
 </p>
 <p>
+    <b>AdaptiveRendering</b> allows you not to generate the HTML code of the <b>FluentGridItem</b> that respects the <code>HiddenWhen</code>.
+    In other words, when <code>AdaptiveRendering=false</code> (default), the content of the <b>FluentGridItem</b> is simply hidden by CSS styles,
+    whereas if <code>AdaptiveRendering=true</code>, the content of the <b>FluentGridItem</b> is not rendered by Blazor.
+    In certain cases where this content is important, it may be more efficient not to generate the HTML code instead of simply hiding it.
+</p>
+<p>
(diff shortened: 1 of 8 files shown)


examples/Demo/Shared/Pages/Grid/GridPage.razor
```diff
@@ -54,6 +54,12 @@
     To display an element only on a given range of screen sizes, you can combine a these values.
     e.g. <code>GridItemHidden.Md | GridItemHidden.Lg</code> will display the element for all screen sizes, except on medium and large devices.
 </p>
+<p>
+    <b>AdaptiveRendering</b> allows you not to generate the HTML code of the <b>FluentGridItem</b> that respects the <code>HiddenWhen</code>.
```
vnbaaij, line 58: <b>AdaptiveRendering</b> allows for not generating HTML code in a  <b>FluentGridItem</b> based on the value of the <code>HiddenWhen</code> parameter.

examples/Demo/Shared/Pages/Grid/GridPage.razor
```diff
@@ -54,6 +54,12 @@
...
+    In other words, when <code>AdaptiveRendering=false</code> (default), the content of the <b>FluentGridItem</b> is simply hidden by CSS styles,
+    whereas if <code>AdaptiveRendering=true</code>, the content of the <b>FluentGridItem</b> is not rendered by Blazor.
+    In certain cases where this content is important, it may be more efficient not to generate the HTML code instead of simply hiding it.
```
vnbaaij, line 61: This allows for fine-grained control over when HTML is generated or not, for example in a case where rendering the grid item takes a lot of time or leads to a lot of data being transferred.

vnbaaij (changes requested):
Nice addition! Just some doc remarks

vnbaaij (approved):

PR #1903, opened 2024-04-20
[Fluent DatePicker] Enhance for month /year selection 

StefanJansson:
Enhancement for selecting a different view, month or year, and a custom DateTimeFormat

# Pull Request

## 📖 Description

FluentDatePicker is missing the possibility to select view like month or year which is available in FluentCalendar. 

## 📑 Test Plan

Create a FluentDatePicker, set view to desired choice of Month or Year. Possibility to add custom DateTimeFormat 

### General

- [X] I have read the [CONTRIBUTING](https://github.com/microsoft/fluentui-blazor/blob/main/docs/contributing.md) documentation and followed the [standards](https://www.fast.design/docs/community/code-of-conduct/#our-standards) for this project.

### Component-specific

### DIFF
--- a/src/Core/Components/DateTime/CalendarTitles.cs
+++ b/src/Core/Components/DateTime/CalendarTitles.cs
@@ -1,4 +1,3 @@
-using Microsoft.VisualBasic;
 
 namespace Microsoft.FluentUI.AspNetCore.Components;
 
@@ -15,15 +14,8 @@ internal class CalendarTitles
         _calendar = calendar;
         CalendarExtended = new CalendarExtended(calendar.Culture, calendar.PickerMonth);
         View = calendar.View;
-        DateTimeFormat = calendar.DateTimeFormat;
     }
     
-    /// <summary>
-    /// Gets the user defined DateTimeFormat to use for the calendar.
-    /// </summary>
-
-    public string? DateTimeFormat { get; set; }
-
     /// <summary>
     /// Gets the CalendarExtended to use for the calendar.
     /// </summary>
--- a/src/Core/Components/DateTime/CalendarTitles.cs
+++ b/src/Core/Components/DateTime/CalendarTitles.cs
@@ -1,3 +1,5 @@
+using Microsoft.VisualBasic;
+
 namespace Microsoft.FluentUI.AspNetCore.Components;
 
 internal class CalendarTitles
@@ -13,7 +15,14 @@ internal class CalendarTitles
         _calendar = calendar;
         CalendarExtended = new CalendarExtended(calendar.Culture, calendar.PickerMonth);
         View = calendar.View;
+        DateTimeFormat = calendar.DateTimeFormat;
     }
+    
+    /// <summary>
+    /// Gets the user defined DateTimeFormat to use for the calendar.
+    /// </summary>
+
+    public string? DateTimeFormat { get; set; }
 
     /// <summary>
     /// Gets the CalendarExtended to use for the calendar.
--- a/src/Core/Components/DateTime/CalendarTitles.cs
+++ b/src/Core/Components/DateTime/CalendarTitles.cs
@@ -15,7 +15,7 @@ internal class CalendarTitles
         CalendarExtended = new CalendarExtended(calendar.Culture, calendar.PickerMonth);
         View = calendar.View;
     }
-    
+
     /// <summary>
     /// Gets the CalendarExtended to use for the calendar.
     /// </summary>
--- a/src/Core/Components/DateTime/FluentCalendarBase.cs
+++ b/src/Core/Components/DateTime/FluentCalendarBase.cs
@@ -47,12 +47,6 @@ public abstract class FluentCalendarBase : FluentInputBase<DateTime?>
     public virtual CalendarViews View { get; set; } = CalendarViews.Days;
 
     /// <summary>
-    /// Defines the appearance of the <see cref="FluentCalendar"/> component.
-    /// </summary>
-    [Parameter]
-    public virtual string? DateTimeFormat { get; set; } = null;
-    
-    /// <summary>
     /// Gets or sets the selected date (two-way bindable).
     /// </summary>
     [Parameter]
--- a/src/Core/Components/DateTime/FluentCalendarBase.cs
+++ b/src/Core/Components/DateTime/FluentCalendarBase.cs
@@ -41,6 +41,18 @@ public abstract class FluentCalendarBase : FluentInputBase<DateTime?>
     public bool CheckIfSelectedValueHasChanged { get; set; } = true;
 
     /// <summary>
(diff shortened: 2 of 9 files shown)


StefanJansson:
@microsoft-github-policy-service agree [company="Veum"]

StefanJansson:

src/Core/Components/DateTime/FluentCalendarBase.cs
```diff
@@ -40,6 +40,18 @@ public abstract class FluentCalendarBase : FluentInputBase<DateTime?>
...
+    /// Defines the appearance of the <see cref="FluentCalendar"/> component.
+    /// </summary>
+    [Parameter]
+    public virtual CalendarViews View { get; set; } = CalendarViews.Days;
+
+    /// <summary>
+    /// Defines the appearance of the <see cref="FluentCalendar"/> component.
+    /// </summary>
+    [Parameter]
+    public virtual string? DateTimeFormat { get; set; } = null;
```
dvoituron, line 53: I don't think it's a good idea to allow date display customization. It's better to continue to use the default OS format: `Culture.DateTimeFormat`.

This would cause too many problems when encoding or managing Accessibility.
For example, in your case, the encoding of `2006-01-01 - 2006-01-01`, which is correct, doesn't work.

peek

dvoituron (changes requested):
Could you add some unit tests (to validate all new features)?
Without them, we won't be able to accept the PR :-(

src/Core/Components/DateTime/CalendarTitles.cs
```diff
@@ -1,3 +1,5 @@
+using Microsoft.VisualBasic;
```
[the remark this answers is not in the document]
StefanJansson: Didn't even notiv´ce it, I'll remove it

src/Core/Components/DateTime/FluentCalendarBase.cs
StefanJansson: OK, I'll fugire out a better way

dvoituron: Why do you want a so specific format for a Year selector? 
Why not to display "the year" like `2006` (based on the Culture)?

src/Core/Components/DateTime/CalendarTitles.cs
```diff
@@ -13,7 +15,14 @@ public CalendarTitles(FluentCalendar calendar)
...
         CalendarExtended = new CalendarExtended(calendar.Culture, calendar.PickerMonth);
         View = calendar.View;
+        DateTimeFormat = calendar.DateTimeFormat;
     }
+    
+    /// <summary>
+    /// Gets the user defined DateTimeFormat to use for the calendar.
+    /// </summary>
+
+    public string? DateTimeFormat { get; set; }
```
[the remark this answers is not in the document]
StefanJansson: Removed

src/Core/Components/DateTime/FluentCalendarBase.cs
StefanJansson: Replaced to use System.Globalizarion class except for year view where I didn't find any usable and used hard coded yyyy

src/Core/Components/DateTime/CalendarTitles.cs
```diff
@@ -1,3 +1,4 @@
+
```
dvoituron, line 1: Could you remove the line (to not have an empty line)?

src/Core/Components/DateTime/CalendarTitles.cs
```diff
@@ -14,7 +15,7 @@ public CalendarTitles(FluentCalendar calendar)
         CalendarExtended = new CalendarExtended(calendar.Culture, calendar.PickerMonth);
         View = calendar.View;
     }
-
+
```
dvoituron, line 18: Same here, could you remove the "Empty chars"?

src/Core/Components/DateTime/FluentDatePicker.razor
```diff
@@ -1,6 +1,16 @@
-@namespace Microsoft.FluentUI.AspNetCore.Components
+﻿@namespace Microsoft.FluentUI.AspNetCore.Components
 @inherits FluentCalendarBase
 
+@code {
```
dvoituron, line 4: I think using `yyyyy` is a good solution. But could you move this method to the .cs class.
Except for the demo site, where we prefer clear explanations, in the Core project we want to put the C# code in the Behind (.cs) part and the HTML tags in the razor file.

src/Core/Components/DateTime/FluentDatePicker.razor.cs
```diff
@@ -1,4 +1,6 @@
 using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
```
dvoituron, line 2: This line is not required.

src/Core/Components/DateTime/FluentDatePicker.razor.cs
```diff
@@ -39,7 +41,13 @@ public FluentDatePicker()
 
     protected override string? FormatValueAsString(DateTime? value)
     {
-        return Value?.ToString(Culture.DateTimeFormat.ShortDatePattern, Culture);
+        return Value?.ToString(View switch
```
dvoituron, line 44: This code is duplicated with `FormatAccardingToView` method

src/Core/Components/DateTime/FluentDatePicker.razor.cs
```diff
@@ -76,7 +84,13 @@ protected Task OnSelectedDateAsync(DateTime? value)
 
     protected override bool TryParseValueFromString(string? value, out DateTime? result, [NotNullWhen(false)] out string? validationErrorMessage)
     {
+        if (Regex.IsMatch(value ?? "", @"^\d{4}$"))
```
dvoituron, line 87: That's not correct: you must do that only when "View = Years".
And you can simplify the code using a `int.TryParse`.

Example
```csharp
if (View == CalendarViews.Years && int.TryParse(value, out int year))
{
    value = new DateTime(year, 1, 1).ToString(Culture.DateTimeFormat.ShortDatePattern);
}
```

dvoituron (changes requested):
Thanks for the updates. 
Don't forget to add some Unit Tests.

src/Core/Components/DateTime/FluentDatePicker.razor.cs
StefanJansson: Added two unit tests to verify that placeholder get correct value

src/Core/Components/DateTime/FluentDatePicker.razor.cs
StefanJansson: Not interested in it containging exactly 4 digits?

dvoituron (approved):

PR #1909, opened 2024-04-22
[FluentWizard] Add StepSequence attribute.

### LINKED ISSUE #1299: feat: allow navigation via `FluentWizardStep` label
# 🙋 Feature Request

First of all, I love the new `FluentWizard` component! You guys really did a great job!
However, I'm missing one essential feature: navigating to a step by clicking on its label.
I could probably achieve this myself using the `StepTemplate` but I think this is a feature that many people would like to get implemented.

You could then maybe add a "Linear" (or something similar) parameter to the wizard, which would only allow navigating to the previous and next steps of the active step (which is somewhat similar to the current behavior, I guess?).

## 🤔 Expected Behavior

When clicking on a `FluentWizardStep`'s label or its icon, the active step should change to the clicked step (at least, if the step is not disabled).

## 😯 Current Behavior

Currently, the label is just simple text with no functionality.

## 💁 Possible Solution

Adding an `@onclick` handler to a step's icon and title elements that updates the wizard's `Value` to the step's `Index` should hopefully do the trick.

dvoituron:
# [FluentWizard] Add StepSequence attribute.

Addition of a new `StepSequence` attribute to allow direct access to previous steps or all steps.

- `Linear`: The user can go to the next/previous step only, using the **Next**/**Previous** buttons.
- `Any`: The use can go to any steps, clicking on an item.
- `Visited`: The user can go to the next step using the **Next** button, or go to any previous step, already visited.

Of course, only the **Not Disabled** steps are accessible.

> For compatibility reasons, the default mode is **Linear**.

That will fix #1299

## Example

peek


src/Core/Components/Wizard/FluentWizardStep.razor.cs
```diff
@@ -183,4 +184,43 @@ internal async Task InvokeOnSubmitForEditFormsAsync()
...
+    {
+        if (!IsStepClickable)
+        {
+            return;
+        }
+
+        await FluentWizard.GoToStepAsync(Index, validateEditContexts: Index > FluentWizard.Value);
+    }
+
+    private bool IsStepClickable
```
vnbaaij, line 198: We don't do `Is...` or `Has...` naming, right?

vnbaaij (changes requested):

src/Core/Components/Wizard/FluentWizardStep.razor.cs
dvoituron: It's my private method :-) The code is clearer with this

PR #1936, opened 2024-04-25
[DataGrid] Apply ItemSize (CSS height) to cells when grid is virtualized. 

### LINKED ISSUE #1932: fix: DataGrid OnRowFocus is currently only applying to the cell area and now the row area
Consider a DataGrid with a defined height.

If you apply onrowFocus to the data grid, the clickable area is currently set to the cell value:

image

So if you have a DataGrid with 120px height and the cell content does not take up 120px, then you end up with:

image

where only the center area is clickable.

Anywhere within the row should be clickable

vnbaaij:
Fix #1932 by applying height to cell style when grid is virtualized. Before, the height (ItemSize) would only be applied to the rows. With this PR hte height is added to the cellls style as well.

image

### DIFF
--- a/src/Core/Components/DataGrid/FluentDataGridCell.razor.cs
+++ b/src/Core/Components/DataGrid/FluentDataGridCell.razor.cs
@@ -50,7 +50,7 @@ public partial class FluentDataGridCell<TGridItem> : FluentComponentBase
     private InternalGridContext<TGridItem> GridContext { get; set; } = default!;
 
     protected string? StyleValue => new StyleBuilder(Style)
-       .AddStyle("height", $"{GridContext.Grid.ItemSize}px", () => GridContext.Grid.Virtualize && Owner.RowType == DataGridRowType.Default)
+       .AddStyle("height", $"{GridContext.Grid.ItemSize:0}px", () => GridContext.Grid.Virtualize && Owner.RowType == DataGridRowType.Default)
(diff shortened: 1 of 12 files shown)


src/Core/Components/DataGrid/FluentDataGridCell.razor.cs
```diff
@@ -37,6 +43,18 @@ public partial class FluentDataGridCell<TGridItem> : FluentComponentBase
...
     public FluentDataGridRow<TGridItem> Owner { get; set; } = default!;
 
+    /// <summary>
+    /// Gets or sets the owning <see cref="FluentDataGrid{TItem}"/> component
+    /// </summary>
+    [CascadingParameter]
+    private InternalGridContext<TGridItem> GridContext { get; set; } = default!;
+
+    protected string? StyleValue => new StyleBuilder(Style)
+       .AddStyle("height", $"{GridContext.Grid.ItemSize}px", () => GridContext.Grid.Virtualize && Owner.RowType == DataGridRowType.Default)
```
dvoituron, line 53: I don't know why the `ItemSize` is `float` but using this condition, you have to specify the `en-US` culture.
Or maybe better: `$"{GridContext.Grid.ItemSize:0}px"` to take only the integer part.

dvoituron (changes requested):

dvoituron (approved):

PR #1939, opened 2024-04-25
[DataGrid] Add ShowHover parameter and implement row hover effect

### LINKED ISSUE #1933: feat: Provide a row hover effect on DataGrid using same style of sortable column header hover effect
# 🙋 Feature Request

Currently row hover has to be done manually per: https://github.com/microsoft/fluentui-blazor/discussions/1745#discussioncomment-8927074

## 🤔 Expected Behavior

Ability to activate a row hover effect in the same style at the hover effect shown on sortable column headers

## 😯 Current Behavior

No hover effect is shown

## 💁 Possible Solution

Configurable option on data grid to activate hover effect on row

## 🔦 Context

When user is hovering over a row we want to make it clear which row they are hovering on.

This is especially important when using the onRowFocus callback to allow users to select rows.

Current workaround is to implement rowClass and css such as:

css isolation example:

```css
::deep fluent-data-grid-row.myrowclass:hover {
    cursor: pointer;
    background-color: #c0c0c01f;
}
```

app.css example:

```css
fluent-data-grid-row.myrowclass:hover {
    cursor: pointer;
    background-color: #c0c0c01f;
}
```

vnbaaij:
Fix #1933 by adding a ShowHover (default=false) parameter to the `FluentDataGrid`

Typically this parameter will be used together with an OnRowFocus event callback to detect which row the cursor is over. By setting ShowHover to true, the current row will be highlighted. By default the system will use the designated hover color for this but you can specify an alternative color by setting the --datagrid-hover-color CSS variable. 

datagrid-hover


src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -1,9 +1,13 @@
...
+// ------------------------------------------------------------------------
+// MIT License - Copyright (c) Microsoft Corporation. All rights reserved.
+// ------------------------------------------------------------------------
+
 using Microsoft.AspNetCore.Components;
 using Microsoft.AspNetCore.Components.Web.Virtualization;
 using Microsoft.FluentUI.AspNetCore.Components.DataGrid.Infrastructure;
 using Microsoft.FluentUI.AspNetCore.Components.Infrastructure;
 using Microsoft.JSInterop;
+using System.Diagnostics.CodeAnalysis;
```
dvoituron, line 10: Probably this line is not required.

dvoituron (approved):

PR #1945, opened 2024-04-26
[TreeView] Add `Items` and `LazyLoadItems` properties

dvoituron:
# [TreeView] Add `Items` and `LazyLoadItems` properties

In response to several requests for a **TreeView** component that could display a large amount of data, we've added these functionalities:

1. Add an `Items` attribute to bind a list of elements and sub-elements (objects implementing `ITreeViewItem`).
   The included `TreeViewItem` class already implements this interface. You can use it.

    ```csharp
    public interface ITreeViewItem
    {
        // Gets or sets the unique identifier of the tree item.
        string Id { get; set; }
        
        // Gets or sets the text of the tree item.
        string Text { get; set; }
        
        // Gets or sets the sub-items of the tree item.
        IEnumerable<ITreeViewItem>? Items { get; set; }
        
        // Gets or sets the <see cref="Icon"/> displayed at the start of tree item,
        // when the node is collapsed.
        // If this icon is not set, the <see cref="IconExpanded"/> will be used.
        Icon? IconCollapsed { get; set; }
        
        // Gets or sets the <see cref="Icon"/> displayed at the start of tree item,
        // when the node is expanded.
        // If this icon is not set, the <see cref="IconCollapsed"/> will be used.
        Icon? IconExpanded { get; set; }
        
        // When true, the control will be immutable by user interaction.
        // See <see href="https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled">disabled</see> HTML attribute for more information.
        bool Disabled { get; set; }
        
        // Gets or sets a value indicating whether the tree item is expanded.
        bool Expanded { get; set; }
        
        // Gets or sets the action to be performed when the tree item is expanded or collapsed
        Func<TreeViewItemExpandedEventArgs, Task>? OnExpandedAsync { get; set; }
    }
    ```

2. Add a `LazyLoadItems` attribute to render HTML only when nodes are expanded.

3. Add an `ItemTemplate` to allow complete customization of each element.

4. Add the bindable `SelectedItem` property to manage the selected `ITreeViewItem`.

5. Add the `OnExpandedAsync` method to enable dynamic loading of sub-items. 
    For example, to display an "unlimited" tree view as in this second example.

## Example

```razor
<FluentTreeView Items="@Items" @bind-SelectedItem="@SelectedItem" LazyLoadItems="true">
    <ItemTemplate>
        @context.Text
        <span style="color: var(--accent-fill-rest); margin: 0 4px;">
            @(context.Items == null ? "" : $"[{context.Items.Count()}]")
        </span>
    </ItemTemplate>
</FluentTreeView>
```

peek

## Example: managing a list of "non-predefined" items.

```razor
<FluentTreeView Items="@Items" @bind-SelectedItem="@SelectedItem" />

@code
{
    private ITreeViewItem? SelectedItem;
    private IEnumerable<ITreeViewItem>? Items = new List<ITreeViewItem>();

    protected override void OnInitialized() => Items = GetItems();

    private Task OnExpandedAsync(TreeViewItemExpandedEventArgs e)
    {
        if (e.Expanded)
        {
            e.CurrentItem.Items = GetItems();
        }
        else
        {
            // Remove sub-items and add a "Fake" item to simulate the [+]
            e.CurrentItem.Items = TreeViewItem.LoadingTreeViewItems;
        }

        return Task.CompletedTask;
    }

    private IEnumerable<ITreeViewItem> GetItems()
    {
        var nbItems = Random.Shared.Next(3, 9);

        return Enumerable.Range(1, nbItems).Select(i => new TreeViewItem()
            {
                Text = $"Item {Random.Shared.Next(1, 9999)}",
                OnExpandedAsync = OnExpandedAsync,
                Items = TreeViewItem.LoadingTreeViewItems, // "Fake" sub-item to simulate the [+]
            }).ToArray();
    }
}
```

TreeView-Unlimited

## Unit Tests

image
image


src/Core/Components/TreeView/FluentTreeItem.razor.cs
```diff
@@ -6,6 +6,20 @@ public partial class FluentTreeItem : FluentComponentBase, IDisposable
...
+    public FluentTreeItem()
+    {
+        Id = Identifier.NewId();
+    }
+
+    /// <summary>
+    /// Gets or sets the list of sub-items to bind to the tree item
+    /// </summary>
+    [Parameter]
+    public IEnumerable<ITreeViewItem>? Items { get; set; }
```
verdie-g, line 21: What about `Children`, `ChildrenItems`, `SubItems`?

dvoituron: What do you mean? About the property name?
- `Children` and `ChildrenItems` are too close to `ChildContent` (I think)
- `SubItems` why not

But why ? :-)

verdie-g:
Does it also solve the very slow select of a TreeItem?

dvoituron:
> Does it also solve the very slow select of a TreeItem?

You can try with your own sample, but my tests were based on over 120,000 items :-)

src/Core/Components/TreeView/FluentTreeItem.razor.cs
verdie-g: Yes for the property name. I think it could clear up some confusion to have an `Items` property in a class named `Item`

src/Core/Components/TreeView/FluentTreeView.razor.cs
```diff
@@ -49,6 +54,18 @@ public partial class FluentTreeView : FluentComponentBase, IDisposable
...
 
+    /// <summary>
+    /// Gets or sets the template for rendering tree items.
+    /// </summary>
+    [Parameter]
+    public RenderFragment<ITreeViewItem>? ItemTemplate { get; set; }
+
+    /// <summary>
+    /// Gets or sets whether the tree should use lazy loading when expanding nodes
+    /// </summary>
```
verdie-g, line 65: Just from the documentation it is not super clear what this parameter does. I'm thinking maybe it will call a user-provided method to get the next nodes or only render the expanded nodes. Though there is already `RenderCollapsedNodes` that should take care of that.

dvoituron: It's a `boolean`, so the list of items is already known, but is not rendered (HTML) if the current element is reduced.
We could perhaps add another method to be called when the Expand event is triggered.

verdie-g: So what's the difference with `RenderCollapsedNodes` ? 🤔

dvoituron: - `RenderCollapsedNodes` is an attribute included in the TreeView Web component. The goal seems similar, but it doesn't work properly with this attribute. probably because it keeps all the data in memory (JavaScript).

- This new `LazyLoadItems` works with Blazor's SignalR rendering and doesn't send sub-elements until they're requested. In the example above, only the first 15 items are actually present in the HTML page, allowing it to be displayed instantly on screen.

I think we should tag `RenderCollapsedNodes` as **Obsolete** and suggest using `LazyLoadItems`.
@vnbaaij What do you think?

verdie-g: > LazyLoadItems works with Blazor's SignalR rendering

Does it mean it would not work for standalone wasm?

dvoituron: I don't understand

vnbaaij: > I think we should tag `RenderCollapsedNodes` as **Obsolete** and suggest using `LazyLoadItems`.

I think that makes sense to do so, yes

verdie-g: Blazor supports several hosting modes:
- Blazor Server
- Blazor WebAssembly
- Blazor Hybrid

In Blazor WebAssembly, there are no SignalR server. Would the LazyLoadItems feature still work then?

dvoituron: Of course, it's Blazor's job to render Blazor Server or Blazor WASM code correctly :-)

verdie-g: Got it thanks!

verdie-g: > I think we should tag RenderCollapsedNodes as Obsolete and suggest using LazyLoadItems.

Alternatively, if the 2 features have similar goal but different implementations it could make sense not to introduce `LazyLoadItems` but to change `RenderCollapsedNodes` implementation with the new one.

dvoituron: Except if some people are already used this feature and don't want to change the unit test, for example, or other stuff.

verdie-g:
Is this change not planned in the end? :(

dvoituron:
I think it's a mistake... @vnbaaij I'm reopening the PR and restoring my branch ;-)

vnbaaij:
> Is this change not planned in the end? :(

Sorry, was spring-cleaning the branches a bit too enthousiastically  🧹🪣🧽

verdie-g:
What are the next steps to go forward with that PR? Is there anything I could to help?

dvoituron:
I will finish the TODO list in few days 😊

src/Core/Components/TreeView/FluentTreeView.razor.cs
```diff
@@ -10,11 +10,34 @@ public partial class FluentTreeView : FluentComponentBase, IDisposable
...
+    [Parameter]
+    public EventCallback<ITreeViewItem?> SelectedItemChanged { get; set; }
+
     /// <summary>
     /// Gets or sets whether the tree should render nodes under collapsed items
     /// Defaults to false
     /// </summary>
     [Parameter]
+    [Obsolete("Please, prefer to use the `LazyLoadItems` attribute.")]
```
vnbaaij, line 40: [Obsolete("Please use the `LazyLoadItems` parameter instead.")]

src/Core/Components/TreeView/TreeViewItemExpandedEventArgs.cs
```diff
@@ -0,0 +1,28 @@
+// ------------------------------------------------------------------------
+// MIT License - Copyright (c) Microsoft Corporation. All rights reserved.
+// ------------------------------------------------------------------------
+
+namespace Microsoft.FluentUI.AspNetCore.Components;
+
+/// <summary>
+/// Class that contains the event arguments for the <see cref="ITreeViewItem.OnExpanded"/> event.
```
vnbaaij, line 8: needs to point to `OnExpandedAsync`

vnbaaij (changes requested):

vnbaaij (approved):

PR #1952, opened 2024-04-26
[DataGrid] Multiselect feature, adding a `SelectColumn`

dvoituron:
# [DataGrid] Multiselect feature, adding a `SelectColumn`

There are two ways of managing this `SelectColumn`. 
1. Either by supplying a list of data via `Items` and letting the grid manage the selected rows entirely, via `SelectedItems`. property.

```razor
<FluentDataGrid Items="@People" ShowHover="true" TGridItem="Person">
    <!-- THIS LINE TO HAVE MULTISELECT -->
    <SelectColumn TGridItem="Person"
                    SelectMode="@DataGridSelectMode.Multiple"
                    @bind-SelectedItems="@SelectedItems" />

    <PropertyColumn Width="100px" Property="@(p => p.PersonId)" Title="ID" />
    <PropertyColumn Width="300px" Property="@(p => p.Name)" />
    <PropertyColumn Width="150px" Property="@(p => p.BirthDate)" Format="yyyy-MM-dd" Sortable="true" />
</FluentDataGrid>
```

2. Or by manually managing the way selected lines are saved, via `Property`, `OnSelect` and `SelectAll`.
This requires more configuration but is probably more interesting when using `Virtualize` or when directly managing a custom `IsSelected` property.

   - `Property`: Function to be executed to display the checked/unchecked icon, depending of you data model.
   - `OnSelect`: Action to be executed when the row is selected or unselected.
   - `SelectAll`: Value indicating whether the [All] checkbox is selected.

```razor
<FluentDataGrid Items="@People" ShowHover="true" TGridItem="Person">
    <!-- THIS LINE TO HAVE MULTISELECT -->
    <SelectColumn TGridItem="Person"
                    SelectMode="@DataGridSelectMode.Multiple"
                    Property="@(e => e.Selected)"
                    OnSelect="@(e => e.Item.Selected = e.Selected)"
                    SelectAll="@(People.All(p => p.Selected))"
                    SelectAllChanged="@(all => People.ToList().ForEach(p => p.Selected = (all == true)))" />

    <PropertyColumn Width="100px" Property="@(p => p.PersonId)" Title="ID" />
    <PropertyColumn Width="300px" Property="@(p => p.Name)" />
    <PropertyColumn Width="150px" Property="@(p => p.BirthDate)" Format="yyyy-MM-dd" Sortable="true" />
</FluentDataGrid>
```

These example are using a class containing:

```csharp
    DataGridSelectMode Mode = DataGridSelectMode.Multiple;
    IEnumerable<Person> SelectedItems = People.Where(p => p.Selected);

    record Person(int PersonId, string Name, DateOnly BirthDate) 
    {
        public bool Selected { get; set; }
    };
```

### Example

peek

# Unit Tests

image


vnbaaij:
Love this!

Would probably be good to store the selected state of rows centrally, for example in the `InternalGridContext`
And also to have `Selected`\`SelectedChanged` on the DataGridRow so that an initial state can be set programmatically

dvoituron:
> Would probably be good to store the selected state of rows centrally, for example in the `InternalGridContext`
> And also to have `Selected``SelectedChanged` on the DataGridRow so that an initial state can be set programmatically

Columns and attributes already updated

As far as storing the selected state is concerned, this could be an additional feature, but it would probably cause a lot of problems. For example, when using the virtualization function, I'm not sure that all "previous" items are stored in this internal memory.
I will check.

dvoituron:
> Finally!! Looking great! Always been using some ugly workarounds to get similar behavior. Could you maybe add the option for an checkbox in the header? Like maybe a parameter `SelectMode` with `SelectMode.Single` and `SelectMode.Multiple`, where the latter adds the header checkbox? And won't this column allow for a bindable "SelectedItems" HashSet in the DataGrid?

On the same remark, adding a checkbox in the header seems like a good thing if we have all the data in memory. But we could add this option and an EventCallback to allow the developer to select all the elements.

About Single/Multiple, yes that will be done :-)

dvoituron:
@vnbaaij @hksalessio I renamed to these attributes: `Property` and `OnSelect`.
And I added a bindable `SelectAll` property which display a [x] in the column header. So the dev can intercept the `SelectAllChanged` (or `:after` event) to update his model.

Example
```razor
SelectAllChanged="@(all => people.ToList().ForEach(p => p.Selected = (all == true)))"
```

sardar97:
Nice work @dvoituron thank u for creating this feature.
I just have one suggestion when you select one row or more the select all checkbox should change like a (minus) inside the select all checkbox and when you select all it should change to select all. for more explanation I will show it with 2 image.

when you select 1 row or more :
select one or more

when you select all rows :
select all

dvoituron:
> Nice work @dvoituron thank u for creating this feature. I just have one suggestion when you select one row or more the select all checkbox should change like a (minus) inside the select all checkbox and when you select all it should change to select all. for more explanation I will show it with 2 image.

That's not the case in Windows Explorer, for example

image

sardar97:
so we could not have it even with CSS ?

sardar97:
and one more suggestion is very important please I haven't done it already! when the user click the row I mean anywhere on the it should select not only click on exactly  the checkbox.

dvoituron:
> and one more suggestion is very important please if you haven't done it already! when the user click the row I mean anywhere on the it should select not only click on exactly the checkbo

Already done and included in the GIF above :-)

dvoituron:
Updated with 
- `SelectedItems` Gets or sets the list of selected items (in addition with `Property` and `OnSelect`).
  I will write a documentation to explain different uses.
- `SelectMode` selection mode (Single or Multiple).
- `IconIndeterminate` Icon to be rendered when some but not all rows are selected.

GIFs in summary are also updated.

dvoituron:
I finished the development and updated the description (top of this PR).
I still need to add the unit tests to finalize the PR.

src/Core/Components/DataGrid/Columns/SelectColumn.cs
```diff
@@ -0,0 +1,364 @@
...
+
+            RefreshHeaderContent();
+        }
+    }
+
+    /// <summary>
+    /// Gets or sets the Icon to be rendered when the row is non selected.
+    /// </summary>
+    [Parameter]
+    public required Icon IconUnchecked { get; set; } = new CoreIcons.Regular.Size20.CheckboxUnchecked().WithColor(Color.FillInverse);
```
vnbaaij, line 83: FluentUI 9 uses a radio for single select row checked/unchecked state. As we have those icons as well, might make sense to use these. And use Checkbox for multi-select. Probably easiest to add two new parameters (SingleSelectIconUnchecked, SingleSelectIconChecked) and rename the existing ones (IconUnchecked -> MultiSelectIconUnchecked, IconChecked -> MultiSelectIconChecked and IconIndeterminate -> MultiSelectIconIndeterminate  )
image

src/Core/Components/DataGrid/FluentDataGridRow.razor.cs
```diff
@@ -99,6 +100,58 @@ private async Task HandleOnCellFocusAsync(DataGridCellFocusEventArgs args)
...
+                await Owner.Grid.OnRowClick.InvokeAsync(row);
+            }
+
+            if (row != null && row.RowType == DataGridRowType.Default)
+            {
+                foreach (var selColumn in Owner.Grid.SelectColumns)
+                {
+                    await selColumn.AddOrRemoveSelectedItemAsync(Item);
+                }
+            }
```
vnbaaij, line 119: If we go with single SelectColumn in FluntDataGrid, change to
```
if (row != null && row.RowType == DataGridRowType.Default && Owner.Grid.SelectColumn is not null)
{
    await Owner.Grid.SelectColumn.AddOrRemoveSelectedItemAsync(Item);
}
```

src/Core/Components/DataGrid/FluentDataGridRow.razor.cs
```diff
@@ -99,6 +100,58 @@ private async Task HandleOnCellFocusAsync(DataGridCellFocusEventArgs args)
...
+        {
+            if (Owner.Rows.TryGetValue(rowId, out var row))
+            {
+                if (row != null && row.RowType == DataGridRowType.Default)
+                {
+                    foreach (var selColumn in Owner.Grid.SelectColumns)
+                    {
+                        await selColumn.AddOrRemoveSelectedItemAsync(Item);
+                    }
+                }
```
vnbaaij, line 149: Sme here:
```
if (row != null && row.RowType == DataGridRowType.Default && Owner.Grid.SelectColumn is not null)
                {
                    await Owner.Grid.SelectColumn.AddOrRemoveSelectedItemAsync(Item);
                }
```

vnbaaij (changes requested):

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -153,10 +167,16 @@ public partial class FluentDataGrid<TGridItem> : FluentComponentBase, IHandleEve
...
     [Parameter] public RenderFragment? LoadingContent { get; set; }
+
     [Inject] private IServiceProvider Services { get; set; } = default!;
     [Inject] private IJSRuntime JSRuntime { get; set; } = default!;
     [Inject] private IKeyCodeService KeyCodeService { get; set; } = default!;
 
+    /// <summary>
+    /// Gets the first (optional) SelectColumn
+    /// </summary>
+    internal IEnumerable<SelectColumn<TGridItem>> SelectColumns => _columns.Where(col => col is SelectColumn<TGridItem>).Cast< SelectColumn<TGridItem>>();
```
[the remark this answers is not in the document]
dvoituron: I prefer to use this feature as a "normal" column. And you can have several columns.
For now, you can have several selection columns :-) 

image

A solution could be to raise an exception when we try to use multiple `SelectColumn`... but do we need to do that? (Performances will not impacted)

src/Core/Components/DataGrid/FluentDataGridRow.razor.cs
dvoituron: As explained, I prefer to keep the "normal" operation of the columns. In most cases, the **foreach** will be performed on a item, but this isn't disturbing and shouldn't affect performance.

src/Core/Components/DataGrid/FluentDataGridRow.razor.cs
dvoituron: Same :-)

src/Core/Components/DataGrid/Columns/SelectColumn.cs
dvoituron: Done... and GIF above updated.

vnbaaij (approved):

boyanmihnev92:
Will this feature be release any time soon?

chandrasbd:
added to milestone [v4.7.2](https://github.com/microsoft/fluentui-blazor/milestone/11?closed=1)

matasbern:
What if I want to only Select item on Checkbox rather than Full row? Is there a way to do it? SOme of my columns contain custom data and <a>DATA</a> that user has @onclick on them.. Trying to solve it?

dvoituron:
> What if I want to only Select item on Checkbox rather than Full row? Is there a way to do it? SOme of my columns contain custom data and DATA that user has @OnClick on them.. Trying to solve it?

Today, this is not possible. Unless you create your own "SelectColumn" using a `TemplateColumn`.

hksalessio:
> What if I want to only Select item on Checkbox rather than Full row? Is there a way to do it? SOme of my columns contain custom data and DATA that user has @OnClick on them.. Trying to solve it?

I am kind of working on a pull request that could solve this problem. Shoud be ready for review later today.

matasbern:
Thanks for such a rapid response guys.
Just one question, from what I understand it works by design.. but it seems to be wrong?
If I open page which has multiselect, my first select gets added to my IEnumerable<Product> SelectedItems, but UI rerenders and removes the checkbox state?
You can see what I mean here. Is there a way to stop this? 
06 06 2024_11 06 47_REC-ezgif com-video-to-gif-converter

dvoituron:
> EDIT EDIT: Ok, it seems its happening only if I'm using ItemsProvider, Items with IQueryable works fine. Seems like first click on ItemsProvider MultiSelect triggers a call to API. Any ideas how to overcome this?

Do you use manual management via `Property` and `OnSelect`? Because an `ItemsProvider` doesn't store your data (and your selection in memory). You have to use "manual management" to store this information in your objects.

matasbern:
> > EDIT EDIT: Ok, it seems its happening only if I'm using ItemsProvider, Items with IQueryable works fine. Seems like first click on ItemsProvider MultiSelect triggers a call to API. Any ideas how to overcome this?
> 
> Do you use manual management via `Property` and `OnSelect`? Because an `ItemsProvider` doesn't store your data (and your selection in memory). You have to use "manual management" to store this information in your objects.

Ok I see, thanks for your reply. Is there a way to somehow stop triggering Grid refresh on any 'onclick' event that triggers it? I have noticed it also happens if I add a button to TemplateColumn, once I clicked on a button first time - it reloads the data. Any consequent clicks don't do it.

vnbaaij:
We can't change that on our side. You could optimize on your side and determine is the reload is necessary.

Issue #1974, opened 2024-04-30
fix: AppBar shows active menu apps for onClick with `href="@(null)"`


StephenOTT:
# 🐛 Bug Report

Based on the docs for AppBar, it would seem to indicate that using Href="@(null)" will let you use the OnClick function.

But doing this on the main page of a app where there is no path, other than the domain, causes the href=null "apps" to show as active.

<img width="403" alt="image" src="https://github.com/microsoft/fluentui-blazor/assets/1994838/bdd400de-f56f-4e52-a679-6a49dca8e366">

## 💻 Repro or Code Sample

Use the demo app bar (https://www.fluentui-blazor.net/AppBar#withonclickevent)  on a `/` page:

such as

```
@page "/"
@inject IDialogService DialogService

<PageTitle>Home</PageTitle>

<FluentStack Orientation="Orientation.Vertical" Style="height: 330px;">
    <FluentAppBar Style="height: 100%;">
        <FluentAppBarItem Href="/"
                          Match="NavLinkMatch.All"
                          IconRest="HomeIcon()"
                          IconActive="HomeIcon(active: true)"
                          Text="Home"
                          OnClick="HandleOnClick" />
        <FluentAppBarItem Href="/AppBar"
                          IconRest="AppBarIcon()"
                          IconActive="AppBarIcon(active: true)"
                          Text="AppBar"
                          OnClick="HandleOnClick" />
        <FluentAppBarItem Href="@(null)"
                          IconRest="WhatsNewIcon()"
                          IconActive="WhatsNewIcon(active: true)"
                          Text="What's New'"
                          OnClick="ShowSuccessAsync" />
        <FluentAppBarItem Href="@(null)"
                          IconRest="IconsIcon()"
                          IconActive="IconsIcon(active: true)"
                          Text="Icons"
                          OnClick="ShowWarningAsync" />
        <FluentAppBarItem Href="/Dialog"
                          IconRest="DialogIcon()"
                          IconActive="DialogIcon(active: true)"
                          Text="Dialog"
                          OnClick="HandleOnClick" />
    </FluentAppBar>
</FluentStack>

...


```

## 🤔 Expected Behavior

using the null for href should not activate the "app".

------

on a related note, is there a way through runtime code to set the active/inactive status of a app item?

dvoituron:
@vnbaaij The `Href` property could be "not required" and "nullable". What do you think?

```csharp
[Parameter]
public required string? Href { get; set; }
```

[...]

PR #1976, opened 2024-04-30
[AppBar] Allow for omitting Href on AppBarItems and don't show active status when Href is null or not specified

vnbaaij:
Changes for fixing #1974 
We have changed the `Href` parameter on the AppBarItem to allow for `null` valure and make it not-`[EditorRequired]` anymore.

In the AppBarItem we now check if `Href` is null. If so, we'll add a class to the AppBarItem indicating it is a local link. We use `NavLink` internally, so an `a` is still rendered and an `active` class is still added. Based on that local class being added, we will remove the CSS that gets applied for 'active' state (i.e remove the accent colored bar)

image

### DIFF
--- a/src/Core/Components/AppBar/FluentAppBarItem.razor.cs
+++ b/src/Core/Components/AppBar/FluentAppBarItem.razor.cs
@@ -78,7 +78,7 @@ public partial class FluentAppBarItem : FluentComponentBase, IDisposable
     public bool? Overflow { get; private set; }
 
     internal string? ClassValue => new CssBuilder("fluent-appbar-item")
-        .AddClass("fluent-appbar-item-local", when: Href == null)
+        .AddClass("fluent-appbar-item-local", when: string.IsNullOrEmpty(Href))
         .AddClass(Class)
         .Build();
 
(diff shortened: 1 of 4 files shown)


src/Core/Components/AppBar/FluentAppBarItem.razor.cs
```diff
@@ -78,6 +78,7 @@ public partial class FluentAppBarItem : FluentComponentBase, IDisposable
     public bool? Overflow { get; private set; }
 
     internal string? ClassValue => new CssBuilder("fluent-appbar-item")
+        .AddClass("fluent-appbar-item-local", when: Href == null)
```
dvoituron, line 81: Just to verify: only null, not `string.IsNullOrEmpty` ?

dvoituron (approved):

src/Core/Components/AppBar/FluentAppBarItem.razor.cs
vnbaaij: IsNullOrEmpy is better, yes. Will change

PR #1994, opened 2024-05-01
[InputLabel] Add Orientation parameter 

vnbaaij:
As stated in the Fluent Desing, a label should be placed above the input component it belongs to. It also says:

>Optionally, you can left align labels if space is limited and if fields are of a consistent type and width. Use caution when left aligning the labels as it can decrease readability.

This could of course also be used in, for example a Toolbar. To facilitate placing the label in front of the input instead of above, an `Orientation` parameter is added to the component. When setting its value to `Orientation.Horizontal` (default is `Orientation.Vertical`) an extra attribute `orientation` with value `horizontal` is emitted which clears out the margin-bottom value of the label element. When this is used in a toolbar the label can now be lined up in the center without needing additional CSS changes in the application.

image

Also (more) unit tests created:
image


src/Core/Components/Label/FluentInputLabel.razor
```diff
@@ -1,15 +1,15 @@
-@namespace Microsoft.FluentUI.AspNetCore.Components
+﻿@namespace Microsoft.FluentUI.AspNetCore.Components
 @inject LibraryConfiguration LibraryConfiguration
 
 
 @if (!string.IsNullOrEmpty(Label) || ChildContent is not null)
 {
-    <label for="@ForId" class="fluent-input-label" @attributes="@AdditionalAttributes" aria-label="@AriaLabel">
+    <label for="@ForId" class="@ClassValue" @attributes="@AdditionalAttributes" aria-label="@AriaLabel">
```
dvoituron, line 7: That could be better to keep `fluent-input-label` and to use a new HTML attribute `orientation`. 

For the future, that will be easier to configure other styles.

dvoituron (changes requested):
Great new feature!

Only one comment. 

And with Unit Tests 😍

src/Core/Components/Label/FluentInputLabel.razor
vnbaaij: Done!

src/Core/Components/Label/FluentInputLabel.razor
```diff
@@ -1,15 +1,17 @@
...
+﻿@namespace Microsoft.FluentUI.AspNetCore.Components
 @inject LibraryConfiguration LibraryConfiguration
 
 
 @if (!string.IsNullOrEmpty(Label) || ChildContent is not null)
 {
-    <label for="@ForId" class="fluent-input-label" @attributes="@AdditionalAttributes" aria-label="@AriaLabel">
+    var orientation = Orientation == Orientation.Horizontal ? Orientation.ToAttributeValue() : null; 
+
+    <label for="@ForId" class="fluent-input-label" orientation=@orientation @attributes="@AdditionalAttributes" aria-label="@AriaLabel">
```
dvoituron, line 9: A detail, but I think it could be better to always use double-quote for all attributes and the variable must start with an upper-case.

```xml
`orientation="@Orientation"`
```

vnbaaij: quotes added. In this case it is a local var (see line above). I added an '_' to make that distinction clearer

dvoituron (approved):

Issue #2046, opened 2024-05-14
fix: DatePicker year selection not working with component within EditForm


msb-incom:
# 🐛 Bug Report

When using a DatePicker within an EditForm, selecting a year using the year-view of a DatePicker will default to the current year. This does not happen when using the DatePicker outside an EditForm.

## 💻 Repro or Code Sample

This can be seen in the example on https://www.fluentui-blazor.net/Forms

## 🤔 Expected Behavior

After selecting a `year`, I would suspect the date to change to `DateTime(year, 1, 1)` and to open the month view for the selected year.

## 😯 Current Behavior

When selecting e.g. "2026"
![image](https://github.com/microsoft/fluentui-blazor/assets/97457019/94786295-e5f6-429a-a1a6-4f8cd2b59d4a)

.. you will return to current date:
![image](https://github.com/microsoft/fluentui-blazor/assets/97457019/6286151c-1434-47ac-b2aa-caeae7269e85)

## 🔦 Context

In the current state, the users of my application are unable to use the year-view in the DatePicker.

## 🌍 Your Environment

* OS & Device: Windows
* Browser: Microsoft Edge, Google Chrome
* .NET and Fluent UI Blazor library Version: 8.0.2 and 4.7.2

msb-incom:
I submitted a PR that seeks to fix the issue 🙂

PR #2047, opened 2024-05-14
Fix DatePicker/Calendar not updating under EditContext

msb-incom:
# Pull Request

## 📖 Description

When using a FluentDatePicker or FluentCalender inside an EditContext, there are issues with selecting a new value in the Year-view of the calendar. My diagnosis suggests that this comes down to a race condition-like issue between invoking ValueChanged and also notifying the EditContext about changes which will propagate the "old" value to the component through the parameters being updated.

The changes I propose changes the approach for updating value of the FluentCalendar (and thus FluentDatePicker) by utilizing `SetCurrentValueAsync` available on `FluentInputBase`. This approach is similar to other components where `SetCurrentValueAsync` is called as part of `ChangeHandlerAsync`.

The changes proposed fixes the issue.

### 🎫 Issues

https://github.com/microsoft/fluentui-blazor/issues/2046

## 📑 Test Plan

I tested all the DatePicker/Calendar examples in the sample project.

### General

### Component-specific

### DIFF
--- a/examples/Demo/Shared/Microsoft.FluentUI.AspNetCore.Components.xml
+++ b/examples/Demo/Shared/Microsoft.FluentUI.AspNetCore.Components.xml
@@ -2251,11 +2251,6 @@
             Defines the appearance of the <see cref="T:Microsoft.FluentUI.AspNetCore.Components.FluentCalendar"/> component.
             </summary>
         </member>
-        <member name="P:Microsoft.FluentUI.AspNetCore.Components.FluentCalendarBase.Value">
-            <summary>
-            Gets or sets the selected date (two-way bindable).
-            </summary>
-        </member>
         <member name="M:Microsoft.FluentUI.AspNetCore.Components.FluentCalendarBase.OnSelectedDateHandlerAsync(System.Nullable{System.DateTime})">
             <summary />
         </member>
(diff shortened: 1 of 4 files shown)


msb-incom:
> @microsoft-github-policy-service agree

dvoituron:
That sounds about right. Give me time to check all this before approving the PR.
Good work!

msb-incom:
It seems it is in fact not passing tests. The reason seems to be that selecting the same month after going to month view will give the same value causing `SetCurrentValueAsync` to return early without notifying about the field change. I'll look into alternative options 🙂

msb-incom:
I fixed above mentioned problem by extending and utilizing `OnSelectedDateHandlerAsync` rather than using `SetCurrentValueAsync`. This is now more similar to how it was before my change.
However, an alternative could be to extend `FluentInputBase` with the `CheckIfSelectedValueHasChanged` from `FluentCalendarBase` and use that parameter when checking for value change. However this only seemed to be a requirement for the FluentCalendar, so it didn't seem appropriate to let that requirement go all the way up and mess with the design of the `FluentInputBase` 🙂

dvoituron (approved):

msb-incom:
@dvoituron  Thanks for taking the time to review the PR! And let me also use this opportunity to thank you and the other collaborators for providing an incredibly useful library that makes for very swift Blazor development 💪

PR #2052, opened 2024-05-15
[FluentAutocomplete] Add `ImmediateDelay` parameter

### LINKED ISSUE #1951: feat: add Immediate/ImmediateDelate to/in FluentAutocomplete
# 🙋 Feature Request

`FluentAutocomplete` does not have `Immediate` / `ImmediateDelay` properties.  Because of this calls are being made to an api for each keypress and it's a bit chatty. 😊

## 🤔 Expected Behavior

Less chatty API calls with each keypress

## 😯 Current Behavior

An API call with each keypress 😞

## 💁 Possible Solution

Implement Immediate/Delay as it is with `FluentTextField` which works perfectly. 🙏

## 🔦 Context

Attempting to reduce the amount of calls to API. 👍

## 💻 Examples

`FluentTextField.Immediate` and `FluentTextField.ImmediateDelay` works amazingly well. 🙏  Thank you for any consideration!

vnbaaij:
Fix #1951 
Add an `ImmediateDelay parameter. Idea was to pass this parameter through to the internal TextField but this proved not to be necessary. Kept the same name though for consistency.
An `Immediate` parameter has not been added because the default behavior on the internal text filed is already to respond to oninput (onchange is not implemented in this component)

### DIFF
--- a/src/Core/Components/List/FluentAutocomplete.razor.cs
+++ b/src/Core/Components/List/FluentAutocomplete.razor.cs
@@ -16,6 +16,7 @@ public partial class FluentAutocomplete<TOption> : ListComponentBase<TOption> wh
 
     public new FluentTextField? Element { get; set; } = default!;
     private Virtualize<TOption>? VirtualizationContainer { get; set; }
+    private readonly Debouncer _debouncer = new();
 
     /// <summary>
     /// Initializes a new instance of the <see cref="FluentAutocomplete{TOption}"/> class.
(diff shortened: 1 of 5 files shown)


src/Core/Components/List/FluentAutocomplete.razor.cs
```diff
@@ -254,6 +261,8 @@ private string ComponentWidth
     /// <summary />
     protected async Task InputHandlerAsync(ChangeEventArgs e)
     {
+
+
```
dvoituron, line 265: Could you remove these multiple blank lines? :-)

dvoituron (approved):
One comment only (details)

PR #2060, opened 2024-05-16
[DataGrid] Make SelectColumn work when using ItemsProvider

vnbaaij:
Some changes needed in SelectColumn logic and the DataGrid itself (by storing items and total item count in the internal gri context).

`IssueTester` page can be used to test the changes

**do not** auto-merge this. Want to clean out `IssueTester` before merging


dvoituron:
I think the problem (and the solution) is not so complex :-)
The problem is due to the `Grid.ItemsProvider` which not detected.

I can create a PR with that, the code is already on my machine :-)

image

vnbaaij:
I think with your code you can't do 'select all' when using an itemsprovider

dvoituron:
Applying these changes on the code from the linked issue, solve the problem. What do you you want to do in addition ? The property selection is made by the "dev" code, not by the component.

dvoituron:
Maybe to display the AllSelected icon and not the Intermediate icon only ?

src/Core/Components/DataGrid/FluentDataGrid.razor
```diff
@@ -18,7 +18,7 @@
                           grid-template-columns=@_internalGridTemplateColumns
                           class="@GridClass()"
                           style="@Style"
-                          aria-rowcount="@(_ariaBodyRowCount + 1)"
+                          aria-rowcount="@(_internalGridContext.TotalItemCount + 1)"
```
dvoituron, line 21: Why did you add one item: `+ 1` generates an invalid value.

dvoituron (changes requested):

src/Core/Components/DataGrid/FluentDataGrid.razor
vnbaaij: The +1 was already there. Did not change that with this PR

dvoituron (approved):

PR #2083, opened 2024-05-22
[AppBar] Introduce IAppBarItem to enable better extensibility

vnbaaij:
Add an `IAppBarItem` interface to make adding different items to the AppBar more flexible


src/Core/Components/AppBar/IAppBarItem.cs
```diff
@@ -0,0 +1,63 @@
+// ------------------------------------------------------------------------
+// MIT License - Copyright (c) Microsoft Corporation. All rights reserved.
+// ------------------------------------------------------------------------
+using Microsoft.AspNetCore.Components;
+using Microsoft.AspNetCore.Components.Routing;
+
+namespace Microsoft.FluentUI.AspNetCore.Components;
+public interface IAppBarItem
+{
+    public string? Id { get; set; }
```
dvoituron, line 10: Could you add a description?

src/Core/Components/AppBar/IAppBarItem.cs
```diff
@@ -0,0 +1,63 @@
...
+
+    /// <summary>
+    /// Wether this app is outside of visible app bar area.
+    /// </summary>
+    public bool? Overflow { get; set; }
+
+    /// <summary>
+    /// Gets or sets the position of the item in the app bar.
+    /// </summary>
+    public int Order { get; set; }
```
dvoituron, line 56: I don't see this property used in the code.
I prefer to use the order in which the elements were created, rather than having an _Index_ that would allow them to be placed out of order.

dvoituron (changes requested):

src/Core/Components/AppBar/IAppBarItem.cs
vnbaaij: That is why it is still a draft PR 😉

vnbaaij: Example that uses this has been added

src/Core/Components/AppBar/FluentAppBar.razor.cs
```diff
@@ -39,16 +39,19 @@ public partial class FluentAppBar : FluentComponentBase
     [Parameter]
     public EventCallback<bool> PopoverVisibilityChanged { get; set; }
 
+    [Parameter]
+    public IEnumerable<IAppBarItem>? Apps { get; set; }
```
dvoituron, line 43: Could you add a comment for the documentation?

I think that `Items` could be a better name. What is your opinion?

src/Core/Components/AppBar/FluentAppBarItem.razor.cs
```diff
@@ -54,6 +54,12 @@ public partial class FluentAppBarItem : FluentComponentBase, IDisposable
     [Parameter]
     public int Count { get; set; } = 0;
 
+    /// <summary>
+    /// Gets or sets the position of the item in the app bar.
+    /// </summary>
+    [Parameter]
+    public int Order { get; set; }
```
dvoituron, line 61: Isn't it possible to use the order of the list elements?
It's like the user writing `<li>` elements, the order defined will be the order of the result.

dvoituron: After checking the example, I think this property should not be included in the interface. It's the user who chooses to order the elements using one criterion or another.

src/Core/Components/AppBar/IAppBarItem.cs
```diff
@@ -0,0 +1,67 @@
...
+
+    /// <summary>
+    /// Gets or sets the tooltip to show when the item is hovered.
+    /// </summary>
+    public string? Tooltip { get; set; }
+
+    /// <summary>
+    /// Gets or sets the count to show on the item with a <see cref="FluentCounterBadge"/>.
+    /// </summary>
+    public int Count { get; set; }
```
dvoituron, line 50: The `FluentCounterBadge` use a nullable `Count` property. Maybe we have a `int?`?

src/Core/Components/AppBar/IAppBarItem.cs
```diff
@@ -0,0 +1,67 @@
...
+
+    /// <summary>
+    /// Gets or sets the text to show under the icon.
+    /// </summary>
+    public string Text { get; set; }
+
+    /// <summary>
+    /// Gets or sets the tooltip to show when the item is hovered.
+    /// </summary>
+    public string? Tooltip { get; set; }
```
dvoituron, line 45: This property is not used in the code?

src/Core/Components/AppBar/IAppBarItem.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// <summary>
+    /// Gets or sets the URL for this item.
+    /// </summary>
+    public string? Href { get; set; }
+
+    /// <summary>
+    /// Gets or sets how the link should be matched.
+    /// Defaults to <see cref="NavLinkMatch.Prefix"/>.
+    /// </summary>
+    public NavLinkMatch Match { get; set; }
```
dvoituron, line 25: This property is not used in the code?

dvoituron (changes requested):

src/Core/Components/AppBar/FluentAppBar.razor.cs
vnbaaij: Yes, was doubting the name as well. Will change to Items

src/Core/Components/AppBar/FluentAppBarItem.razor.cs
vnbaaij: Yes, they can use the order of the list elements (by just using `Items="the _list"`)

The `Order` parameter is meant to explicitly set a custom order. I'll remove it from the interface as it can easily be  added when creating a class derived from `FluentAppBarItem`. I'll adjust the sample to that

src/Core/Components/AppBar/IAppBarItem.cs
vnbaaij: Yes it is used. Now also in a sample:
image

src/Core/Components/AppBar/IAppBarItem.cs
vnbaaij: Yes, it is being used:
image

src/Core/Components/AppBar/IAppBarItem.cs
vnbaaij: Done

src/Core/Components/AppBar/IAppBarItem.cs
dvoituron: Hups... sorry :-)

dvoituron (approved):

PR #2104, opened 2024-05-26
Add the ability for CssBuilder to properly Add multiple classes without duplicates

StevenTCramer:
# Pull Request

## 📖 Description

1. **Adding AddClasses Method:**
   - Introduced `AddClasses` to handle multiple space-separated CSS classes efficiently, leveraging the `HashSet` to eliminate duplicates.

2. **Refactoring AddClass Methods:**
   - Updated `AddClass` methods to internally call `AddClasses`, ensuring backward compatibility.

3. **Conditional AddClasses Methods:**
   - Added conditional `AddClasses` methods to add classes based on boolean conditions or functions.

These changes optimize class addition by preventing duplicates and improving performance.

### 🎫 Issues

The current AddClass method actually allows for adding multiple but it doesn't split the input and utilize the hashset and thus would allow duplicates.

## 📑 Test Plan

Update CSSBuilder Tests cases to cover 

### General

### Component-specific


vnbaaij (approved):

dvoituron:
This could be an interesting improvement. But why not keep the existing ˋAddClass` method, improving the internal code? And don't add new methods that don't solve the current usage.

The CSS class name must be composed of numbers, hyphens, underscores and letters.

So the existing ˋAddClass` method could be split into spaces, commas, ... to use your code.

StevenTCramer:
* Update AddClass to support single or multiple classes.  
* Ignore invalid classes based on the regex. 
* Updated tests cases to validate.
*  Fixed StyleBuilder Tests copy and paste errors.

tests/Core/Utilities/StyleBuilderTests.cs
```diff
@@ -71,20 +71,6 @@ public void CssBuilder_WithComplexUserStyle()
         Assert.Equal("color: red; font-size: 12px; font-name: courier;", styleBuilder.Build());
     }
 
-    [Fact]
```
dvoituron, line 74: You can perhaps keep (and correct as follows) this (additional) test to check the use of several classes.

```csharp
[Fact]
public void CssBuilder_WithComplexUserClasses()
{
    // Assert
    var cssBuilder = new CssBuilder("   my-user-class1     my-user-class2  ");

    // Act
    cssBuilder.AddClass("  class1  ");
    cssBuilder.AddClass("  class2  ");

    // Assert - Values are sorted
    Assert.Equal("class1 class2 my-user-class1 my-user-class2", cssBuilder.Build());
}
```

dvoituron (changes requested):
Good work. I like it :-)

tests/Core/Utilities/StyleBuilderTests.cs
StevenTCramer: That was a CssBuilder test in the `StyleBuilderTests.cs` file.  And we have similar tests

CssBuilder_AddMultipleValidUserClasses
CssBuilder_AddClassesWithExtraSpaces

In the appropriate CssBuilderTests.cs file.

But I did see I was missing a few  test cases.  I added.

CssBuilder_AddMultipleClassesAtOnce
CssBuilder_AddMultipleClassesAtOnceWithUserClass
CssBuilder_AddMultipleClassesAtOnceWithExtraSpaces

azunyuuuuuuu:
This commit sadly breaks using Tailwind CSS with this library. Classes like `hover:bg-red-400`, `grid-cols-[1fr_3fr]` or `[&>a]:text-sm` get eaten up by the regex.

Is there a way to disable or change the behaviour of the validation?

vnbaaij:
No, currently there is no way to prevent that from happening.

dvoituron:
> This commit sadly breaks using Tailwind CSS with this library. Classes like `hover:bg-red-400`, `grid-cols-[1fr_3fr]` or `[&>a]:text-sm` get eaten up by the regex.

Weird. These strings are not valid class names.

> a name must begin with an underscore (_), a hyphen (-), or a letter(a–z), followed by any number of hyphens, underscores, letters, or numbers.
https://stackoverflow.com/questions/448981/which-characters-are-valid-in-css-class-names-selectors

PR #1489, opened 2024-02-09
Adds FieldIdentifier parameter to FluentValidationMessage

thebarrettlo:
# Pull Request

## 📖 Description

Adds a new parameter, `Field` (of type `FieldIdentifier`) to the `FluentValidationMessage` component. This allows validation messages to be specified with a direct reference to a field within an `EditContext`/`EditForm`, rather than only through an accessor.

For fields that represent a property on a pre-defined model passed to an `EditContext`, this parameter is extraneous since those model (auto-)properties double as a field accessor. However, for dynamic models that do not have design-time accessors, a `FieldIdentifier` is more reliable than an accessor.

### Example usage

With some additional work, an `EditForm` can work with a dictionary rather than an object model:

```razor
<EditForm EditContext=@_editContext>
    @foreach (KeyValuePair<string, string?> kvp in _dynamicModel)
    {
        <FluentTextField Id=@kvp.Key
                         Label=@kvp.Key
                         Placeholder="Enter a string"
                         Value=@(_dynamicModel[kvp.Key])
                         ValueChanged=@(value => OnValueChanged(kvp.Key, value)) />
        <FluentValidationMessage Field=@(_editContext.Field(kvp.Key)) TValue=FieldIdentifier />
        @* <FluentValidationMessage For@(() => _dynamicModel[kvp.Key]) /> or something similar will not work *@
    }
</EditForm>

@code {
    private EditContext _editContext = new(new object());
    private Dictionary<string, string?> _dynamicModel = new();
    private ValidationMessageStore? _validationMessageStore;

    protected override void OnInitialized()
    {
        // Example model with a dictionary
        _dynamicModel = new()
        {
            { "Key 0", null },
            { "Key 1", null }
        };
        _editContext = new EditContext(_dynamicModel);
        _validationMessageStore = new ValidationMessageStore(_editContext);
    }

    private void OnValueChanged(string key, string value)
    {
        _dynamicModel[key] = value;

        // Validate the field here for example
        FieldIdentifier fieldIdentifier = _editContext.Field(key);
        _validationMessageStore?.Clear(fieldIdentifier);

        if (string.IsNullOrWhiteSpace(value))
        {
            _validationMessageStore?.Add(fieldIdentifier, "The field is required.");
        }
    }
}
```

No input given:

> image

Valid input and invalid (white space) input:

> image
> (I don't have a fix for the "valid" border color showing even though the field is invalid. Suggestions very welcome 😃)

### 🎫 Issues

- #1465

## 📑 Test Plan

[...]

PR #2114, opened 2024-05-27
Adds FieldIdentifier parameter to FluentInputBase

chris-gander:
# Pull Request

## 📖 Description

This PR adds a new parameter ```Field``` (of type ```FieldIdentifier```) to ```FluentInputBase```. I'm working on a system that allows users to create their own dynamic forms, this means my model contains a list of fields and each of those has a base class where the value is stored as an ```object?``` and inheriting classes override this with a specific type. When passing to input components the ```Value``` parameter gets cast to the correct type, but the ```FluentValidationMessage``` component can only be passed the ```object?``` version due to ```FieldIdentifiers``` not supporting the ```UnaryExpression``` that is created by casting. This means that the ```FluentInputBase``` has a different internal ```FieldIdentifier``` due to the unboxing, and means the validation message shows correctly when the value is invalid but the field stays showing as valid. By allowing the ```FieldIdentifier``` to be manually set this limitation can be overcome.

In normal statically typed models this additional parameter is not needed as everything will already work. However for dynamic use cases like mine, where design-time accessors are not possible, a ```FieldIdentifier``` works best. ```ValueExpression``` will also not work in this case as it also requires statically typed accessors. 

#1489 added a similar feature for ```FluentValidationMessage``` so I largely used the same code.

### Example

This is a basic example of how this works, it uses the FluentValidation library to simplify validating the dynamic model. This example is simpler than my use case but demonstrates the same issue:

```razor
<EditForm Model="FormState">
    <FluentValidationValidator />
    @foreach(var field in FormState.Fields)
    {
        @if(field.FieldType == "int")
        {
            <FluentNumberField Id="@field.FieldName"
                TValue="int?"
                Label="@field.FieldName" 
                Required="true"
                Placeholder="Enter a number"
                Value="(int?)field.Value"
                Field="@(FieldIdentifier.Create(() => field.Value))"
                ValueChanged="@(value => OnValueChanged(field, value))" />
        } else if (field.FieldType == "string")
        {
            <FluentTextField Id="@field.FieldName"
                Label="@field.FieldName"
                Required="true"
                Placeholder="Enter a string"
                Value="@((string?)field.Value)"
                Field="@(FieldIdentifier.Create(() => field.Value))"
                ValueChanged="@(value => OnValueChanged(field, value))" />
        }
        <FluentValidationMessage For="@(() => field.Value)" />
    }
</EditForm>

@code {
    private MyFormState FormState = new()
    {
        Fields = new()
        {
            new() {FieldType = "string", FieldName = "StringField", Value = ""},
            new() {FieldType = "int", FieldName = "IntField", Value = 1}
        }
    };

    private void OnValueChanged(MyFormField field, object? value)
    {
        field.Value = value;
    }

    public class MyFormState
    {
        public List<MyFormField> Fields { get; set; } = [];
    }

    public class MyFormField
    {
        public string FieldType { get; set; } = "string";
        public required string FieldName { get; set; }
        public object? Value { get; set; }
    }

    public class MyFormStateValidator : AbstractValidator<MyFormState>
    {
        public MyFormStateValidator()
        {
            RuleForEach(x => x.Fields).SetValidator(new MyFormFieldValidator());
        }
    }

    public class MyFormFieldValidator : AbstractValidator<MyFormField>
    {
        public MyFormFieldValidator()
        {
            RuleFor(x => (int?)x.Value).GreaterThan(2).WithName(x => x.FieldName).When(x => x.FieldType == "int");
            RuleFor(x => (string?)x.Value).NotEmpty().WithName(x => x.FieldName).When(x => x.FieldType == "string");
        }
    }
}
```

Without using ```Field``` and invalid the fields are still marked as valid
image

When using ```Field``` and invalid the fields are now correctly marked as invalid
image

### 🎫 Issues

No issues currently reported, but for this niche use case it is required

## 📑 Test Plan

No tests affected by this extra parameter, but all existing tests for ```FluentInputBase``` components pass. I have tested this code by creating custom versions of the number and text inputs in my project and verifying it works in both the above example and a more complex scenario, as well as making sure existing forms that don't use this new parameter still work.

### General

### Component-specific

### DIFF
--- a/src/Core/Components/Base/FluentInputBase.cs
+++ b/src/Core/Components/Base/FluentInputBase.cs
@@ -96,6 +96,13 @@ public abstract partial class FluentInputBase<TValue> : FluentComponentBase, IDi
     public Expression<Func<TValue>>? ValueExpression { get; set; }
 
     /// <summary>
+    /// Gets or sets the <see cref="FieldIdentifier"/> that identifies the bound value.
+    /// If set, this parameter takes precedence over <see cref="ValueExpression"/>.
+    /// </summary>
+    [Parameter]
+    public FieldIdentifier? Field { get; set; }
+
+    /// <summary>
     /// Gets or sets the display name for this field.
     /// <para>This value is used when generating error messages when the input value fails to parse correctly.</para>
     /// </summary>
@@ -115,7 +122,7 @@ public abstract partial class FluentInputBase<TValue> : FluentComponentBase, IDi
     public virtual string? Placeholder { get; set; }
 
     /// <summary>
-    /// Gets or sets if the derived component is embedded in another component. 
+    /// Gets or sets if the derived component is embedded in another component.
     /// If true, the ClassValue property will not include the EditContext's FieldCssClass.
     /// </summary>
     [Parameter]
@@ -132,7 +139,7 @@ public abstract partial class FluentInputBase<TValue> : FluentComponentBase, IDi
     /// </summary>
     protected internal FieldIdentifier FieldIdentifier { get; set; }
 
-    internal bool FieldBound => ValueExpression != null || ValueChanged.HasDelegate;
+    internal bool FieldBound => Field is not null || ValueExpression != null || ValueChanged.HasDelegate;
 
     protected async Task SetCurrentValueAsync(TValue? value)
     {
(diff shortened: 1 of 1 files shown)


chris-gander:

dvoituron (approved):

PR #2118, opened 2024-05-28
[Lists] Make list components inherit from InputBase

### LINKED ISSUE #2086: fix: Updating EditContext state in FluentSelect
# 🐛 Bug Report

`FluentSelect` never marks the bound field as modified in the edit context.

## 💻 Repro or Code Sample

```razor

<EditForm Model="@this">
    <p>Modified: @(context.IsModified())</p>

    <FluentSelect TOption="MyThing" @bind-SelectedOption="@currentThing"
        Items="@AllThings" OptionText="@(x => x.DisplayText)" />

    <FluentTextField @bind-Value="@someString" />
</EditForm>

@code {
    MyThing? currentThing;
    string? someString;
    MyThing[] AllThings = new[] { new MyThing(1, "One"), new MyThing(2, "Two") };

    record MyThing(int Id, string DisplayText);
}
```

## 🤔 Expected Behavior

If you change the `FluentSelect` dropdown, then "Modified" should become true. Notice that this *does* happen correctly if you change the text in the `FluentTextField` box.

## 😯 Current Behavior

`FluentSelect` never tells the editcontext that the bound value was modified, so "Modified" remains false.

This breaks many scenarios around showing "save" only if the form is edited, or warning the user not to exit if the form has unsaved changes.

## 💁 Possible Solution

As a (rather unpleasant) workaround, the developer could add:

```razor
@bind-SelectedOption:after="@(() => { context.NotifyFieldChanged(FieldIdentifier.Create(() => currentThing)); })"
```

However this should not be necessary as form components should notify the containing EditContext when they mutate the bound field.

vnbaaij:
Fix #2086 by making `FluentAutoComplete`, `FluentCombobox`, `FluentListbox` and `FluentSelect` inherit from `InputBase` instead of `FluentComponentBase`.
This allows for the components to partake in form validation and be marked as invalid.

WIP. We will probably do separate PRs per list component to not make them too big for review


dvoituron:
After quickly tested the samples:
1. Autocomplete no longer works when the user try to select an item using the mouse.
2. Combobox no longer works when the user try to select an item using the keyboard.

peek

src/Core/Components/Base/FluentInputBase.cs
```diff
@@ -139,7 +143,7 @@ public abstract partial class FluentInputBase<TValue> : FluentComponentBase, IDi
     /// </summary>
     protected internal FieldIdentifier FieldIdentifier { get; set; }
 
-    internal bool FieldBound => Field is not null || ValueExpression != null || ValueChanged.HasDelegate;
+    internal bool FieldBound { get; set; }
```
dvoituron, line 146: Why did you change this, why didn't you keep `bool FieldBound => Field is not null || ValueExpression != null || ValueChanged.HasDelegate;` ?

I find the property clearer if it retains its verification criteria.


I found this line in `ListComponentBase` class.
```csharp
FieldBound = ValueChanged.HasDelegate || SelectedOptionChanged.HasDelegate || SelectedOptionsChanged.HasDelegate;`
```

It might be preferable to adapt the `bool FieldBound` to a virtual property and override it in `ListComponentBase`.

dvoituron (changes requested):

src/Core/Components/Base/FluentInputBase.cs
vnbaaij: Done with override now

vnbaaij:
> After quickly tested the samples:
> 
> 1. Autocomplete no longer works when the user try to select an item using the mouse.
> 2. Combobox no longer works when the user try to select an item using the keyboard.
> 
1. Haven't found what change caused this (yet)
2. is actually a different issue (and is also not working correctly today in the current version). I've made some changes and you can now select an option with 'enter' again.

dvoituron:
`Combobox.SelectedOption` is not still working using keyboard

image

src/Core.Assets/src/index.ts
```diff
@@ -50,6 +50,17 @@ body:has(.prevent-scroll) {
     --presence-unknown: #d13438;
     --highlight-bg: #fff3cd;
 }
+
+
+[role='checkbox'].invalid::part(control),
```
dvoituron, line 55: Why did you added this code? This has nothing to do with `InputBase` inheritance.

src/Core/Components/Base/FluentInputBase.cs
```diff
@@ -303,17 +307,20 @@ public override Task SetParametersAsync(ParameterView parameters)
             // This is the first run
             // Could put this logic in OnInit, but its nice to avoid forcing people who override OnInit to call base.OnInit()
 
-            if(Field is not null)
+            if (Field is not null)
             {
                 FieldIdentifier = (FieldIdentifier)Field;
+                //FieldBound = true;
```
dvoituron, line 313: Could you remove these commented lines?

src/Core/Components/List/ListComponentBase.razor.cs
```diff
@@ -322,6 +293,8 @@ protected override void OnInitialized()
         {
             Id = Identifier.NewId();
         }
+
+        //FieldBound = ValueChanged.HasDelegate || SelectedOptionChanged.HasDelegate || SelectedOptionsChanged.HasDelegate;
```
dvoituron, line 297: Same here, could you remove this commented line?

src/Core/Components/List/ListComponentBase.razor.cs
```diff
@@ -369,6 +342,26 @@ protected override void OnParametersSet()
...
+    protected override bool TryParseValueFromString(string? value, [MaybeNullWhen(false)] out string? result, [NotNullWhen(false)] out string? validationErrorMessage)
+        => this.TryParseSelectableValueFromString(value, out result, out validationErrorMessage);
+
+    /// <inheritdoc />
+    protected override string? FormatValueAsString(string? value)
+    {
+        // We special-case bool values because BindConverter reserves bool conversion for conditional attributes.
+        if (typeof(TOption) == typeof(bool))
+        {
+            return (bool)(object)value! ? "true" : "false";
```
dvoituron, line 355: By definition, `value` can be null. Maybe you could add a test like this (to remove the `!`)
```csharp
if (value is not null && typeof(TOption) == typeof(bool))
```

dvoituron (changes requested):

src/Core.Assets/src/index.ts
vnbaaij: This is needed for validation of all inputBase components. Earlier source location was not correct and code was not complete

vnbaaij:
> `Combobox.SelectedOption` is not still working using keyboard

As discussed, this is not related to thee changes. Will be resolved outside of this PR

dvoituron (approved):

PR #2131, opened 2024-05-31
[Package versions] Refactoring

dvoituron:
# [Package versions] Refactoring

- Set package versions to stable versions. Except for `Microsoft.CodeAnalysis.Analyzers` where the latest stable version is 3 years old.
- Consider all Warnings like Errors, during the build process: `-warnaserror`
- Update C# code to respect the Analyzer (missing `Async`).


vnbaaij (approved):

vnbaaij:
The latest coverlet and Markdig are not available in dotnet-public feed.

dvoituron:
Why do we need to use this feed and not NuGet.org?

vnbaaij:
Requirements of the 1ES templates/security initiative.

dvoituron:
Ok. I will update that.

Issue #2148, opened 2024-06-04
DataGrid: add access to columns


ksbecker:
# 🙋 Feature Request

Give developers access to the collection of columns in a DataGrid.

## 🤔 Expected Behavior

In order to programmatically set the sort, remove sort, or show column options, you need to pass in a `ColumnBase<TGridItem>` into one of the methods, `RemoveSortByColumnAsync`, `ShowColumnOptionsAsync`, and `SortByColumnAsync`. As far as I can tell, there is no way to get a reference to the columns.

## 😯 Current Behavior

Currently there is way to access this collection.

## 💁 Possible Solution

There is a [private field](https://github.com/microsoft/fluentui-blazor/blob/5a65947e464d4e2f48411da65b8eb651606d2943/src/Core/Components/DataGrid/FluentDataGrid.razor.cs#L191) that contains a list of `ColumnBase<TGridItem>` that could be exposed through a getter property.

## 🔦 Context

What I'm trying to accomplish is to store the state of the DataGrid and restore it when navigating back to the page. I have filters and paging working, now I need to be able to restore the previous sort order. I can accomplish this by storing an index/unique id of the column that was sorted and then applying the `SortByColumnAsync` method.

## 💻 Examples

I think this is fairly straight-forward. Basically add a new property in `fluentui-blazor/src/Core/Components/DataGrid/FluentDataGrid.razor.cs`

`public List<ColumnBase<TGridItem>> Columns { get => _columns; }`

Alternatively, methods could be created to give access to these columns, `GetColumnByTitle` or `GetColumnByIndex`

vnbaaij:
By exposing a getter, you would have _read only_ access to the _collection_ of columns. But then nothing would stop you from manipulating the _individuial_ columns from that collection, right?

I see more value in providing the other type of methods you mention. 

As we only have limited time for working on this, care to work on a PR?

PR #2156, opened 2024-06-05
[DataGrid] Allow programmatic sorting of columns by name or index

ksbecker:
# Pull Request

## 📖 Description

The publicly exposed method of programmatically setting a DataGrid's sort order requires the user to have a reference to the `ColumnBase<TGridItem>` which the API does not expose. In order to restore the state of a DataGrid, the sort order needs to be settable.

### 🎫 Issues

[#2148](https://github.com/microsoft/fluentui-blazor/issues/2148) - DataGrid: add access to columns

## 👩‍💻 Reviewer Notes

I added 2 logical methods that one would sort by. I did find another `internal` method that could have been used but decided to just add these. I think it would be OK to mark `SortByColumnAsync`, 'RemoveSortByColumnAsync', and `ShowColumnOptionsAsync` as `internal` as it is necessary to have a `ColumnBase<TGridItem>` reference in order to execute those methods, which is something that's not exposed outside of the library.

## 📑 Test Plan

I added new tests for this function based on other existing tests. I used methodology from other tests. I did have to use the `cut.InvokeAsync` method as there is a `StateHasChanged()` call within the component.

### General

I did some spellchecking on the DataGrid page.

### Component-specific

- [X] I have validated the [Unit Tests](https://github.com/Microsoft/fluentui-blazor/blob/master/unit-tests.md) for an existing component

### DIFF
--- a/src/Core/Components/DataGrid/FluentDataGrid.razor.cs
+++ b/src/Core/Components/DataGrid/FluentDataGrid.razor.cs
@@ -374,7 +374,7 @@ public partial class FluentDataGrid<TGridItem> : FluentComponentBase, IHandleEve
     /// </summary>
     /// <param name="title">The title of the column to sort by.</param>
     /// <param name="direction">The direction of sorting. The default is <see cref="SortDirection.Auto"/>. If the value is <see cref="SortDirection.Auto"/>, then it will toggle the direction on each call.</param>
-    public Task SortByColumnTitleAsync(string title, SortDirection direction = SortDirection.Auto)
+    public Task SortByColumnAsync(string title, SortDirection direction = SortDirection.Auto)
     {
         var column = _columns.FirstOrDefault(c => c.Title?.Equals(title, StringComparison.InvariantCultureIgnoreCase) ?? false);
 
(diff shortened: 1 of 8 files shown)


ksbecker:

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -369,6 +369,38 @@ public Task SortByColumnAsync(ColumnBase<TGridItem> column, SortDirection direct
         return RefreshDataAsync();
     }
 
+    /// <summary>
+    /// Sorts the grid by the specified column <paramref name="title"/> found first. If the title is not found, nothing happens.
+    /// </summary>
+    /// <param name="title">The title of the column to sort by.</param>
+    /// <param name="direction">The direction of sorting. The default is <see cref="SortDirection.Auto"/>. If the value is <see cref="SortDirection.Auto"/>, then it will toggle the direction on each call.</param>
+    public Task SortByColumnTitleAsync(string title, SortDirection direction = SortDirection.Auto)
```
dvoituron, line 377: Might it be better to have the same method name, but with these two signatures?

```csharp
public Task SortColumnAsync(string title, SortDirection direction = SortDirection.Auto)
public Task SortColumnAsync(int index, SortDirection direction = SortDirection.Auto)
```

So, later we could add another way to identify the column.

dvoituron (changes requested):

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
ksbecker: Just a personal preference, but I went ahead and changed it to overload methods. Tests have been updated as well.

dvoituron (approved):

vnbaaij (approved):

Issue #2130, opened 2024-05-31
feat: [DataGrid] Deselect all other items when row is clicked (not checkbox)


hksalessio:
# 🙋 Feature Request

Sorry for the title, I didn't really know how to express myself better.

I would like the `FluentDataGrid` to have an option for deselecting all items when a row is clicked and only keeping this row as selected. This is not the same as single select mode. The following video might be better at explaining things than I am:

https://github.com/microsoft/fluentui-blazor/assets/143792830/99c428cd-957d-434e-bf7b-7f274696a59f

The video was taken in the Microsoft admin center, so I guess if this is something Microsoft is using, it might be worth adding to the Library :D

hksalessio:
Maybe I could get around a new feature for this if you provided a "OnCellClick" event. I mean, "OnCellFocus" exists as well, so why not add it for the sake of completeness.

vnbaaij:
We are not going to implement that (for now). It is difficult to do because the web component does not have a click event on the cell

hksalessio:
But wasn't the "OnRowClick" functionality implemented without the web component's click event as well? I could provide a feature request but only if you would be willing to add it to the library

vnbaaij:
Yes, that is true. But implementing that for all cells as well will lead to performance challenges, I think.

If you want to work on this and it turns out performance stays on par then we will want to merge it.

PR #2160, opened 2024-06-06
[DataGrid] Add OnCellClick and allow restricting selection to checkboxes

hksalessio:
# Pull Request

## 📖 Description

This PR adds an `OnCellClick` functionality to `FluentDataGrid` and allows restricting the selection of rows to the checkboxes of the select column.

### 🎫 Issues

This PR should allow for implementing the features requested in #2130 and https://github.com/microsoft/fluentui-blazor/pull/1952#issuecomment-2150964551.

## 👩‍💻 Reviewer Notes

~Since I renamed `FluentDataGridCell.GridColumn` to `FluentDataGridCell.ColumnIndex` (because I think this is more fitting), this PR technically contains a breaking change. If you don't want this change, I will revert it.~

### General

### Component-specific


src/Core/Components/DataGrid/FluentDataGridCell.razor.cs
```diff
@@ -24,12 +24,18 @@ public partial class FluentDataGridCell<TGridItem> : FluentComponentBase
...
+    [Parameter]
+    public ColumnBase<TGridItem>? Column { get; set; }
+
     /// <summary>
     /// Gets or sets the column index of the cell.
     /// This will be applied to the css grid-column-index value applied to the cell.
     /// </summary>
     [Parameter]
-    public int GridColumn { get; set; }
+    public int ColumnIndex { get; set; }
```
dvoituron, line 38: You cannot change the name of an existing property. This modification is a breaking change.

vnbaaij:
Yeah, you definitely need to revert the renaming first. We use that name because it is also used in the web component and we want to keep that (logical) link

vnbaaij (changes requested):
Revert `GridColumn` name change

dvoituron (changes requested):

src/Core/Components/DataGrid/FluentDataGridCell.razor.cs
hksalessio: Done

src/Core/Components/DataGrid/FluentDataGridCell.razor.cs
```diff
@@ -24,12 +24,18 @@ public partial class FluentDataGridCell<TGridItem> : FluentComponentBase
     [Parameter]
     public DataGridCellType? CellType { get; set; } = DataGridCellType.Default;
 
+    /// <summary>
+    /// Gets or sets the column of the cell.
+    /// </summary>
+    [Parameter]
+    public ColumnBase<TGridItem>? Column { get; set; }
```
dvoituron, line 31: I'm not sure that adding this property is a good idea.

I'm not sure adding this property is a good idea. Placing an additional Column reference for all the cells in the **DataGrid**, which can contain  several tens of thousands of cells, could lead to performance problems.

I did not verify but you could use the column index to retrieve the Column: updating the `FluentDataGridRow.Owner` property to `internal` and you will have an access to the Grid and the list of columns: `Owner.Owner.Grid...`.

src/Core/Components/DataGrid/FluentDataGridCell.razor.cs
```diff
@@ -61,4 +67,23 @@ protected override void OnInitialized()
 
     public void Dispose() => Owner.Unregister(this);
 
+    /// <summary />
+    internal async Task HandleOnCellClickAsync(string cellId)
```
dvoituron, line 71: You could change that to `internal async Task HandleOnCellClickAsync(FluentDataGridCell<TGridItem> cell)` and calling this method via `@onclick="@(e => HandleOnCellClickAsync(this))"`

src/Core/Components/DataGrid/FluentDataGridCell.razor.cs
```diff
@@ -61,4 +67,23 @@ protected override void OnInitialized()
...
+            if (GridContext.Grid.OnCellClick.HasDelegate)
+            {
+                await GridContext.Grid.OnCellClick.InvokeAsync(cell);
+            }
+
+            if (cell != null && cell.CellType == DataGridCellType.Default)
+            {
+                if (Column is SelectColumn<TGridItem> selColumn)
+                {
+                    await selColumn.AddOrRemoveSelectedItemAsync(Item);
```
dvoituron, line 84: Won't this be a duplicate with the same call in `HandleOnRowClickAsync` (if `RestrictToCheckbox=false`? And generate double selections at the same time?

src/Core/Components/DataGrid/FluentDataGridRow.razor.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.FluentUI.AspNetCore.Components;
 public partial class FluentDataGridRow<TGridItem> : FluentComponentBase, IHandleEvent, IDisposable
 {
     internal string RowId { get; set; } = string.Empty;
-    private readonly Dictionary<string, FluentDataGridCell<TGridItem>> cells = [];
+    internal Dictionary<string, FluentDataGridCell<TGridItem>> Cells { get; set; } = [];
```
dvoituron, line 16: Why did you remove the `readonly`

dvoituron (changes requested):

dvoituron:
Could you add some unit tests to `FluentDataGridColumSelectTests` to validate the new `RestrictToCheckbox` attribute?

src/Core/Components/DataGrid/FluentDataGridCell.razor.cs
hksalessio: That's what the `@onclick:stopPropagation` in `FluentDataGridCell` is used for. Would probably be cleaner to just add a condition, right?

src/Core/Components/DataGrid/FluentDataGridRow.razor.cs
hksalessio: Had to remove it because of the getter and setter. Copy and pasted a lot of code from `FluentDataGridRow` but I will simply remove the getter and setter for the cell.

dvoituron: Make sense. But in this case, the `setter` can be removed (to have a "readonly" property).

src/Core/Components/DataGrid/FluentDataGridCell.razor.cs
hksalessio: That's actually pretty smart! Just one question: Is there a reason you didn't do this for OnRowClick as well? Or should I change that method, too, while I'm at it?

dvoituron: The reason for this is probably that we've learned during this project :-)
Don't use this PR. It's better to have small PRs that add small features.

src/Core/Components/DataGrid/FluentDataGridCell.razor.cs
hksalessio: Was a bit more complicated than that but I followed this idea :D

hksalessio:
> Could you add some unit tests to `FluentDataGridColumSelectTests` to validate the new `RestrictToCheckbox` attribute?

I added some 👍

src/Core/Components/DataGrid/FluentDataGridCell.razor.cs
```diff
@@ -61,4 +61,29 @@ protected override void OnInitialized()
 
     public void Dispose() => Owner.Unregister(this);
 
+    private bool StopClickPropagation()
+    {
+        var selColumn = Owner.Owner.Grid.SelectColumns.FirstOrDefault();
```
dvoituron, line 66: You need to manage all `SelectColumns`, not just the first one. It is possible to have several `SelectColumns` in the DataGrid.

examples/Demo/Shared/Pages/DataGrid/Examples/DataGridMultiSelect.razor
```diff
@@ -4,16 +4,20 @@
...
+                    Label="Use `RestrictToCheckbox` property" />
 </FluentStack>
 
 @if (UseSelectedItems)
 {
     @* Sample using SelectedItems  *@
     <div>Using SelectedItems</div>
 
-    <FluentDataGrid Items="@People" ShowHover="true" TGridItem="Person">
+    <FluentDataGrid Items="@People" ShowHover="true" TGridItem="Person" OnRowClick="@(row => Console.WriteLine($"Row {row.RowIndex} clicked"))" OnCellClick="@(cell => Console.WriteLine($"Cell {cell.GridColumn} clicked"))">
```
dvoituron, line 17: You must also add these two events to the second `FluentDataGrid`. If not, no results will be displayed in the demo example.

dvoituron: And could you align these attributes to avoid too long lines?

```xml
<FluentDataGrid Items="@People" ShowHover="true" TGridItem="Person"
                OnRowClick="@(row => Console.WriteLine($"Row {row.RowIndex} clicked"))"
                OnCellClick="@(cell => Console.WriteLine($"Cell {cell.GridColumn} clicked"))">
```

dvoituron (changes requested):

dvoituron:
An UI issue occurs when I change the **RestrictToCheckbox** property

peek

dvoituron:
The OnRowClick should be raised in all cases: it's correct when `RestrictToCheckbox` is unchecked.
But is not correct when `RestrictToCheckbox` is checked.

peek

hksalessio:
> The OnRowClick should be raised in all cases: it's correct when `RestrictToCheckbox` is unchecked. But is not correct when `RestrictToCheckbox` is checked.
> 
> peek [ peek ](https://private-user-images.githubusercontent.com/8350694/337575694-955a96a0-1b91-4af7-83af-6cc949fbdc2e.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTc3NDc2NzgsIm5iZiI6MTcxNzc0NzM3OCwicGF0aCI6Ii84MzUwNjk0LzMzNzU3NTY5NC05NTVhOTZhMC0xYjkxLTRhZjctODNhZi02Y2M5NDlmYmRjMmUuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDYwNyUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA2MDdUMDgwMjU4WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9OWMxNTAwODM2YzRhMzRlNmRhNTgyYWM4Mzc3YmE4NDlhZTI3ZTU5OTg4M2VjZDRlZWQyOTBhNWIwYzI2NzU5ZiZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.HzCjR1XPrx-gV_BrD--2GAhHx1A7FjT1YQ4hVWV1Abo) [ ](https://private-user-images.githubusercontent.com/8350694/337575694-955a96a0-1b91-4af7-83af-6cc949fbdc2e.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTc3NDc2NzgsIm5iZiI6MTcxNzc0NzM3OCwicGF0aCI6Ii84MzUwNjk0LzMzNzU3NTY5NC05NTVhOTZhMC0xYjkxLTRhZjctODNhZi02Y2M5NDlmYmRjMmUuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDYwNyUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA2MDdUMDgwMjU4WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9OWMxNTAwODM2YzRhMzRlNmRhNTgyYWM4Mzc3YmE4NDlhZTI3ZTU5OTg4M2VjZDRlZWQyOTBhNWIwYzI2NzU5ZiZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.HzCjR1XPrx-gV_BrD--2GAhHx1A7FjT1YQ4hVWV1Abo)

The problem with this is that I would not see any way to implement #2130 since the row click is always triggered after the cell click. When clicking the checkbox, the clicked row would be selected but if `OnRowClick` is executed afterwards to deselect all other rows, the previous selection would be cleared (of course, this only applies to multi select mode).

dvoituron:
> If `OnRowClick` is executed afterwards to deselect all other rows, the previous selection would be cleared (of course, this only applies to multi select mode).

Yes, that was my initial question. I haven't analyzed the problem. But it would be better to have this event in any case. If not, we'll get some "feedback" :-) I can look at it in a few days if you like, see if I have another idea.

hksalessio:
Maybe we could add a new parameter `StopClickPropagation` or `PreventRowClick` to columns?

dvoituron:
> Maybe we could add a new parameter `StopClickPropagation` or `PreventRowClick` to columns?

No. That's not ideal. We should find a solution to do that automatically.

hksalessio:
> I can look at it in a few days if you like, see if I have another idea.

Hi, any news on this? 😅

dvoituron:
> Hi, any news on this?

I haven't had time yet :-(

franklupo:
Normally you select only from the checkbox. the current behavior is not correct. If you want something different you use the envents on the row.

franklupo:
for #2130 the solution is in the event. Call deselect all and select the current row.

hksalessio:
> Normally you select only from the checkbox. the current behavior is not correct. If you want something different you use the envents on the row.

I definitely agree with you. However, this could be considered a breaking change by the repository owners.

franklupo:
> > Normally you select only from the checkbox. the current behavior is not correct. If you want something different you use the envents on the row.
> 
> I definitely agree with you. However, this could be considered a breaking change by the repository owners.

the solution might be to create a new selection column that behaves correctly.

dvoituron:
@hksalessio @vnbaaij I created this new PR: #2252 
Give me your feedbacks

vnbaaij:
Other PR LGTM. Merged it and think we can close this one.

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -175,7 +181,7 @@ public partial class FluentDataGrid<TGridItem> : FluentComponentBase, IHandleEve
     /// <summary>
     /// Gets the first (optional) SelectColumn
     /// </summary>
-    internal IEnumerable<SelectColumn<TGridItem>> SelectColumns => _columns.Where(col => col is SelectColumn<TGridItem>).Cast< SelectColumn<TGridItem>>();
+    internal IEnumerable<SelectColumn<TGridItem>> SelectColumns => Columns.Where(col => col is SelectColumn<TGridItem>).Cast< SelectColumn<TGridItem>>();
```
franklupo, line 184: why accept multiple selection columns? 
I will raise an exception,

hksalessio:
> ⚠️ @hksalessio @vnbaaij I created this new PR: #2252 Give me your feedbacks

Looks good to me, thank you!

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
franklupo: never seen other grids with multiple selection columns

vnbaaij: It will not be _the_ most used function, but let's not rule it out. Think about maybe switching out a multi select column for a single select one, showing and hiding one or the other based on some condition or user interaction.

PR #2176, opened 2024-06-10
[FluentCalendar] Select multiple dates and range of dates

dvoituron:
# [FluentCalendar] Select multiple dates and range of dates

We're thinking of adding a `SelectMode` property so that the user can select a range of dates `CalendarSelectMode.Range` or a discontinuous list of dates `CalendarSelectMode.Multiple`.

Initially, this feature will only be available for the FluentCalendar with `View=Days`.

A new bindable property `SelectedDates` will retrieve these dates, rendering the existing `Value` property useless in these `SelectModes` only.

Added a `SelectDatesHover` method that returns a list of dates to be pre-selected (hover) and activated when the user clicks on this selection. This allows you to manage choices such as selecting an entire week, with a single mouse click. Or selecting 3 days automatically, with a single click.

The `@bind-SelectedDates:after` command can be used to validate the content of selected dates, block or remove them if necessary, or even display an error message.

## Example

```razor
<FluentCalendar @bind-SelectedDates="@SelectedRangeCustomized"
                SelectDatesHover="@SelectOneWeek"
                SelectMode="CalendarSelectMode.Range" />

<FluentCalendar @bind-SelectedDates="@SelectedDaysCustomized"
                SelectDatesHover="@Select3Days"
                SelectMode="CalendarSelectMode.Multiple" />

@code
{
    private ICollection<DateTime> SelectedRangeCustomized = new List<DateTime>();
    private ICollection<DateTime> SelectedDaysCustomized = new List<DateTime>();

    private IEnumerable<DateTime> SelectOneWeek(DateTime date)
    {
        return new[] { date.StartOfWeek(), date.StartOfWeek().AddDays(6) };
    }

    private IEnumerable<DateTime> Select3Days(DateTime date)
    {
        return Enumerable.Range(0, 3).Select(i => date.AddDays(i));
    }
}
```

DateSelection

## TODO
- [x] Development
- [x] Validation
- [x] Unit Tests


GamerGambit:
This looks great. Does this work across months? For example in the provided demo's would I be able to select May 27 through to June 2?

dvoituron:
Yep. You can select days across months views, like in your example.

Seanxwy:
The data range should provide a setting to get the start and end times. Such a component is usually needed for time period queries

dvoituron:
> The data range should provide a setting to get the start and end times. Such a component is usually needed for time period queries

This PR is **only** to add a feature to select a range of Dates (from/to) or a list of multiple days (IEnumerable).
All other existing properties (disabled days, min/max days) are still available.

osmanm786:
I live it!. I will be using this as soon as this comes out.

vnbaaij (approved):

PR #2183, opened 2024-06-12
[NavMenu] Enhance working in mobile view

vnbaaij:
> When using a `FluentNavGroup` and tap on it in mobile view it closes the opened navigation. When you use the dropdown buttons instead the navigation will open just like expected without closing the mobile menu.

The demo site was already configured for this but it would not work that way in a site created with the template.  With these changes clicking on either the button or the group will just expand/collapse that group (when no href set on the group) for template created sites. 

The template created sites had an (JS) `onclick` event on the `nav` element so that the menu would be closed after navigating to an item. This PR moves that logic to the `FluentNavGroup`/`FluentNavLink` (when an href is set)

With these change we get consistent behavior in both demo and template sites whare a menu group can be expanded/collapsed by clicking it and clicking an item navigates to that item and closes the menu:

### Before:
better-menu-before

### After:
better-menu

Changes:
- Add CustomToggle parameter to FluentNavMenu
- Add CustomToggleId to FluentNavBase
- Add onclick to FluentNavGroup/FluentNavLink

### DIFF
--- a/src/Core/Components/NavMenu/FluentNavBase.cs
+++ b/src/Core/Components/NavMenu/FluentNavBase.cs
@@ -75,17 +75,6 @@ public abstract class FluentNavBase : FluentComponentBase
     [Parameter]
     public string? Tooltip { get; set; }
 
-    [CascadingParameter]
-    public FluentNavMenu Owner { get; set; } = default!;
-
-    [CascadingParameter]
-    public FluentMenu? SubMenu { get; set; }
-
-    /// <summary>
-    /// Returns <see langword="true"/> if the item has an <see cref="Icon"/> set.
-    /// </summary>
-    internal bool HasIcon => Icon is not null;
-
     /// <summary>
     /// The callback to invoke when the item is clicked.
     /// </summary>
@@ -98,8 +87,32 @@ public abstract class FluentNavBase : FluentComponentBase
     [Parameter]
     public bool ForceLoad { get; set; }
 
+    /// <summary>
+    /// Gets or sets the id of the custom toggle element
+    /// Defaults to navmenu-toggle
+    /// </summary>
+    [Parameter]
+    public string CustomToggleId { get; set; } = "navmenu-toggle";
+
+    [CascadingParameter]
+    public FluentNavMenu Owner { get; set; } = default!;
+
+    [CascadingParameter]
+    public FluentMenu? SubMenu { get; set; }
+
     [Inject]
     private NavigationManager NavigationManager { get; set; } = default!;
+
+    /// <summary>
+    /// Returns <see langword="true"/> if the item has an <see cref="Icon"/> set.
+    /// </summary>
+    internal bool HasIcon => Icon is not null;
+
+    /// <summary>
+    /// If a custom toggle is being used to hide/show the menu, this defines the 'onclick' code
+    /// Uses the <see cref="CustomToggleId"/> as the id of the elemet that will be clicked
+    /// </summary>
+    internal string? CustomToggleCode => Owner.CustomToggle ? $"document.getElementById('{CustomToggleId}').click();" : null;
 
     protected async Task OnClickHandlerAsync(MouseEventArgs ev)
     {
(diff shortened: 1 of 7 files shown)


src/Core/Components/NavMenu/FluentNavBase.cs
```diff
@@ -98,9 +87,33 @@ public abstract class FluentNavBase : FluentComponentBase
...
 
+    /// <summary>
+    /// Gets or sets the id of the custom toggle element
+    /// Defaults to navmenu-toggle
+    /// </summary>
+    [Parameter]
+    public string CustomToggleId { get; set; } = "navmenu-toggle";
+
+    [CascadingParameter]
+    public FluentNavMenu Owner { get; set; } = default!;
```
dvoituron, line 98: Could we set these 2 `CascadingParameter` to `internal` ?

src/Core/Components/NavMenu/FluentNavMenu.razor.cs
```diff
@@ -68,6 +68,9 @@ public partial class FluentNavMenu : FluentComponentBase
     [Parameter]
     public string? Margin { get; set; }
 
+    [Parameter]
+    public bool CustomToggle { get; set; } = false;
```
dvoituron, line 72: Could you add a documentation ?

src/Core/Components/NavMenu/FluentNavBase.cs
```diff
@@ -98,9 +87,33 @@ public abstract class FluentNavBase : FluentComponentBase
     [Parameter]
     public bool ForceLoad { get; set; }
 
+    /// <summary>
+    /// Gets or sets the id of the custom toggle element
+    /// Defaults to navmenu-toggle
+    /// </summary>
+    [Parameter]
+    public string CustomToggleId { get; set; } = "navmenu-toggle";
```
dvoituron, line 95: 1. How to be sure that multiple Navmenu will have a unique `CustomToggleId`?
1. Why not to use the `Id` + `toggle` string to have a unique ID?

dvoituron (changes requested):

src/Core/Components/NavMenu/FluentNavBase.cs
vnbaaij: The 'happy path' is that there will be 1 `FluentNavMenu` in a site. In that case you don't need to set the id and the defaults will 'just work'.
In case someone wants to use multiple nav menu's in one site **and** also wants to make them all toggle-able, they will need to supply these id's and make sure they are unique,

src/Core/Components/NavMenu/FluentNavBase.cs
vnbaaij: Making them internal leas to errors in tests. Blazor also expects `[Parameter]` and `[CascadingParameter]` properties to be public

dvoituron: I think `CascadingParameter` can be private or internal (ex in `FluentDataGridCell`).
And to access internal properties from unit tests, we've added
`[assembly: System.Runtime.CompilerServices.InternalsVisibleTo("Microsoft.FluentUI.AspNetCore.Components.Tests")]` 

Which error do you have in tests?

vnbaaij: When Owner is internal:  System.ArgumentException : The parameter selector 'p => p.Owner' does not resolve to a public property on the component 'Microsoft.FluentUI.AspNetCore.Components.FluentNavGroup' with a [Parameter] or [CascadingParameter]attribute. (Parameter 'parameterSelector') (in tests project)

Don't think it is worth the hassle to 'fight' this error and just leave them public as Blazor expects them to be

dvoituron (approved):

nhwilly:
This is working perfectly for me - thank you for the update.  I didn't think of the approach of making the menu take 100% but it makes great sense in operation.  

In my application, a `user` has their own set of functions, and I put them in a `navmenu`.  In mobile mode this appears in the top right header just like your approach.

But they also have a set of `account` functions and I want to put them in a `navmenu` in the opposite corner in a mobile app.  This would mean appearing on the left.

I can't seem to make that work.   I thought I could use the `CustomToggleId` somehow, but that seems to be assigned at the NavGroup level.

I think it is perfectly OK to close one menu when another one opens when the screen is less than 600px.  In normal mode, having two menus works great.

Are there any tips or samples to achieve this?  Would it help if I put up a repo?

vnbaaij:
This k you should start a post I the discussion area. Include some screens and code examples.
Having a repo with the 'problematic' code always helps.

Issue #2197, opened 2024-06-17
fix: Datetime not selected after remove in Demo site


dungntw:
# 🐛 Bug Report

Datetime not selected after remove

https://github.com/microsoft/fluentui-blazor/assets/119403879/54874cdf-8115-4a6e-a6c9-1957ec59ac5a

I also checked in
- Form -> Overview
- Components -> DateTime -> Default Date Picker

## 🌍 Your Environment

* OS & Device: Windows on PC
* Browser Microsoft Edge, Google Chrome, Mozilla FireFox
* .NET and Fluent UI Blazor library Version 4.8.0

SquaredLoops:
This **USED** to work and I think the bug was introduced as part of #2176 (certainly something in 4.8.0 release).

I know this because I have just spent the last couple of hours trying to figure out why my date picker that used to work has now stopped. My date picker is blank / empty on page load and a mandatory field, so was very obvious when it broke as the flow could no longer work.

When I looked in GitHub I saw this change, and this bug. When I looked at my VS project (that was created from the Fluent UI template) I noticed it references `4.*-*` so picked up this change. To fix it locally I forced to `4.7.2` so change your `csproj` files...

From:
```
<PackageReference Include="Microsoft.FluentUI.AspNetCore.Components" Version="4.*-* " />
```

To:
```
<PackageReference Include="Microsoft.FluentUI.AspNetCore.Components" Version="4.7.2 " />
```

I hope this can help others whose controls have suddenly started behaving strangely. I should have checked from the outset that my versions were pinned, so my fault, but a timely reminder for everyone to check! 😄

MarvinKlein1508:
@vnbaaij this has something todo with the `@bind` value being null. When you create a nullable model setting the date won't work from the start on.

https://github.com/MarvinKlein1508/FluentUI-DateIssue

PR #2233, opened 2024-06-20
[DatePicker] Change OnSelectedDateAsync logic

MarvinKlein1508:
# Pull Request

## 📖 Description

This fixes the issue that the `FluentDatePicker` does not work, when the value of `@bind` is null. The issue comes from the fact that  
`(value ?? DateTime.MinValue).Date + Value?.TimeOfDay` returns always null when `Value is null`

### 🎫 Issues

* #2197

### General


src/Core/Components/DateTime/FluentDatePicker.razor.cs
```diff
@@ -66,10 +66,15 @@ protected async Task OnSelectedDateAsync(DateTime? value)
     {
         Opened = false;
 
-        var updatedValue = Value?.TimeOfDay != TimeSpan.Zero
+        DateTime? updatedValue = value;
+
+        if (Value is not null && value is not null)
+        {
+            updatedValue = Value?.TimeOfDay != TimeSpan.Zero
             ? (value ?? DateTime.MinValue).Date + Value?.TimeOfDay
```
dvoituron, line 74: Why do you check `value` with `??` if the previous test already that?

dvoituron (changes requested):
Could you add a UnitTest to validate this particular case?

src/Core/Components/DateTime/FluentDatePicker.razor.cs
MarvinKlein1508: my bad. I removed it :)

MarvinKlein1508:
@dvoituron added UNIT-Test

dvoituron (approved):

Issue #2225, opened 2024-06-20
fix: FluentOverflow does not fit to container when ChildContent is modified


S-Syrichko:
# 🐛 Bug Report

When using FluentOverflow with FluentBadge and modifying badge values, the FluentOverflow component does not update it's width.

## 💻 Repro or Code Sample

```csharp 
<button id="myPopoverButton" style="width: 200px; height: 30px; overflow-x: visible; border-width: 1px; border-radius: 4px; box-sizing: content-box; padding: 4px;" @onclick="() => _showPopover = !_showPopover">
    <FluentOverflow Style="width: 100%;">
        <ChildContent>
            @foreach(var item in categoryItems)
            {
                <FluentOverflowItem Style="background-color: #ffd800; border-radius: 4px;">
                    <FluentBadge Appearance="Appearance.Lightweight">@item.Name</FluentBadge>
                </FluentOverflowItem>
            }
        </ChildContent>
        <MoreButtonTemplate >
            <FluentBadge Appearance="Appearance.Lightweight" Style="width: 32px; border-radius: 4px; background-color: #ffd800;">
                @($"+{context.ItemsOverflow.Count()}")
            </FluentBadge>
        </MoreButtonTemplate>
    </FluentOverflow>
</button>

<FluentPopover AnchorId="myPopoverButton" Style="width: 300px" @bind-Open="_showPopover">
        <Body>
            <FluentGrid Justify="JustifyContent.FlexStart" Spacing="3">
                <FluentGridItem xs="11">
                    <FluentSortableList Items="categoryItems" OnUpdate="@SortList" Style="width: 100%;">
                        <ItemTemplate>
                            <FluentTextField @bind-value="@context.Name" Minlength="4" Style="width: 80%; margin-right: 15px;"></FluentTextField>
                            <FluentIcon Value="@(new Icons.Regular.Size16.ChevronUpDown())" />
                        </ItemTemplate>
                    </FluentSortableList>
                </FluentGridItem>
                <FluentGridItem xs="11">
                    <FluentButton Appearance="Appearance.Stealth" OnClick="@Save">
                        <FluentIcon Value="@(new Icons.Regular.Size28.Save())" />
                    </FluentButton>
                </FluentGridItem>
            </FluentGrid>
        </Body>
    </FluentPopover>
    
@code 
{
    protected string ClassValue => "customOverflow";
    protected string CategoryButton => "categoryButton";

    bool _showPopover;
    
    [Parameter]
[...]

Issue #2224, opened 2024-06-19
fix: FluentOverflow initially paints and then repaints if an overflow is detected, leading to a flash in FluentOverflow


adamint:
# 🐛 Bug Report

As mentioned in title, this is caused by FluentOverflow components initially painting and then re-painting, leading to the flash you mention. This can also be seen on initial page load in any FluentOverflow component, including the example component [here](https://www.fluentui-blazor.net/Overflow) if you make the browser width small enough.

FluentOverflow should not visibly paint and then repaint on first component render. Perhaps it could paint but have `visibility: hidden`, calculate whether overflow should be applied, and then make itself visible.

## 💻 Repro or Code Sample

Make screen width on [this page](https://www.fluentui-blazor.net/Overflow) small enough such that component will have the overflow button appear immediately after component load. Notice a flash.
 
## 🔦 Context

Appears in https://github.com/dotnet/aspire/issues/4417

## 🌍 Your Environment

Windows, Edge (latest stable), v4.8.0

vnbaaij:
The overflow component needs to initialize and assess which items don't fit and then hide them. This is what is causing the 'flash'. You see it rendering and then **must** repaint to hide the badges that are outside of the viewable area. 
I'm not considering this a bug but more a consequence of the way that this component works.

Initially setting `visibility: hidden` and then changing that once calculations are done, works but does not lead to a better result in my opinion as the 'flash' is then much more prominent with the whole block appearing.

![overflow-test](https://github.com/microsoft/fluentui-blazor/assets/1761079/02317b30-94f3-41da-a75a-2e81b4aeda34)

Maybe the effect is less visible with a different background color/other styling.

Shall we create a parameter so that the developer can choose which behavior the component uses? Something like `VisibleOnLoad` with a default of true to not break existing usage.

vnbaaij:
New parameter added with PR #2236

PR #2236, opened 2024-06-20
[Overflow] Add refresh method, add VisibleOnLoad parameter

vnbaaij:
- Rename functions in js to be in line with rest of scripts
- Add public RefreshAsync method to call refresh function in script (#2225)
- Add VisibleOnLoad parameter (#2224)


src/Core/Components/Overflow/FluentOverflow.razor.cs
```diff
@@ -88,8 +90,16 @@ protected override async Task OnAfterRenderAsync(bool firstRender)
         {
             _jsModule = await JSRuntime.InvokeAsync<IJSObjectReference>("import", JAVASCRIPT_FILE);
             _dotNetHelper = DotNetObjectReference.Create(this);
+            await _jsModule.InvokeVoidAsync("fluentOverflowInitialize", _dotNetHelper, Id, IsHorizontal, Selectors);
```
dvoituron, line 93: Could you call `RefreshAsync` to have only one call to this JS method?

dvoituron (changes requested):

src/Core/Components/Overflow/FluentOverflow.razor.cs
vnbaaij: No, the 'initialize code' is different JS call then for the 'refresh code'. Initialize sets up the MutationObserver and ResizeObserver and then calls the (re)calculation on the JS side. Resize just calls the (re)calculation

dvoituron: Hups... I didn't have my glasses ;-)

src/Core/Components/Overflow/FluentOverflow.razor.cs
```diff
@@ -24,6 +25,7 @@ public partial class FluentOverflow : FluentComponentBase, IAsyncDisposable
 
     /// <summary />
     protected string? StyleValue => new StyleBuilder(Style)
+        .AddStyle("visibility",_visibility)
```
dvoituron, line 28: Probably better to use `.AddStyle("visibility", "hidden", when: VisibleOnLoad == true)`.
So, you can remove the `_visibility` variable and the style will be applied only when the property is True (and the user could override the style in other cases).

dvoituron (changes requested):

dvoituron (approved):

Issue #2230, opened 2024-06-20
fix: Nullable value not working in `FluentTimePicker`


hksalessio:
# 🐛 Bug Report

`FluentTimePicker` supports nullable values but when inserting and removing a value, the value stays at the last value (which is 00:00 because backspace sets it first back to 00:00 and then to --:--).
This is observable in the example on the demo page: https://www.fluentui-blazor.net/DateTime#defaulttimepicker

## 💻 Repro or Code Sample

https://github.com/microsoft/fluentui-blazor/assets/143792830/88f79532-3b7a-4bec-ad06-9be8da2717db

## 🤔 Expected Behavior

When using backspace to clear the value, it should still change to 00:00 first and when using backspace again it should change to null.

## 😯 Current Behavior

Using backspace twice leaves the value at 00:00.

PR #2245, opened 2024-06-23
[TimePicker] Add null check for empty string

MarvinKlein1508:
# Pull Request

## 📖 Description
This PR adds a null check for the `FluentTimePicker` which converts empty strings to `DateTime?`.

### 🎫 Issues

* #2230

## 📑 Test Plan

I've added a new UNIT Test `FluentTimePicker_EmptyStringToNull` to verfiy the results.

### General

### DIFF
--- a/src/Core/Components/DateTime/FluentTimePicker.razor.cs
+++ b/src/Core/Components/DateTime/FluentTimePicker.razor.cs
@@ -26,7 +26,7 @@ public partial class FluentTimePicker : FluentInputBase<DateTime?>
         }
         else
         {
-            result = Value?.Date;
+            result = string.IsNullOrWhiteSpace(value) ? null : Value?.Date;
         }
 
         validationErrorMessage = null;
--- a/src/Core/Components/DateTime/FluentTimePicker.razor.cs
+++ b/src/Core/Components/DateTime/FluentTimePicker.razor.cs
@@ -20,13 +20,17 @@ public partial class FluentTimePicker : FluentInputBase<DateTime?>
     {
         DateTime currentValue = Value ?? DateTime.MinValue;
 
-        if (value != null && DateTime.TryParse(value, out var valueConverted))
+        if (string.IsNullOrWhiteSpace(value))
+        {
+            result = null;
+        }
+        else if (value != null && DateTime.TryParse(value, out var valueConverted))
         {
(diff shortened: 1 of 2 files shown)


src/Core/Components/DateTime/FluentTimePicker.razor.cs
```diff
@@ -26,7 +26,7 @@ protected override bool TryParseValueFromString(string? value, out DateTime? res
         }
         else
         {
-            result = Value?.Date;
+            result = string.IsNullOrWhiteSpace(value) ? null : Value?.Date;
```
dvoituron, line 29: Could you use something like that, to avoid to call `DateTime.TryParse` for empty string.

```csharp
if (string.IsNullOrWhiteSpace(value)) 
{
    result = null;
}
else if (value != null && DateTime.TryParse(value, out var valueConverted))
{
    result = currentValue.Date + valueConverted.TimeOfDay;
}
else
{
    result = Value?.Date;
}
```

dvoituron (changes requested):

tests/Core/DateTime/FluentTimePickerTests.cs
```diff
@@ -55,6 +55,19 @@ public void FluentTimePicker_TryParseValueFromString(string? value, string? time
         }
     }
 
+    [Fact]
+    public void FluentTimePicker_EmptyStringToNull()
```
dvoituron, line 59: This test does not validate the correction made: it works even with the code in the `dev` branch.

You should specify a initialized `Value`.

```csharp
[Fact]
public void FluentTimePicker_ResetValue()
{
    // Arrange
    var picker = new TestTimePicker()
    {
        Value = System.DateTime.Today,   // Default date 
    };

    // Set a value
    picker.CallTryParseValueFromString("01:23", out var initialDate, out var _);
    Assert.Equal("01:23:00", initialDate?.ToString("HH:mm:ss"));

    // Reset
    picker.CallTryParseValueFromString(string.Empty, out var resultDate, out var _);
    Assert.Null(resultDate?.ToString("HH:mm:ss"));
}
```

dvoituron (changes requested):

MarvinKlein1508:
@dvoituron should be good now :)

dvoituron (approved):

vnbaaij (approved):

PR #2246, opened 2024-06-23
[Demo] Fixes link for Time picker in demo projects

MarvinKlein1508:
# Pull Request

## 📖 Description

this PR removes the fragment from the demo navigation for the `/DateTime` link. 

### General

### DIFF
--- a/examples/Demo/Shared/Shared/DemoNavProvider.cs
+++ b/examples/Demo/Shared/Shared/DemoNavProvider.cs
@@ -257,7 +257,7 @@ public class DemoNavProvider
                         title: "Text Field"
                     ),
                     new NavLink(
-                        href: "/DateTime#fluenttimepicker-class",
+                        href: "/DateTime",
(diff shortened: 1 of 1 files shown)


vnbaaij (changes requested):
I don't think the fragment should be removed. It just needs to point to the default time picker example on that page.

MarvinKlein1508:
@vnbaaij I removed it since all other components don't have one as well. It would be strange to start the page right in the middle since there is no nav item for FluentDatePicker

vnbaaij:
Yes, it is strange but even stranger to click on a link to time picker and land on a date picker segment 😀.

Time picker is an InputBase derived component. That is why it is shown in Form and input menu. DatePicker is not InputBase derived (yet?). 

We need to address the underlying issue and probably just make separate pages.

MarvinKlein1508:
Got it. Added the fragment. I didn't even know that you can find the FluentDatePicker under components instead.

vnbaaij (approved):

PR #2250, opened 2024-06-24
[TimePicker] Fix TimePicker Null value

dvoituron:
# [TimePicker] Fix TimePicker Null value

**FluentTimePicker** supports nullable values but when inserting and removing a value, the value stays at the last value (which is 00:00 because backspace sets it first back to `00:00` and then to `--:--`).
See #2230

## Before
Before

## After
After


dvoituron:
Sorry, I didn't see it was the same as #2245. Closing.

PR #2258, opened 2024-06-25
Rating component

franklupo:
Rating component

## 📖 Description

Rating component like https://mudblazor.com/components/rating

image


vnbaaij:
This is great! Has been on my list of things to add for quite some time. Will do a full review soon.

Couple of quick remarks. 
- Please follow our 'convention' of giving example pages a "...Page" name (so `RatingPage.razor` in your case)
- In `CoreIcons` I see both stars have `IconVariant.Regular`. Shouldn't one have `IconVariant.Filled`?
- We really need to have Unit Test. Can you try to add some, using the existing tests as your guidance. Thanks

Looking forward to getting this in!

franklupo:
For CoreIcons I copied the previous line. There is an error in the previous line. I fixed it.

dvoituron:
Adding this component in the page `BasicFormFluentUIComponents.razor`, the `Required` attribute doesn't work.

franklupo:
> BasicFormFluentUIComponents
like https://mudblazor.com/components/rating

when the pointer is on a star the previous ones and this one become full, the following ones empty

dvoituron:
> when the pointer is on a star the previous ones and this one become full, the following ones empty

Why this comment? I don't understand :-)
If it's related to my GIF, the problem is that the user is between two stars and you consider this area to be "external". You'd have to apply the "PointerOver" effect to the whole area (in purple below) and not just to the icon.

image

This effect does not occur with MudBlazor.

vnbaaij:
One other thing...we have a pretty high bar on accessibility so adding keyboard support like [the MudBlazor one has](https://mudblazor.com/components/rating#keyboard-navigation) should be included as well.

We can use the `FluentKeyCodeProvider` for it. See for example `FluentNavGroup.razor`

src/Core/Components/Rating/FluentRating.razor.cs
```diff
@@ -0,0 +1,88 @@
...
+        else
+        {
+            validationErrorMessage = string.Format(CultureInfo.InvariantCulture,
+                                                   "The {0} field must be a number.",
+                                                   FieldBound ? FieldIdentifier.FieldName : UnknownBoundField);
+            return false;
+        }
+    }
+
+    protected internal async Task HandleKeyDownAsync(KeyboardEventArgs keyboardEventArgs)
```
franklupo, line 55: Keyboard usage

vnbaaij: Are you sure about the logic? 
Shift+ArrowRight is now MaxValue (say 10) - Value (say 4) so with `SetCurrentValueAsync` you set Value to 6?

Also, please make ArrowDown do the same as ArrowLeft and make ArrowUp do same as ArrowRight (and shift + down/up as well of course)

vnbaaij (changes requested):
Would you think Size24 is too big for the icons? If not, we could change the component to also support half values for a rating. (StarHalf icon is not available in Size20). Would of course require more change like `FluentInputBase<double>`, etc but think it could be valuable. 
We have an internal React rating component version that also supports half ratings

src/Core/Components/Rating/FluentRating.razor.cs
dvoituron: I agree, I would find it cleaner to use a classic `switch` and add `Value++` or `Value = MaxValue` in the cases.

franklupo:
> Would you think Size24 is too big for the icons? If not, we could change the component to also support half values for a rating. (StarHalf icon is not available in Size20). Would of course require more change like `FluentInputBase<double>`, etc but think it could be valuable. We have an internal React rating component version that also supports half ratings

IMO for now we could ignore the half values.

src/Core/Components/Rating/FluentRating.razor
```diff
@@ -0,0 +1,31 @@
...
+             @onkeydown="HandleKeyDownAsync"
+             tabindex="@(Disabled ? -1 : 0)"
+             aria-readonly="@(ReadOnly.ToString().ToLower())"
+             HorizontalGap="0">
+    @for (int i = 1; i <= MaxValue; i++)
+    {
+        var currentValue = i;
+        @if (ReadOnly || Disabled)
+        {
+            <FluentIcon Value="@GetIcon(currentValue)" Color="@(Disabled ? Color.Disabled :  IconColor)" />
```
dvoituron, line 20: We attach great importance to the accessibility of our components. Unfortunately, it's not currently compatible. I think this is mainly because there is no `aria-xxx` on the Icons components.

You can check this by installing https://www.nvaccess.org/download/ and checking the React lib component (which is Accessible) and this new FluentRating component.

If it's too complex in this PR, I suggest adding a note to the documentation and opening a second PR to correct it as soon as possible.

src/Core/Components/Rating/FluentRating.razor.cs
```diff
@@ -0,0 +1,87 @@
...
+public partial class FluentRating : FluentInputBase<int>
+{
+    private int? _mouseOverValue;
+
+    public FluentRating() => Id = Identifier.NewId();
+
+    /// <summary>
+    /// Gets or sets the maximum value.
+    /// </summary>
+    [Parameter] public int MaxValue { get; set; } = 10;
```
dvoituron, line 17: A default value of `5` is preferable (to be consistent with the React Lib)

examples/Demo/Shared/Pages/Rating/Examples/RatingDefault.razor
```diff
@@ -0,0 +1,2 @@
+﻿
```
dvoituron, line 1: Can you explain the purpose of this component? 

Example:
> A **FluentRating** component allows users to provide a rating for a particular item.
> **FluentRating** allows customers to determine a sense of value of a good or a service. 
> By default, the rating is selected out of 5 stars, but the number and symbol used can be customized.

dvoituron (commented):
⚠️ Don't forget to add unit tests (we can't publish this component if no tests are included).

Check examples in `FluentButtonTests.razor`
```csharp
[Fact]
public void FluentButton_Title()
{
    // Arrange && Act
    var cut = Render(@<FluentButton Title="My Title">My button</FluentButton>);

    // Assert
    cut.Verify();
}
```

src/Core/Components/Rating/FluentRating.razor.cs
```diff
@@ -32,6 +32,11 @@ public partial class FluentRating : FluentInputBase<int>
     /// </summary>
     [Parameter] public Icon IconEmpty { get; set; } = new CoreIcons.Regular.Size20.Star();
 
+    /// <summary>
+    /// Fires when hovered value changes. Value will be null if no rating item is hovered.
+    /// </summary>
+    [Parameter] public EventCallback<int?> HoveredValueChanged { get; set; }
```
dvoituron, line 46: I wouldn't use this event. If the user wants to display a label, it shouldn't be at the MouseOver but after selecting a value. Making a component too complex risks confusing it :-)

For example, in the case of this event, what about mobiles (and Touch events) and what about accessibility?

dvoituron (changes requested):

src/Core/Components/Rating/FluentRating.razor.cs
franklupo: I don't use mouseover but onpointerover. I use it to manage the icon change and then I invoke HoveredValueChanged

src/Core/Components/Rating/FluentRating.razor
```diff
@@ -0,0 +1,31 @@
...
+             Class="ClassValue"
+             Style="StyleValue"
+             role="radiogroup"
+             @onkeydown="HandleKeyDownAsync"
+             tabindex="@(Disabled ? -1 : 0)"
+             aria-readonly="@(ReadOnly.ToString().ToLower())"
+             HorizontalGap="0">
+    @for (int i = 1; i <= MaxValue; i++)
+    {
+        var currentValue = i;
```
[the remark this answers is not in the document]
franklupo: if I use **i** not work in event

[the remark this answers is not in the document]
dvoituron: Oh. Got it... reference issue. OK

src/Core/Components/Rating/FluentRating.razor.cs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        await SetCurrentValueAsync(value);
+    }
+
+    private void OnPointerOut() => _mouseOverValue = null;
+    private void OnPointerOver(int value) => _mouseOverValue = value;
+    private async Task OnClickAsync(int value)
+    {
+        if (value == Value) { value = 0; }
```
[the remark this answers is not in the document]
franklupo: Same problem with arrow left/right.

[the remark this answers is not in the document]
franklupo: I add AllowReset to enable reset, allow clear when click again

vnbaaij (changes requested):
With the current code, using up/down arrow moves the screen up/down as well:

rating-not-using-keycode

By using the `FluentKeyCode` component, we can stop propagation of key events.

rating-using-keycode

Changing to this is quite simple.
1) in `FluentRating.razor` add before input label:
```
<FluentKeyCode Anchor="@Id"
               PreventDefaultOnly="new[] { KeyCode.Left, KeyCode.Right, KeyCode.Up, KeyCode.Down, KeyCode.Shift }"
               StopPropagation="@true"
               OnKeyDown="@HandleKeyDownAsync" />
```
2) remove `@onkeydown=...` from the FluentStack

3) In `FluentRating.razor.cs` change `HandleKeyDownAsync` to:
```
protected internal async Task HandleKeyDownAsync(FluentKeyCodeEventArgs e)
 {
     if (e.TargetId != Id || Disabled || ReadOnly)
	 {
         return;
     }

     int value = e.Key switch
     {
         KeyCode.Right or KeyCode.Up when e.ShiftKey => value = MaxValue,
         KeyCode.Right or KeyCode.Up => value = Math.Min(Value + 1, MaxValue),
         KeyCode.Left or KeyCode.Down when e.ShiftKey => value = 0,
         KeyCode.Left or KeyCode.Down => value = Math.Max(Value - 1, 1),
         _ => Value
     };

     _mouseOverValue = null;
     _mouseOverDisabled = true;

     await SetCurrentValueAsync(value);
 }
```

vnbaaij:
In the rendered component we get:
image

This is because `ClassValue` and `StyleValue` have not been declared in the code behind but are being assigned to the component parameters.

In `FluentRating.razor.cs` you need to add:
```
 /// <summary />
 protected string? ClassValue => new CssBuilder(Class)
     .AddClass("fluent-rating")
     .Build();

 /// <summary />
     .Build();
```

vnbaaij:
@franklupo Both Denis and I are creating a lot of comments/remarks/suggestions. 

Please, please **do not see this as criticism**. We are really happy with your contributions. We are just trying to look after quality and coding standards.

franklupo:
> @franklupo Both Denis and I are creating a lot of comments/remarks/suggestions.
> 
> Please, please **do not see this as criticism**. We are really happy with your contributions. We are just trying to look after quality and coding standards (and can be real nitpickers about that sometimes 😑).

I understand, no problem. I have a lot to learn.

src/Core/Components/Rating/FluentRating.razor.cs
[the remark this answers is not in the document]
dvoituron: You can add the `AllowReset` property, but the problem is still there :-)

I think the simplest solution is to manage the process as React lib does: avoid resetting the value, unless the developer wants to do so, by adding an external link/button `Value = 0`.
This is also how Amazon works: you can't double-click to clear the selection, but you have an external link.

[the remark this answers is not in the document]
franklupo: I fix a problem of overrides "mouseover" using internal variable _mouseOverDisabled.

src/Core/Components/Rating/FluentRating.razor
franklupo: do you mean there should be aria-labels for each icon?

dvoituron: I think so... but you need to validate using NVDA or the Windows Narrator.

franklupo:
> ⚠️ Don't forget to add unit tests (we can't publish this component if no tests are included).
> 
> Check examples in `FluentButtonTests.razor`
> 
> ```cs
> [Fact]
> public void FluentButton_Title()
> {
>     // Arrange && Act
>     var cut = Render(@<FluentButton Title="My Title">My button</FluentButton>);
> 
>     // Assert
>     cut.Verify();
> }
> ```

I'm having trouble with tests. Could you help me create one?

dvoituron:
> one

Sure. I will create some Unit Tests later in the day 😊

src/Core/Components/Rating/FluentRating.razor
```diff
@@ -0,0 +1,31 @@
...
+        @if (ReadOnly || Disabled)
+        {
+            <FluentIcon Value="@GetIcon(currentValue)" Color="@(Disabled ? Color.Disabled :  IconColor)" />
+        }
+        else
+        {
+            <FluentIcon Value="@GetIcon(currentValue)"
+                        Color="IconColor"
```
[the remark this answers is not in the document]
dvoituron: To have the same "size" as the React component, the default value should be "28px".
if not, the size is about 16px...
image

src/Core/Components/Rating/FluentRating.razor.cs
```diff
@@ -0,0 +1,142 @@
...
+        {
+            return;
+        }
+
+        int value;
+        switch (e.Key)
+        {
+            case "ArrowRight" when e.ShiftKey: value = MaxValue; break;
+            case "ArrowRight" or "ArrowUp": value = Math.Min(Value + 1, MaxValue); break;
+            case "ArrowLeft" when e.ShiftKey: value = 1; break;
```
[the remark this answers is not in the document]
dvoituron: These keys must be explained in the documentation

dvoituron:
Example of `FluentRatingTests.razor` file:
1. Create this test file and insert this code.
2. Run the Unit Tests (using the Text Explorer) and you will find 2 tests failed (due to `cut.Verify();`)
    a. You will have 2 files `FluentRatingTests.FluentRating_[xxx].received.razor.html`
    b. Rename the `received` part to `verified` and check if the HTML content seems correct.
3. Run again the tests... all will be fine.

Your job:
1. Add a test for each property/attribute to validate each feature.
2. Add other tests to validate other parts of the code.
3. Run the `_StartCodeCoverage.cmd` file to verify the Code Coverage of your component (**at least 80%**)
    Read the first line of this file to install the necessary tools.

image

```csharp
@using Xunit;
@inherits TestContext
@code
{
    public FluentRatingTests()
    {
        JSInterop.Mode = JSRuntimeMode.Loose;
        Services.AddSingleton(new LibraryConfiguration());
    }

    [Fact]
    public void FluentRating_Empty()
    {
        // Arrange && Act
        var cut = Render(@<FluentRating />);

        // Assert
        cut.Verify();
    }

    [Fact]
    public void FluentRating_Value()
    {
        // Arrange && Act
        var cut = Render(@<FluentRating MaxValue="10" Value="2" />);

        // Assert
        cut.Verify();
    }

    [Fact]
    public void FluentRating_SelectValue()
    {
        int value = 0;

        // Arrange
        var cut = Render(@<FluentRating @bind-Value="@value" />);

        // Act: Click on the second star
        cut.FindAll("svg").ElementAt(1).Click();

        // Assert
        Assert.Equal(2, value);
    }

    [Fact]
    public void FluentRating_AllowResetFalse()
    {
        int value = 0;

        // Arrange
        var cut = Render(@<FluentRating @bind-Value="@value" AllowReset="false" />);

        // Act: Click twice on the second star
        cut.FindAll("svg").ElementAt(1).Click();
        cut.FindAll("svg").ElementAt(1).Click();

        // Assert
        Assert.Equal(2, value);
    }

    [Fact]
    public void FluentRating_AllowResetTrue()
    {
        int value = 0;

        // Arrange
        var cut = Render(@<FluentRating @bind-Value="@value" AllowReset="true" />);

        // Act: Click twice on the second star
        cut.FindAll("svg").ElementAt(1).Click();
        cut.FindAll("svg").ElementAt(1).Click();

        // Assert
        Assert.Equal(0, value);
    }
}

```

src/Core/Components/Rating/FluentRating.razor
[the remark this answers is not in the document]
franklupo: ok, I changed the default value `28px`. the size React component is of the container not of icon

dvoituron:
Could you tell us when your PR will be finished (all discussions resolved)?
That way, we can check the code one last time before merging it.

franklupo:
If I haven't forgotten anything, that's fine with me.

dvoituron:
> If I haven't forgotten anything, that's fine with me.

And Unit Tests (to have 80%) and the [Accessiblity issue](https://github.com/microsoft/fluentui-blazor/pull/2258/files/868c559ab736082f2f50056d12c4a45e9a6317f7#r1653023692) ?

src/Core/Components/Rating/FluentRating.razor
franklupo: I would open a new PR later. Now I don't know how to solve it.

vnbaaij: No, this accessibility needs to be good from the start.

I believe it is only a matter of adding `aria-label="@currentValue"` to the FluentIcon

vnbaaij (approved):

dvoituron (approved):

vnbaaij:
@franklupo thanks again for your contribution! We are merging this in now and will make some final tweaks before release.

franklupo:
To add aria-label="@currentValue" and the new unit test do I need to create a new PR?

dvoituron:
@franklupo I don't know why (not yet check), but this component is instable (flicker effect) :-(
Can you check how to fix that?

You can go to https://preview.fluentui-blazor.net/Rating to reproduce this behavior.

https://github.com/microsoft/fluentui-blazor/assets/8350694/f06a5b63-1dec-4185-a309-0b23a43f3839

PR #2238, opened 2024-06-20
[DataGrid] Add WCAG 2.2 single-click column resize capability

vnbaaij:
**Do not auto-merge. Need to clear out IssueTester before.**

**Update:** the `three-dot/more` icon has been replaced with the chevron down icon. Also, all header icons have an opacity set now to make them a bit 'softer'

image

Allows for resizing columns in a DataGrid by using the column options menu. Two different modes are available: discrete and exact.
- Discrete: resize the column is steps of -10/+10 pixels (same as +/- keyboard resizing)
- Exact: specify the column width in pixels

Both options offer a reset button to go back to the original grid setting (same as shift+r keyboard reset)

As the options menu shows more then just a column filter, the icon to indicate an option menu is available is changed to the three dots. The indicator for an applied filter has been moved to be shown besides the title akin to the sort icon.

### Resize discrete:
datagrid-resize-discrete
### Resize direct:
datagrid-resize-exact

PR #2265, opened 2024-06-25
[Slider] Fix getting caught in update loop

vnbaaij:
This PR fixes the following behavior:

slider-issue

Is needed for PR #2238


jaydeboer:
We just ran into this problem too!  Glad to see the fix is in the works.

dvoituron (approved):

PR #2270, opened 2024-06-26
[Docs] Improve API documentation menu button

franklupo:
Reorder menu for button

image


vnbaaij (changes requested):
Good idea to group these (there will be more variants with the next version of the web components).
To prevent having to come up with group icons I propose to use `Icons.Regular.Size20.AppFolder()` for all groups (we can change that for the other menu items in a separate PR. Please make this change in this PR and then also change the Button entry to use the Button icon (instead of ControlButton).

vnbaaij (approved):

PR #2275, opened 2024-06-27
[Test] Fix temp folder coverage

franklupo:
Use temp folder user `%USERPROFILE%\AppData\Local\Temp`

### DIFF
--- a/tests/Core/_StartCodeCoverage.cmd
+++ b/tests/Core/_StartCodeCoverage.cmd
@@ -18,6 +18,8 @@ REM     $:\> explorer C:\Temp\Coverage\index.html
 echo on
 cls
 
+mkdir %USERPROFILE%\AppData\Local\Temp\FluentUI\Coverage
+
 dotnet test /p:CollectCoverage=true /p:CoverletOutputFormat=cobertura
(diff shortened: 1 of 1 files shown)


dvoituron:
Why this request?
What is the advantage to use `%USERPROFILE%\AppData\Local\Temp` and not `C:\Temp` (shorter)?

tests/Core/_StartCodeCoverage.cmd
```diff
@@ -18,6 +18,8 @@ REM     $:\> explorer C:\Temp\Coverage\index.html
 echo on
 cls
 
+mkdir %USERPROFILE%\AppData\Local\Temp\FluentUI\Coverage
```
dvoituron, line 21: You don't need to create the folder. This will be done with the `reportgenerator` command.

dvoituron (changes requested):

franklupo:
normally it does not exist, and for safety the directory on c:\ is not recommended

vnbaaij:
> normally it does not exist, 

As @dvoituron says, it gets created automatically.

> and for safety the directory on c:\ is not recommended

On a general level I agree, but this is only ever being used by developers that are contributing to and writing test code for this specific repository. So very aware and highly professional individuals that understand what is happening. We don't need to burie the (temp) result that deep in the file system.

PR #2278, opened 2024-06-27
[DataGrid] Rewriting event propagation

franklupo:
To simplify event propagation and for new columns I added events to ColumnBase. FluentDataGridCell and FluentDataGridRow call FluentDataGrid. FluentDataGrid calls events in columns.

It seems like a clean solution and the logic for each column stays inside it. In FluentDataGrid you don't need to add anything.

I'm not sure what the HandleEventAsync event does in FluentDataGridRow


vnbaaij (commented):
I can understand the changes on the `FluentDataGrid...` component although I don't necessarily feel it makes the code clearer. 
You are moving handling of row/cell specific events to a different location then from where it is actually being invoked (I mean handle cell click in datagrid instead of in datagridcell). No better separation of concerns, right?

What I don't like is having row based events in the realm of a column. In fact, I think that makes the code more unclear than it is.

Feels to me a bit like a refactoring because of refactoring but not providing any real added value or benefit. Am I missing something?

dvoituron:
> Feels to me a bit like a refactoring because of refactoring but not providing any real added value or benefit. Am I missing something?

I'm essentially of the same opinion. Moreover, why make these `Func` public (and not overloadable). In addition, Blazor events generally pass through `EventCallback`.

franklupo:
If I created a new column, I could use the events without asking for any changes to the library.
Now SelectColumn is too integrated into the grid code.
If I wanted to make my own SelectColumnNew now I couldn't do it.

Ok instead of Func use EventCallback

vnbaaij:
> If I created a new column, I could use the events without asking for any changes to the library. 

There are a lot of things that need to be taken into account for creating a new column. I don't see this as something we will be doing a lot. `SelectColumn` was the first new one after the `DataGrid`  was introduced over 2 years ago.

> Now SelectColumn is too integrated into the grid code. If I wanted to make my own SelectColumnNew now I couldn't do it.

Yes, it s integrated but that is not a bad thing. There are a lot of cogs that need to work together. You not being able to create your own columns is not a bad thing, I think.
 
> Ok instead of Func use EventCallback

But that does not address the fact that we have row functionality in the columns. For me that is a no-go.

vnbaaij:
> How it worked for SelectColumn could work for others. In one of my projects I'm creating custom columns. 

We are not stiving for making the DataGrid expandable like that

> IMO it might be a nice feature.

But again, when the column components contain row functionality it is a no-go

franklupo:
You are right that columns should not have row events but for SelectColumn it happened. 

In this implementation I did not have to create a collection of SelectColumn and check if the event was for the right column, everything is encapsulated in the code of the SelectColumn class.

SelectColumn could be one of those columns that has special implementation on rowclick.

example
```c#
if(MyGrid.AllowSelectRowByClick && !(column is _selectColumn)) 
{
.....
}
```

vnbaaij:
Can you make it work with EventCallback and overloadable instead of public? Then we can maybe just do something with the naming to make it less rubbing me the wrong way

franklupo:
What is this code in **FluentDataGridRow** 

```c#
    /// <summary />
    Task IHandleEvent.HandleEventAsync(EventCallbackWorkItem callback, object? arg)
        => callback.InvokeAsync(arg);

```

dvoituron:
The idea seems good to me, but the way it's proposed seems too complex. Mainly for debugging. 
For example, where currently when you click on a cell, it makes sense to start with the `FluentDataGridCell` component and the code to debug is in the method. With your proposition, we will navigate...
```
FluentDataGridCell.HandleOnCellClickAsync
   > FluentDataGrid.HandleOnCellClickAsync
      => FluentDataGrid.OnCellClick
         ColumnBase.OnCellClick
         => SelectColumn.OnCellClick
```

Give use some days to analyze your proposal to "extract" events to the `ColumnBase' class, for example using inheritance or other solution to simplify the usages.

PS: I will probably create another separate PR to have a new discussion thread.

vnbaaij:
Closing this now that we have #2298

PR #2284, opened 2024-06-27
[Api Doc] Add icon preview

franklupo:
Added preview icon on API documentation

### DIFF
--- a/examples/Demo/Shared/Components/ApiDocumentation.razor
+++ b/examples/Demo/Shared/Components/ApiDocumentation.razor
@@ -35,7 +35,6 @@
                     @* Display list of Enumerations (if available) *@
                     @if (context.EnumValues.Length > 0)
                     {
-
                         var id = Identifier.NewId();
                         <span id="@id">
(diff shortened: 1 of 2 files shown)


dvoituron (commented):
Can you add an screenshot to explain the result?

dvoituron:
Won't this cause confusion with the other icons (i) for example? I'm afraid users will wonder why we put an "Information" symbol there.

Maybe in a tooltip?

vnbaaij:
Or show "Preview: {image}" and show the dual name in a tooltip?

dvoituron:
Also a good solution

franklupo:
image

image

examples/Demo/Shared/Components/ApiDocumentation.razor
```diff
@@ -49,7 +48,28 @@
 
                     }
                 </TemplateColumn>
-                <PropertyColumn Property="@(c => c.Default)" />
+
+@*                 <PropertyColumn Property="@(c => c.Default)" /> *@
```
vnbaaij, line 52: Please remove commented line (to not confuse people) and then I think we are good to go!

vnbaaij (changes requested):

dvoituron (approved):

vnbaaij (approved):

PR #2297, opened 2024-06-28
[Doc] Search for NavMenu

franklupo:
I added Search for NavMenu.

image

### DIFF
--- a/examples/Demo/Shared/Shared/DemoNavMenu.razor
+++ b/examples/Demo/Shared/Shared/DemoNavMenu.razor
@@ -1,14 +1,59 @@
-﻿@inject DemoNavProvider NavProvider
+﻿@using NavLink = FluentUI.Demo.Shared.NavLink
+@inject DemoNavProvider NavProvider
 
 <div class="navmenu">
     <input type="checkbox" title="Menu expand/collapse toggle" id="navmenu-toggle" class="navmenu-icon" />
     <label for="navmenu-toggle" class="navmenu-icon"><FluentIcon Value="@(new Icons.Regular.Size20.Navigation())" Color="Color.Neutral" /></label>
-    <nav class="sitenav" aria-labelledby="main-menu" >
-        <FluentNavMenu Id="main-menu" Title="Main menu" CustomToggle="true" >
-            @foreach(var item in NavProvider.NavMenuItems)
+    <nav class="sitenav" aria-labelledby="main-menu">
+        <FluentSearch @bind-Value="@_term" Placeholder="search"/>
+        <FluentNavMenu Id="main-menu" Title="Main menu" CustomToggle="true">
+            @foreach (var item in navMenuItems.Where(a => a.Visible))
             {
-                <DemoNavMenuItem Value="item"/>
+                <DemoNavMenuItem Value="item" />
             }
         </FluentNavMenu>
     </nav>
 </div>
+
+
+@code
+{
+    string _term = default!;
+
+    IEnumerable<NavItem> navMenuItems
+    {
+        get
+        {
+            foreach (var item in NavProvider.NavMenuItems)
+            {
+                SetVisible(item);
+            }
+
+            return NavProvider.NavMenuItems;
+        }
+    }
+
+    private void SetVisible(NavItem item)
+    {
+        switch (item)
+        {
+            case NavGroup group:
+                foreach (var link in group.Children)
+                {
+                    SetVisible(link);
+                }
+
+                item.Visible = group.Children.Any(a => a.Visible);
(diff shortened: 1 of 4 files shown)


examples/Demo/Shared/Shared/NavItem.cs
```diff
@@ -13,16 +13,23 @@ public abstract record NavItem
     public string? Href { get; init; }
     public NavLinkMatch Match { get; init; } = NavLinkMatch.Prefix;
     public Icon Icon { get; init; } = new Icons.Regular.Size20.Document();
+    public bool Visible { get; set; }
+    public IEnumerable<string> Tags { get; set; } = [];
```
franklupo, line 17: Possible to add tags for search

examples/Demo/Shared/Shared/DemoNavMenu.razor
```diff
@@ -1,14 +1,59 @@
...
+                }
+
+                item.Visible = group.Children.Any(a => a.Visible);
+                break;
+
+            case NavLink:
+                item.Visible = string.IsNullOrEmpty(_term)
+                                    ? true
+                                    : item.Title.Contains(_term, StringComparison.OrdinalIgnoreCase)
+                                        || item.Tags.Any(a => a.Contains(_term, StringComparison.OrdinalIgnoreCase));
```
franklupo, line 53: Filter

vnbaaij:
What is the difference between this and the search top right?

Like the tags idea though. That could be useful.

franklupo:
I like it above the menu, and I think the mobile experience is better.

vnbaaij:
So there is no real difference between the two?

Could you include a video of how it works so we don't have to check out the branch ourselves? (tip, we use [Peek](https://github.com/firatkiral/pypeek) to create animated gifs. Bit lighter then mp4 files)

I agree on the mobile experience argument. But if we are going to do this, then this needs to be a complete PR that also removes the other search option. I see no point in having both.

franklupo:
peek_1

vnbaaij:
Hmm, I would expect it to work more like the current search. So typing 'serv' would give you:
image

I think when showing a search result, having the groups as entry point is not desirable. Just show the results. But then again, it does not give you the context of in what group an item lives anymore. I think adding this this will always introduce trade-offs.

Maybe it is better to not mix these two functionalities (search and navmenu) but just to move existing search to show above the nav menu (so we have a better mobile experience) and integrate in this existing search the tags functionality you added (to get better results) and not filter the menu.

dvoituron:
I don't see the advantage of adding functionality that already exists on the website.

On the other hand, I prefer this location (above the navigation)... but then several points need to be reviewed:
1. It's not a "Search" but a "Filter".
2. Add `Immediate="true" ImmediateDelay="100"`.
3. Groups must be automatically Expanded when filtering is applied (this is currently not the case), and using the "normal" state when no filter is applied.
4. This filter zone is not responsive, which contradicts the advantage of using it on Mobile.
5. I don't agree with changing the data model, adding `Visible` to handle a UI feature. We need to find another way, via a method that returns the list of useful **navMenuItems** for example. And put this code in a `DemoNavMenu.razor.cs` file and not in `@code` section.
6. Who defines tag content? What criteria? It's great to have a feature, but it needs to be used.

dvoituron:
> the idea is to help the user know where he is

You have the title of the page and the indication in the menu on the left for that.

vnbaaij:
> I don't see the advantage of adding functionality that already exists on the website.
> 
Agreed, there should only be one

> On the other hand, I prefer this location (above the navigation)...

Agreed


> 1. It's not a "Search" but a "Filter".

I think and end user does not make that distinction.

> 2. Add `Immediate="true" ImmediateDelay="100"`.

Agreed

> 3. Groups must be automatically Expanded when filtering is applied (this is currently not the case), and using the "normal" state when no filter is applied.

Agreed, but this wil be difficult to achieve as we don't have a notions of parent/child in the menu data structures

> 4. This filter zone is not responsive, which contradicts the advantage of using it on Mobile.
Agreed

> 5. I don't agree with changing the data model, adding `Visible` to handle a UI feature. We need to find another way, via a method that returns the list of useful **navMenuItems** for example. And put this code in a `DemoNavMenu.razor.cs` file and not in `@code` section.

Agreed

> 6. Who defines tag content? What criteria? It's great to have a feature, but it needs to be used.

We could perhaps better use the `Data` property that we already have for this tagging info (every `FluentComponentBase` derived component has `Data` parameter). Then the dev can just add that either in a declarative ay or when building the `Items` collection

vnbaaij:

PR #2308, opened 2024-07-02
[NumberField] Add sbyte type

franklupo:
Add sbyte type to FluentNumberField

### DIFF
--- a/src/Core/Components/Base/InputHelpers.cs
+++ b/src/Core/Components/Base/InputHelpers.cs
@@ -50,7 +50,6 @@ internal static class InputHelpers<TValue>
         return value;
     }
 
-
     internal static void ValidateSByteInputs(string max, string min)
     {
         var maxValue = Convert.ToSByte(max, CultureInfo.InvariantCulture);
--- a/src/Core/Components/Base/InputHelpers.cs
+++ b/src/Core/Components/Base/InputHelpers.cs
@@ -50,10 +50,10 @@ internal static class InputHelpers<TValue>
         return value;
     }
 
-    internal static void ValidateSByteInputs(string max, string min)
+    internal static void ValidateSByteInputs(string? max, string? min)
     {
-        var maxValue = Convert.ToSByte(max, CultureInfo.InvariantCulture);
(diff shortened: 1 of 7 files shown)


src/Core/Components/Base/InputHelpers.cs
```diff
@@ -51,17 +50,27 @@ public static string GetMinValue()
         return value;
     }
 
-    internal static void ValidateIntegerInputs(string? max, string? min)
+
```
dvoituron, line 53: Can you Put only 1 space line?

src/Core/Components/Base/InputHelpers.cs
```diff
@@ -51,17 +50,27 @@ public static string GetMinValue()
...
+    internal static void ValidateSByteInputs(string max, string min)
     {
+        var maxValue = Convert.ToSByte(max, CultureInfo.InvariantCulture);
+        var minValue = Convert.ToSByte(min, CultureInfo.InvariantCulture);
+
+        if (maxValue < minValue)
+        {
+            throw new ArgumentException("Float Max value is smaller than Min value.");
```
dvoituron, line 61: Message incorrect

src/Core/Components/Base/InputHelpers.cs
```diff
@@ -51,17 +50,27 @@ public static string GetMinValue()
         return value;
     }
 
-    internal static void ValidateIntegerInputs(string? max, string? min)
+
+    internal static void ValidateSByteInputs(string max, string min)
```
dvoituron, line 54: Why arguments are nullable in all other methods and not your new one?

dvoituron (changes requested):

src/Core/Components/Base/InputHelpers.cs
franklupo: `Convert.ToSByte ` not accept null value

vnbaaij: Docs say yes:
image

dvoituron (approved):

PR #2311, opened 2024-07-02
[Combobox, Select] Fix the Placeholder attribute

dvoituron:
# [Select] Fix the Placeholder attribute

This PR modifies the `FluentSelect` component to correctly handle the `Placeholder` attribute. 
Initially, this attribute is not available for the webcomponent `fluent-select`. But following the #2301 discussion, a workaround is possible by adding a "fake" invisible first element. This allows it to be displayed when the user opens the form for the first time. only. It then disappears when the user selects an item, and is not present in the list displayed on the screen.

Thanks @gotmoo for this tip :-)

```xml
<FluentSelect TOption="Person"
              Label="Select a person"
              Placeholder="Make a selection..."
              ... />
```
peek_2

## Unit tests

A unit test has been added to validate this functionality.


vnbaaij (approved):
Nice!

PR #2334, opened 2024-07-05
add 'PickerMonthChanged' event and 'DaysTemplate' property to date picker

ehsangfl:
## 📖 Description

add `PickerMonthChanged` event and `DaysTemplate` property to date picker and bind Calender to those

### 🎫 Issues
cannot use `DaysTemplate `  Picker `PickerMonthChanged ` and  while using Datepicker


src/Core/Components/DateTime/FluentDatePicker.razor.cs
```diff
@@ -35,6 +35,15 @@ public FluentDatePicker()
     [Parameter]
     public EventCallback<bool> OnCalendarOpen { get; set; }
 
+    [Parameter]
```
dvoituron, line 38: Can you add the XML comment for the documentation?

dvoituron (changes requested):

PR #2336, opened 2024-07-07
[DatePicker] Add 'PickerMonthChanged' event and 'DaysTemplate' parameter

ehsangfl:
add PickerMonthChanged event and `DaysTemplate` property to date picker and bind Calender to those

🎫 Issues
cannot use `DaysTemplate`  Picker `PickerMonthChanged`  and while using Datepicker


src/Core/Components/DateTime/FluentDatePicker.razor.cs
```diff
@@ -32,9 +32,24 @@ public FluentDatePicker()
...
+    /// raised when calendar popup opened
+    /// </summary>
     [Parameter]
     public EventCallback<bool> OnCalendarOpen { get; set; }
 
+    /// <summary>
+    /// Defines the appearance of a Day cell.
+    /// </summary>
+    [Parameter]
+    public RenderFragment<FluentCalendarDay> DaysTemplate { get; set; }
```
vnbaaij, line 45: This needs to be a nullable type. Will give compilation errors otherwise,

vnbaaij (changes requested):

vnbaaij (approved):

PR #2342, opened 2024-07-08
[DataGrid] Make `Grid` in `ColumnBase` public again

vnbaaij:
Title says it all

### DIFF
--- a/src/Core/Components/DataGrid/Columns/ColumnBase.razor.cs
+++ b/src/Core/Components/DataGrid/Columns/ColumnBase.razor.cs
@@ -126,7 +126,7 @@ public abstract partial class ColumnBase<TGridItem>
     /// <summary>
     /// Gets a reference to the enclosing <see cref="FluentDataGrid{TGridItem}" />.
     /// </summary>
-    internal FluentDataGrid<TGridItem> Grid => InternalGridContext.Grid;
+    public FluentDataGrid<TGridItem> Grid => InternalGridContext.Grid;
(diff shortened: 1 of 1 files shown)


dvoituron:
Why do we need to set `public`? Can we set `protected` ?

vnbaaij:
When devs want to create their own custom column (derived from ColumnBase), they need to be able to access the Grid parameter.

liaan-mynhardt:
I have another use case for Grid to be public. If I wanted to create a reusable column header component that I pass in using HeaderCellItemTemplate I can do so using the "context" parameter eg. 
`<CustomizableDataGridColumnHeader Column="context" />`
by changing it to protected I will have to do 
`<CustomizableDataGridColumnHeader Grid="topGrid" Column="context" />`
which opens me up to passing in the wrong grid reference.
Thoughts?

vnbaaij:
I can live with that,

Protected is meant for giving access for derived classes so it covers the use case well. Dev should be able to spot/fix easily

aDisplayName:
> When devs want to create their own custom column (derived from ColumnBase), they need to be able to access the Grid parameter. Protected should suffice for that as well yes. Will change.

Oh god, it just broken our existing code......

vnbaaij:
> Oh god, it just broken our existing code......

Take a look at how it is done in the Aspire repo. They have created their own `PropertyColumn` and `TemplateColumn` implementations with Grid being protected.

PR #2377, opened 2024-07-15
[Docs] Improve API Documentation performance

franklupo:
This change moves the object creation to read `default values` ​​only if needed. This improve speed and allows adding static classes to the documentation.


vnbaaij (changes requested):
This code is good. 

Could you please also change line 56 in ApiDocumentation.razor to `<div>Preview:&nbsp;</div>` in this PR. the missing whitespace between the word and icon is nagging me 😑

vnbaaij (approved):

PR #2388, opened 2024-07-16
[JavaScript Caching] Cache management

dvoituron:
# [JavaScript Caching] Cache management

1. Addition of a `LibraryConfiguration` property, to be compatible with existing code and to allow the developer to configure how the URL is formatted. By default, the new version will include `?v=4.9.2.0` (the assembly version) at the end of all imported JavaScript files (only the files we control); But the dev could customize this Query String.

```csharp
builder.Services.AddFluentUIComponents(config => config.CollocatedJavaScriptQueryString = (url) => $"v={AssemblyVersion}");
```

2. The result will be...

image

## Updates

All `import` commands are updated using:
1. Inject the `LibraryConfiguration`
   ```csharp
   [Inject]
   internal LibraryConfiguration LibraryConfiguration { get; set; } = default!;
   ```
2. Add the `FormatCollocatedUrl(LibraryConfiguration)` extension method.
   ```csharp
   _jsModule = await JSRuntime.InvokeAsync<IJSObjectReference>("import",
                                      JAVASCRIPT_FILE.FormatCollocatedUrl(LibraryConfiguration));
   ```

## Unit Tests

Updated to inject the `LibraryConfiguration` service.

```csharp
Services.AddSingleton(LibraryConfiguration.ForUnitTests);
```

image


florianwachs:
this would be a great solution! thank you so much

vnbaaij (approved):

PR #2389, opened 2024-07-16
[DataGrid] Row Detail

franklupo:
Hi all, 
I implemented `HierarchyColumn ` like `SelectColumn`. 
It works if `Virtualkiza = False`.

image

image

I know there are no tests, but I want to understand if it makes sense to you


vnbaaij:
I think this could make sense, yes. Couple of questions/remarks (having only glanced at the code) 
- Does it work together with the `SelectColumn`?
- The name is not right. Hierarchy implies that it would be a kind of tree structure of rows, which it is not. Maybe `ExpandDetailColumn` or Just `DetailColumn` (don't change it yet, just thinking out loud here)
- The expand/collapse icons (chevrons in this case) should be configurable (parameters). Can imagine someone wanting to use '+'/'-' for example

franklupo:
> * Does it work together with the `SelectColumn`?

yes, they are two different columns

> * The expand/collapse icons (chevrons in this case) should be configurable (parameters). Can imagine someone wanting to use '+'/'-' for example

Yes, exists `IconExpanded ` and `IconCollapsed`.

franklupo:
> * The name is not right. Hierarchy implies that it would be a kind of tree structure of rows, which it is not. Maybe `ExpandDetailColumn` or Just `DetailColumn` (don't change it yet, just thinking out loud here)

IMO `ExpandDetailColumn` for Column and `ExpandDetailTemplate` for RenderFragment .

franklupo:
> > * The name is not right. Hierarchy implies that it would be a kind of tree structure of rows, which it is not. Maybe `ExpandDetailColumn` or Just `DetailColumn` (don't change it yet, just thinking out loud here)
> 
> IMO `ExpandDetailColumn` for Column and `ExpandDetailTemplate` for RenderFragment .

Have you flnd right name?

dvoituron:
This feature is more similar to the "Expand details".

I haven't found this feature in the FluentUI specification. 
We can integrate this column, but I think it's better to add a note to explain that it's not recommended to use it, and to prefer the display of detailed information for a line, in the form of a Dialog Panel (right).

> @vnbaaij It would be interesting to investigate (internally) whether or not this component should be added to the library.
> And why it doesn't yet exist in other FluentUI libs (UI, accessibility, ...)
> On the other hand, there's a "Grouping rows" feature that we don't have yet. 
> But it may conflict visually with this one. It would be preferable to use other icons [+] and [-] for this PR.
> 
> image


I also don't understand the example that uses `SelectedItems` when we're in a "Master/Details" (we'll probably have to change the name).
I would rename `CanBeExpanded` to `Expandable`

franklupo:
> I also don't understand the example that uses `SelectedItems` when we're in a "Master/Details" (we'll probably have to change the name). I would rename `CanBeExpanded` to `Expandable`
The example is now correct. I copied the SelectedColumn page. What is the name of the column?

franklupo:
why does the test fail?

dvoituron:
> why does the test fail?

You need to inject the missing service: `TestContext.Services.AddSingleton(LibraryConfiguration.ForUnitTests);`

franklupo:
> > why does the test fail?
> 
> You need to inject the missing service: `TestContext.Services.AddSingleton(LibraryConfiguration.ForUnitTests);`

I have no problems in visual studio

dvoituron:
> I have no problems in visual studio

If you updated your branch with the current `dev` branch, and running Unit Tests, you should have the same problem.

franklupo:
what is missing?

franklupo:
I thought. What if we did something to add content to the row?

in src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```c#
[Parameter]
public Func<TGridItem, bool>? RowHasAdditionalContent { get; set; }


[Parameter]
public RenderFragment<TGridItem>? RowAdditionalContentTemplate { get; set; }
```

in src/Core/Components/DataGrid/FluentDataGrid.razor
https://github.com/microsoft/fluentui-blazor/blob/921dc78c8fb9b5d2f10d094632f86ddba655d415/src/Core/Components/DataGrid/FluentDataGrid.razor#L111
```c#
@if (!Virtualize && RowAdditionalContentTemplate != null && RowHasAdditionalContent != null && RowHasAdditionalContent(item))
{
   @RowAdditionalContentTemplate (item)
}
```

The column could be an external component. More flexibility for other implementations.

What do you think?

franklupo:
For performance Is Better check null function or execute function and evaluate result?

vnbaaij:
> For performance Is Better check null function or execute function and evaluate result?

Better to check before, I think

franklupo:
I close this PR and created a new one for RowAdditionalContent

PR #2393, opened 2024-07-17
[Overflow] Add `Fixed` parameter

### LINKED ISSUE #2390: fix: popover doesn't close in FluentOverflow
# 🐛 Bug Report

This is a regression in FluentPopover when updating from 4.7.2 to 4.9.1. A FluentPopover that's nested inside a FluentOverflow isn't closing when clicking on the page outside of the popup.

## 💻 Repro or Code Sample

Code here:
https://github.com/dotnet/aspire/blob/1f94dac8188391804b22a8441c979d033b7b43b9/src/Aspire.Dashboard/Components/ResourcesGridColumns/EndpointsColumnDisplay.razor#L17-L60

Notably the popover is bound to a field, `@bind-Open="_popoverVisible"`.

I changed it to a property for debugging and found clicking outside of the popup set it to false in 4.7.2:
image

But in 4.9.2 that changed to true:
image

## 🤔 Expected Behavior

Clicking on the page outside of the popup will close the popup.

## 😯 Current Behavior

The popup flickers but stays visible:
fluentpopover-open

## 🌍 Your Environment

* OS & Device: Windows
* Browser: Chrome
* .NET 8

### LINKED ISSUE #2391: feat: add setting to always display the first item in FluentOverflow
# 🙋 Feature Request

FluentOverflow hides items if there aren't enough room to display them. However, sometimes you might want to display the first item, even if it is cutoff.

For example, given a list of names, seeing the first name, but having it cut off with an ellipsis (someone can add CSS to add ellipsis) so the user can interact with it is preferrable to just showing the overflow button and no content.

## 🤔 Expected Behavior

The first item is displayed is the space available, then the overflow button.

This is what it could look like:

fluentoverflow-firstitem

## 😯 Current Behavior

FluentOverflow only shows the first item if there is room for it. If there isn't then just the overflow button is displayed.

This is what it looks like today:

fluentoverflow-firstitem-current

## 💁 Possible Solution

Add a setting to FluentOverflow that controls whether the first item is always rendered inline.

vnbaaij:
- Fix #2390 by moving logic inside popover close code
- Fix #2391 by adding `Fixed` parameter to `FluentOverflowItem`

The script used to determine which items overflow actually already accounted for this...

By supplying a fixed attribute to an item, it will be excluded from the calculations (we already use that internally with the AppBar). We just did not expose this in any way in the FluentOverflowItem as a parameter.

## Before
aspire-before-fixed

## After
aspire-fixed


vnbaaij:
I've not done aby attempt at styling what is shown in the overflow/grid. Getting the effect as shown in your gif will probably be a bit more complex. I need to dive a bit deepee on how things are being rendered with this new set up.

dvoituron:
> I've not done aby attempt at styling what is shown in the overflow/grid. Getting the effect as shown in your gif will probably be a bit more complex. I need to dive a bit deepee on how things are being rendered with this new set up.

I can work on that tomorrow, based on a new PR.

dvoituron (approved):

JamesNK:
That looks great. I'll try it out!

PR #2397, opened 2024-07-17
[Search] Add AutoComplete parameter

### LINKED ISSUE #2394: FluentSearch AutoComplete
# 🙋 Feature Request

Can the AutoComplete property be added to the FluentSearch component so we can turn auto complete off when using these?

vnbaaij:
Follows the same method as used in TextField. Result:

search-autocomplete

Fix #2394

### DIFF
--- a/src/Core/Components/Search/FluentSearch.razor.cs
+++ b/src/Core/Components/Search/FluentSearch.razor.cs
@@ -63,6 +63,13 @@ public partial class FluentSearch : FluentInputBase<string?>
     public FluentInputAppearance Appearance { get; set; } = FluentInputAppearance.Outline;
 
     /// <summary>
+    /// Specifies whether a form or an input field should have autocomplete "on" or "off" or another value.
+    /// An Id value must be set to use this property.
+    /// </summary>
+    [Parameter]
+    public string? AutoComplete { get; set; }
+
+    /// <summary>
     /// Gets or sets the content to be rendered inside the component.
     /// </summary>
     [Parameter]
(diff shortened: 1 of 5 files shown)


src/Core/Components/Search/FluentSearch.razor.cs
```diff
@@ -79,11 +86,14 @@ protected override async Task OnAfterRenderAsync(bool firstRender)
...
         {
-            //if (!string.IsNullOrEmpty(Id))
-            //{
             Module ??= await JSRuntime.InvokeAsync<IJSObjectReference>("import", JAVASCRIPT_FILE.FormatCollocatedUrl(LibraryConfiguration));
             await Module.InvokeVoidAsync("addAriaHidden", Id);
-            //}
+
+            if (AutoComplete != null)
+            {
+                Module ??= await JSRuntime.InvokeAsync<IJSObjectReference>("import", JAVASCRIPT_FILE.FormatCollocatedUrl(LibraryConfiguration));
```
dvoituron, line 94: The `Module` is already created via the line 89.

src/Core/Components/Search/FluentSearch.razor.cs
```diff
@@ -62,6 +62,13 @@ public partial class FluentSearch : FluentInputBase<string?>
     [Parameter]
     public FluentInputAppearance Appearance { get; set; } = FluentInputAppearance.Outline;
 
+    /// <summary>
+    /// Specifies whether a form or an input field should have autocomplete "on" or "off" or another value.
```
dvoituron, line 66: Can you start the comment using `Gets or sets...`? To be coherent with the other properties?

dvoituron (approved):

Issue #2425, opened 2024-07-20
Fix For: Missing LibraryConfiguring Injection in DesignToken.razor.cs


TimonFeldmann:
# 🐛 Bug Report

This issue occurs exclusively in version 4.9.2, downgrading to 4.9.1 resolves the issue.

When injecting any design token and calling `.SetValue()` with any element reference, the `DesignToken` class fails to inject the registered `LibraryConfiguration` causing an `Object reference not set to an instance of an object`  exception to be thrown inside `UrlFormatterExtensions.cs`.

### Component:

```csharp
@using Microsoft.FluentUI.AspNetCore.Components.DesignTokens

<div class="container">
	<FluentTextField @ref="Test" @bind-Value=Title></FluentTextField>
</div>

@code {
	[Parameter, EditorRequired]
	public string Title { get; set; } = null!;

	FluentTextField Test;

	[Inject]
	private Density _density { get; set; } = default!;

	protected override async Task OnAfterRenderAsync(bool firstRender)
	{
		if (firstRender)
		{
			await _density.SetValueFor(Test.Element, 10);

			StateHasChanged();
		}
	}
}
```

### Startup:

```csharp
    public static class MauiProgram
    {
        public static MauiApp CreateMauiApp()
        {
            var builder = MauiApp.CreateBuilder();
            builder
                .UseMauiApp<App>()
                .ConfigureFonts(fonts =>
                {
                    fonts.AddFont("OpenSans-Regular.ttf", "OpenSansRegular");
                });

            builder.Services.AddMauiBlazorWebView();
	    builder.Services.AddHttpClient();
            builder.Services.AddFluentUIComponents();

#if DEBUG
            builder.Services.AddBlazorWebViewDeveloperTools();
            builder.Logging.AddDebug();
#endif

            return builder.Build();
        }
    }
```

## 🤔 Expected Behavior

Calling `.SetValue()` on any `DesignToken` should not throw a null reference exception.

## 😯 Current Behavior

`LibraryConfiguration` is injected as `null` into `DesignToken.razor.cs` causing a null reference exception when attempting to set *any* `DesignToken` value.

## 🌍 Your Environment

* OS & Device: Windows
* .NET MAUI desktop application
* .NET and Fluent UI Blazor library Version: .NET 8 / Fluent UI 4.9.2

TimonFeldmann:
[...]

PR #2430, opened 2024-07-21
Fix ToolbarDefault

### LINKED ISSUE #2428: Toolbar example preview error
In https://www.fluentui-blazor.net/Toolbar, the first example preview shows an error : 
`AggregateException_ctor_DefaultMessage (Arg_NullReferenceException)`

image

agriffard:
Attempt to fix #2428, by removing unknow classes.

Result:

image


vnbaaij (commented):
This is due to an error introduced with LibraryConfiguration. See #2425. We will check if a pr is still needed after we fix that error

PR #2448, opened 2024-07-24
[DataGrid] Row Additional Content

franklupo:
Ability to add additional content to DataGrid row, with check if add.
Original PR #2389


franklupo:
something is missing?

franklupo:
Are there any problems applying this change?

vnbaaij:
Hi Daniele, still not have time to take a proper look at this and won't be able for the coming weeks because of vacation period. There is no rush to get this in as it will require at least a minor release (i.e. 4.X)

franklupo:
you're right, this change gives me the ability to create components outside of this library.

vnbaaij:

PR #2451, opened 2024-07-24
[DataGrid] ColumnBase improve access information

franklupo:
To create columns outside this library I need information that isn't there now.


src/Core/Components/DataGrid/Columns/ColumnBase.razor.cs
```diff
@@ -128,6 +128,16 @@ public abstract partial class ColumnBase<TGridItem>
...
 
+    /// <summary>
+    /// Get items.
+    /// </summary>
+    protected ICollection<TGridItem> Items => InternalGridContext.Items;
+
+    /// <summary>
+    /// Get Total Item Count.
+    /// </summary>
+    public int TotalItemCount => InternalGridContext.TotalItemCount;
```
dvoituron, line 139: Probably better to have a `protected´ property.

src/Core/Components/DataGrid/Columns/ColumnBase.razor.cs
```diff
@@ -128,6 +128,16 @@ public abstract partial class ColumnBase<TGridItem>
     /// </summary>
     protected FluentDataGrid<TGridItem> Grid => InternalGridContext.Grid;
 
+    /// <summary>
+    /// Get items.
```
dvoituron, line 132: Can you give a different explanation of the property name?
`Gets a list of all elements included in the grid.`

dvoituron (changes requested):

dvoituron:
I think you can bypass these changes, adding this Cascading property in your new column code.
```csharp
[CascadingParameter]
internal InternalGridContext<TGridItem> InternalGridContext { get; set; } = default!;
```

src/Core/Components/DataGrid/Columns/ColumnBase.razor.cs
franklupo: the items are calculated they are not all

franklupo:
> I think you can bypass these changes, adding this Cascading property in your new column code.
> 
> ```cs
> [CascadingParameter]
> internal InternalGridContext<TGridItem> InternalGridContext { get; set; } = default!;
> ```

Ok but the InternalGridContext<TGridItem> Is internal

dvoituron:
> Ok but the InternalGridContext Is internal

Indeed ;-)

franklupo:
IMO InternalGridContext could be deleted and move properties/methods to FluentDataGrid. FluentDataGrid is found in CascadingParameter, or change visibility InternalGridContext

dvoituron:
> IMO InternalGridContext could be deleted and move properties/methods to FluentDataGrid. FluentDataGrid is found in CascadingParameter, or change visibility InternalGridContext

Nop. `InternalGridContext` exists to centralize some internal properties. The next major release could refactor this part to simplify the extensions.

vnbaaij:
> IMO InternalGridContext could be deleted and move properties/methods to FluentDataGrid. FluentDataGrid is found in CascadingParameter, or change visibility InternalGridContext

The `InternalGridContext` should be seen as the object holding the grid's internal state. We do _not_ want to expose that.

franklupo:
the items comes from here and it is not all the items

https://github.com/microsoft/fluentui-blazor/blob/cfb055ced298d26f83ebcb21ed31a636025db63c/src/Core/Components/DataGrid/FluentDataGrid.razor.cs#L512

src/Core/Components/DataGrid/Columns/ColumnBase.razor.cs
franklupo: the items comes from here and it is not all the items

https://github.com/microsoft/fluentui-blazor/blob/cfb055ced298d26f83ebcb21ed31a636025db63c/src/Core/Components/DataGrid/FluentDataGrid.razor.cs#L512

franklupo:
something is missing?

franklupo:
Are there any problems applying this change?

vnbaaij:
Hi Daniele, still not have time to take a proper look at this and won't be able for the coming weeks because of vacation period. There is no rush to get this in as it will require at least a minor release (i.e. 4.X)

franklupo:
you're right, this change gives me the ability to create components outside of this library.

vnbaaij:

Issue #2495, opened 2024-08-08
feat: add auto-fit to DataGrid


ksbecker:
# 🙋 Feature Request

Add some sort of auto-fit feature to the DataGrid component to make the columns render better

## 🤔 Expected Behavior

By setting a flag on the DataGrid component, the columns of the DataGrid should make an attempt to balance the columns better.

## 😯 Current Behavior

With no explicit formatting, the columns evenly distribute the width of each column.

## 💁 Possible Solution

Use JS and some CSS techniques to calculate column widths and apply widths dynamically to each column. 

## 🔦 Context

With dynamic content it's difficult to predict exact widths for each column of a DataGrid. This feature would automatically determine better widths for each column based on initial load.

## 💻 Examples

![image](https://github.com/user-attachments/assets/e2f3adee-a6ee-42d4-b6b7-3ef4824321df)

MarvinKlein1508:
I would also love to see the ability to not cut-off the text in the columns using this apporach here.

ksbecker:
> I would also love to see the ability to not cut-off the text in the columns using this apporach here.

Good news, I've already submitted a PR for this new feature. It's not perfect, but it's a simple way to format the columns for a bit better presentation out of the box.

vnbaaij:
Closing this as PR has been merged

PR #2496, opened 2024-08-08
[DataGrid] Add AutoFit feature to size columns automatically as good as possible

ksbecker:
# Pull Request

## 📖 Description

Introduced the AutoFit property to the FluentDataGrid component, enabling columns to auto-fit the grid width using JavaScript. Updated XML documentation and added a new demo section in DataGridPage.razor to showcase this feature. Created DataGridAutoFit.razor component for demonstration purposes.

Modified FluentDataGrid.razor and FluentDataGrid.razor.cs to implement AutoFit functionality, including invoking a JavaScript function to adjust column widths. Simplified SortByColumnAsync methods using ternary operators.

Updated CSS in FluentDataGrid.razor.css and FluentDataGridCell.razor.css to support AutoFit, ensuring proper styling and visibility of grid cells. Added autoFitGridColumns JavaScript function in FluentDataGrid.razor.js. Removed redundant CSS rules and adjusted formatting for consistency.

### 🎫 Issues

#2495 

## 👩‍💻 Reviewer Notes

Not much to this.

## 📑 Test Plan

This is mostly client side code.

### General

### Component-specific


dvoituron:
Can you explain how do you compute the column width?

And why do you use a JavaScript code to do that, and why not use the Flex `auto' property? One problem with using a JS is there are some side effects, like "the columns are not resized with the browser resize" or "the column sizes are recalculated when the user clicks on the col header to sort", ...

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -212,8 +213,21 @@ public partial class FluentDataGrid<TGridItem> : FluentComponentBase, IHandleEve
...
     /// A default fragment is used if loading content is not specified.
     /// </summary>
-    [Parameter]
-    public RenderFragment? LoadingContent { get; set; }
+    [Parameter] public RenderFragment? LoadingContent { get; set; }
+
+    /// <summary>
+    /// Uses JS to auto-fit the columns to the grid width.
+    /// </summary>
+    [Parameter] public bool AutoFit { get; set; }
```
dvoituron, line 221: Why did you add these lines (already includedin the file)

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -2,13 +2,14 @@
 // MIT License - Copyright (c) Microsoft Corporation. All rights reserved.
 // ------------------------------------------------------------------------
 
+using System.Diagnostics.CodeAnalysis;
```
dvoituron, line 5: Why did you move this using before other usings?

dvoituron (changes requested):

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -212,8 +213,21 @@ public partial class FluentDataGrid<TGridItem> : FluentComponentBase, IHandleEve
     /// Gets or sets the content to render when <see cref="Loading"/> is true.
     /// A default fragment is used if loading content is not specified.
     /// </summary>
-    [Parameter]
-    public RenderFragment? LoadingContent { get; set; }
+    [Parameter] public RenderFragment? LoadingContent { get; set; }
+
+    /// <summary>
+    /// Uses JS to auto-fit the columns to the grid width.
```
dvoituron, line 219: Can you give a more explicit (and less technical: no `JS`) documentation to explain what `auto-fit` means?

dvoituron (changes requested):

ksbecker:
> Can you explain how do you compute the column width?
> 
> And why do you use a JavaScript code to do that, and why not use the Flex `auto' property? One problem with using a JS is there are some side effects, like "the columns are not resized with the browser resize" or "the column sizes are recalculated when the user clicks on the col header to sort", ...

My skill in CSS is limited, so I don't know much about Flex or other CSS abilities which is why I didn't go that route. Also, there is the `grid-template-columns` attribute, which I've never heard of, so wasn't sure how that works.

Here's how my solution works:

I start by overriding some CSS with the `.auto-fit` class on the data grid that disables the `overflow` and `text-overflow` properties of the cells. I also set the `visibility` to `hidden` to temporarily hide the ugliness of the process. This is a technique I've used in the past (Internet Explorer days). It makes the cells invisible but takes up the space it needs to render the content.

Here's the JS pseudo code:
* Get the available width for the grid
* Loop through the columns and find (`flatMap`) the biggest width of each cell of the column
* Add up those values to see how much room the cell widths exceed the overall available width
* Calculate the ratio and apply said ratio to each column width while building the `grid-template-columns` string

Scenario:
1. Grid width: 800px
2. Column widths (needed to render full content): 100px, 100px, 50px, 750px = 1000px
3. Ratio is 800/1000 = 0.8
4. New column widths: 80px, 80px, 40px, 600px = 800px

For me, this was a simple way to make that first render better. This could be broadened to execute on resize pretty easily. I don't think it's too necessary on sort/pagination as the content typically would be similar in length. This also does not interfere with column resizing by the UI. It simply sets the widths, and the user can resize columns as they see fit. 

I implemented this in my project and thought the community could start with this and make it better. If you, or anyone else, has suggestions, I'm happy to learn and would apply changes that make it better.

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
ksbecker: Sorry, merge conflict and I missed this.

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
ksbecker: My Cleanup Profile automatically does this.

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
ksbecker: I updated the summary in the code. Not sure if that's what you are looking for?

dvoituron:
That's interesting, I think you should add this explanation above the examples, to explain how these column widths are calculated:
> 1. Get the available width for the grid
> 2. Loop through the columns and find the biggest width of each cell of the column
> 3. Add up those values to see how much room the cell widths exceed the overall available width
> 4. Calculate the ratio and apply said ratio to each column width while building the `GridTemplateColumns` string

Next, you need to find a way to avoid "duplicating the configuration": you should only use `AutoFit="true"` but not force the developer to define `GridTemplateColumns`. When using `AutoFit`, `GridTemplateColumns` should be ignored.
```xml
<FluentDataGrid Items="@people" GridTemplateColumns="auto-fit" AutoFit="true">
```

I also wonder how this "calculation with cells temporarily rendered invisible" functionality will behave on platforms such as MAUI, or with the **Virtualization** feature. Could you check these two cases?

On the other hand, there's the [fit-content](https://developer.mozilla.org/en-US/docs/Web/CSS/fit-content_function) attrribute which might help. But I don't think it would be a viable solution.

dvoituron:
Other points.
1. Can you round off the calculated value to avoid unnecessary precision? Which is copied on all lines.
    `grid-template-columns="160.18437118437117px 174.6892551892552px 180.9957264957265px 517.1306471306472px`
  => `grid-template-columns="160px 174px 180px 517px`

2. Why not use the `fr` unit instead of `px`? This allows the display to be calculated proportionally when the page is loaded, but adapted if the user resizes the screen.
     `grid-template-columns="160fr 174fr 180fr 517fr`
     
peek_4

ksbecker:
> That's interesting, I think you should add this explanation above the examples, to explain how these column widths are calculated:
> 
> > 1. Get the available width for the grid
> > 2. Loop through the columns and find the biggest width of each cell of the column
> > 3. Add up those values to see how much room the cell widths exceed the overall available width
> > 4. Calculate the ratio and apply said ratio to each column width while building the `GridTemplateColumns` string
> 
> Next, you need to find a way to avoid "duplicating the configuration": you should only use `AutoFit="true"` but not force the developer to define `GridTemplateColumns`. When using `AutoFit`, `GridTemplateColumns` should be ignored.
> 
> ```
> <FluentDataGrid Items="@people" GridTemplateColumns="auto-fit" AutoFit="true">
> ```
> 
> I also wonder how this "calculation with cells temporarily rendered invisible" functionality will behave on platforms such as MAUI, or with the **Virtualization** feature. Could you check these two cases?
> 
> On the other hand, there's the [fit-content](https://developer.mozilla.org/en-US/docs/Web/CSS/fit-content_function) attrribute which might help. But I don't think it would be a viable solution.

I tested with virtualization and, no surprise, it doesn't work there. `OnAfterRenderAsync` is called before the data is rendered. Is there an event/callback after the virtualized rows are rendered? I see that `RenderRow` is called for each row but don't see where we can know when all rows are done rendering. Maybe disallow auto-fit when `Virtualized` is turned on?

I've never done a MAUI project so I can't test quickly there. If I have time I can try. Maybe for now just note that it's untested in MAUI?

One thing I noticed with `grid-template-columns` is that its use here seems atypical. The W3Schools site has a more basic implementation so it doesn't line up with what's going on with `FluentDataGrid`. This auto-fit wouldn't be necessary (and may not be) if the implementation was more basic. However, this component isn't basic, so I understand that it may have evolved beyond the original CSS design.

ksbecker:
> Other points.
> 
> 1. Can you round off the calculated value to avoid unnecessary precision? Which is copied on all lines.
>    `grid-template-columns="160.18437118437117px 174.6892551892552px 180.9957264957265px 517.1306471306472px`
>    => `grid-template-columns="160px 174px 180px 517px`
> 2. Why not use the `fr` unit instead of `px`? This allows the display to be calculated proportionally when the page is loaded, but adapted if the user resizes the screen.
>    `grid-template-columns="160fr 174fr 180fr 517fr`

As I was addressing this, I realized that there was no need to calculate the ratio and all that mess. Basically the `fr` unit will do that for the browser. So, I reduced it down to loop and build the string using the full max width of each column and using the `fr` unit instead. I refactored the `.js` code to make it cleaner.

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -212,8 +213,17 @@ public partial class FluentDataGrid<TGridItem> : FluentComponentBase, IHandleEve
...
+
+    /// <summary>
+    /// Sets <see cref="GridTemplateColumns"/> to automatically fit the columns to the available width as best it can.
+    /// </summary>
+    [Parameter] public bool AutoFit { get; set; }
+
+    /// <summary>
+    /// Gets the first (optional) SelectColumn
+    /// </summary>
+    internal IEnumerable<SelectColumn<TGridItem>> SelectColumns => _columns.Where(col => col is SelectColumn<TGridItem>).Cast<SelectColumn<TGridItem>>();
```
vnbaaij, line 226: Why is this here? I don't see it used in your code. Please remove if you are not using it.

src/Core/Components/DataGrid/FluentDataGrid.razor.js
```diff
@@ -159,7 +158,7 @@ export function enableColumnResizing(gridElement) {
 }
 
 export function resetColumnWidths(gridElement) {
-
+
```
vnbaaij, line 161: whitespace/blank line not needed

vnbaaij (changes requested):

src/Core/Components/DataGrid/FluentDataGrid.razor.js
vnbaaij: I did not add anything in 😀

vnbaaij (approved):

PR #2510, opened 2024-08-10
[Menu] Add a FluentMenuProvider

dvoituron:
# [FluentMenu] Add a FluentMenuProvider

With this PR, the `FluentMenu` component has been updated to use the `FluentMenuProvider`.

The `<FluentMenuProvider />` is a new component to be placed at the bottom of your HTML page (like the other **Providers**), which draws all menus to this location. This allows menus to be drawn **on top** of other components.

You can disable this feature by adding the `UseMenuService="false"` attribute. In this case, the menu is drawn at the location of your HTML code.

## Example

The **first** example contains the `UseMenuService="false"` attribute. The menu is created where it is declared, which can pose encapsulation problems, depending on the container used (in this case, a `FluentCard`).

The **second** example uses the `FluentMenuProvider` to display the menu. In this case, the menu is generated at the end of the HTML page and attached to the button using the `Anchor` tag. It's mandatory to use a `Id` identifier on the clickable element and an identical `Anchor` value on the menu, to link the two elements.

> By default (without changes), all menus will use this new feature.

```xml
<p>Click this button to open a Menu.</p>
<FluentButton Id="btnOpenMenuProvider1" Appearance="Appearance.Accent" OnClick="@(() => open1 = !open1)">
    Open menu
</FluentButton>

<!-- UseMenuService="false" or "true" (default) -->
<FluentMenu UseMenuService="true" Anchor="btnOpenMenuProvider1" @bind-Open="open1">
    <FluentMenuItem>Menu item 1</FluentMenuItem>
    <FluentMenuItem>Menu item 2</FluentMenuItem>
</FluentMenu>
```

peek_3

## Unit Tests

No change


vnbaaij (approved):

vnbaaij (changes requested):
See comments

vnbaaij (approved):

Issue #2549, opened 2024-08-18
RawEventHandlers fluent-menu-item has error


zandiarash:
There is an error in **RawEventHandlers.razor** file which prevents to run.
![bug](https://github.com/user-attachments/assets/eb33b277-791f-493f-8ffb-7a49d9ad1cef)

dvoituron:
I don't understand why you're getting an error on this attribute. `role` is a `string` attribute, so you can place any string you like.

On the other hand, we don't have this error with the latest version of Visual Studio Preview.
Which version are you using? Which version of .NET?

zandiarash:
> I don't understand why you're getting an error on this attribute. `role` is a `string` attribute, so you can place any string you like.
> 
> On the other hand, we don't have this error with the latest version of Visual Studio Preview.
> Which version are you using? Which version of .NET?

I am using Dotnet 9 and Visual studio the latest preview version.

zandiarash:
> I don't understand why you're getting an error on this attribute. `role` is a `string` attribute, so you can place any string you like.
> 
> On the other hand, we don't have this error with the latest version of Visual Studio Preview. Which version are you using? Which version of .NET?

I cloned again the **dev** branch it has this error yet
![image](https://github.com/user-attachments/assets/6957c6fb-0c40-4123-b394-88bd6429f3c0)

dvoituron:
Can you try to build using DotNet8 (and not 9)?

zandiarash:
> Can you try to build using DotNet8 (and not 9)?

I tried **dotnet build --framework net8.0** error still exists.

vnbaaij:
This is caused by an error in your environment, not by an issue in the library/code. Another thing you can try is to add a `global.json` file to the root of your repo folder containing the following:

```
{
  "sdk": {
    "version": "8.0.302",
    "rollForward": "major"
  }
}
```

Closing this as it can't be solved from our side.

PR #2550, opened 2024-08-18
RawEventHandlers fluent-menu-item error

zandiarash:
[This ](https://github.com/microsoft/fluentui-blazor/issues/2549)closes.


dvoituron:
See https://github.com/microsoft/fluentui-blazor/issues/2549#issuecomment-2295272415

vnbaaij:
As @dvoituron said in the comments in the issue. The value needs to be a string. Replacing that with an enum value is not the correct fix for this. Closing this PR for that reason.
I'll take a closer look if the file in question is actually still needed.

PR #2579, opened 2024-08-24
[Menu] Pass through more parameters to region the menu is anchored to

vnbaaij:
Passing `HorizontalViewportLock` and `HorizontalScaling` through to the anchored region so a DataGrid header menu can be set to occupy the full header button width:

image


src/Core/Components/Menu/FluentMenu.razor.cs
```diff
@@ -161,6 +161,18 @@ public bool Open
     [Parameter]
     public int HorizontalThreshold { get; set; } = 200;
 
+    /// <summary>
```
dvoituron, line 164: You need to add these properties to the `FluentMenuProvider.razor` file.

dvoituron (changes requested):

src/Core/Components/Menu/FluentMenu.razor.cs
vnbaaij: Done

dvoituron (approved):

PR #2567, opened 2024-08-22
[DatePicker] Added OnDoubleClick event and DoubleClickToDate parameter

RickyLin:
# Pull Request

## 📖 Description

1. Added `OnDoubleClick` event to DatePicker component.
2. Added `DoubleClickToDate` parameter to DatePicker component.
3. Updated DatePickerDefault page in demo site to add `DoubleClickToDate` option

### 🎫 Issues

The PR was related to
1. The discussion #2524 How can I extend FluentDatePicker to support double-click event on its text field?
2. The PR #2563 [DatePicker] Added OnDoubleClick event and DoubleClickToToday parameter.

## 👩‍💻 Reviewer Notes

This is the first PR of the 2 separated PRs asked by @dvoituron which implemented the `DoubleClickToDate` feature only.

## 📑 Test Plan

This PR kept the `FluentOverlay` as-is, so it's very difficult to trigger the double-click event on the text field of the date picker, thus difficult to see this feature works. Commenting out the `FluentOverlay` temporarily could help a lot during the manual tests.

However, the `FluentOverlay` has no side-effect when running the test method `FluentDatePicker_DoubleClickToSetTodayInTextField`, so it runs successfully. I think that's a sample of the cases that automatic tests are not exactly the same as manual tests.

### General

- [x] I have added tests for my changes.
- [x] I have tested my changes.
- [x] I have updated the project documentation to reflect my changes.
- [x] I have read the [CONTRIBUTING](https://github.com/microsoft/fluentui-blazor/blob/master/docs/contributing.md) documentation and followed the [standards](https://www.fast.design/docs/community/code-of-conduct/#our-standards) for this project.

### Component-specific

- [ ] I have added a new component
- [ ] I have added [Unit Tests](https://github.com/Microsoft/fluentui-blazor/blob/master/unit-tests.md) for my new component
- [x] I have modified an existing component
- [x] I have validated the [Unit Tests](https://github.com/Microsoft/fluentui-blazor/blob/master/unit-tests.md) for an existing component

## ⏭ Next Steps

I'll move the rest "`FluentOverlay` replacement" part of the original PR #2563 to a new PR.

PR #2580, opened 2024-08-25
[Overlay] Add Interactive and InteractiveExceptId parameters

dvoituron:
# [Overlay] Add FullScreenInteractiveExceptElementId

## Introduction

In some use cases (see [#2567#issuecomment-2307956048](https://github.com/microsoft/fluentui-blazor/pull/2567#issuecomment-2307956048), it's useful to allow the user to interact with components placed under the **FluentOverlay** (with the possible exception of certain zones). To manage this, we've added the `pointer-events: none` style, which disables event handling on the **FluentOverlay** and allows interaction with other elements on the web page. It is then necessary to add a global event `document.addEventListener` via the `FluentOverlay.razor.js` script (register it and remove it at the end of use).

## Stay interactive
By using the `InteractiveExceptId` property, only the targeted element will not close the **FluentOverlay** panel. The user can click anywhere else to close the **FluentOverlay**.

In this example, the **FluentOverlay** will only close when the user clicks outside the white zone and the user can increment the counter before to close the Overlay

## Simplified example

```xml
<FluentOverlay @bind-Visible=@visible
               FullScreen="true"
               Interactive="true"
               InteractiveExceptId="my-zone">
  <div id="my-zone">
    <p>Non-interactive zone</p>
    <FluentProgressRing />
  </div>
</FluentOverlay>
```

peek_3


RickyLin:
The `FullScreenInteractiveExceptElementId` property implies two things:
1. The overlay is full screen interactive
2. Except for an element id

Is there a reasonable case that someone wants a full screen interactive overlay and no exceptions? If yes, then they can't achieve it with this property and probably we need a separate `FullScreenInteractive` property which is of type `bool` and current `FullScreenInteractiveExceptElementId` is for the exception element id only, null or empty means no exceptions.

Actually, there's a `FullScreen` property, so perhaps just have two more properties: `Interactive` of type `bool` and `InteractiveExceptElementId` of type `string?`, three properties, each one has simple and dedicated purpose.

dvoituron:
If the dev want no exception, he can set an unknown ID. This ElementId will never found to have the expected result. 

We don't check if the ID exists in the HTML DOM, we try to find it.

RickyLin:
I would say it's not straightforward and more like a workaround, but it does the trick.

vnbaaij:
I think the functionality this offers is good but the name of the parameter is a bit verbose and unclear. Maybe change to something like `DissmisExemptId`?

dvoituron:
> I would say it's not straightforward and more like a workaround, but it does the trick.

I agree. I could add a paragraph in the documentation to explain that. And to add a global constant `NoException = Identifier.NewId()` to be more explicit for the dev.

dvoituron:
> I think the functionality this offers is good but the name of the parameter is a bit verbose and unclear. Maybe change to something like `DissmisExemptId`? 

I was waiting for you on the name 😀. I couldn't think of a simple name. Your proposal is easier but remove the info of Fullscreen and Interactivity (maybe not interesting). 

Other proposals ? `InteractiveExceptId`, ...

vnbaaij:
> I was waiting for you on the name 😀. I couldn't think of a simple name.

😂 I think the fact the name says you exempt an ID from dismissing (I misspelled earlier) the overlay kind of implies interactivity.
The full screen aspect can be clarified in xnl and developer docs

RickyLin:
I still think it may be helpful to separate three concepts:
1. Full Screen
2. Interactive
3. Interactive Exception Element Id

Separating 1 and 2 can make the interactive available for non-full-screen as well if it could be a potential scenario.
Separating 2 and 3 can be intuitive for people who wants the interactive overlay only, simply set `Interactive` to `true` other than an `NoException` to `InteractiveExceptionElementId` because the `Interactive` emphasize interactive, can be true/false, the `InteractiveExceptId` emphasize an element id, set it a special `NoException` is not very intuitive for interactive only even if it's elaborated in document, unless the name is Interactive**But**ExceptId, a longer name but emphasize both interactive and except id.

vnbaaij:
@RickyLin we can't separate the interactivity from the full screen part because it relies on a `addEventListener` on the document on the JS side.
Denis _will_ spilt out the `Interactive` and the `InteractivityExcludedElement` concepts.

dvoituron:
Yep, Probably using `Interactive = true` and `InteractiveExceptId = "my-id"` will be easier for the dev (without this ID, the entire page will be interactive).

Of course, these properties will be dependent: `InteractiveExceptId` will be ignored if `Interactive = false`; and using `Interactive` will automatically set `FullScreen = true` .

Make sense?

vnbaaij:
Yes, makes sense to me!

RickyLin:
It's almost perfect to me. One last thought, can it support non-full-screen scenario? In that case, shall we look for the container element that the `FluentOverlay` covers then add event listener to that container element instead of the document element? I know it may be a rare case, but it would make the feature perfect if it won't take much effort.

RickyLin:
It's nearly perfect for me. One last thought, for non-full-screen scenario, if we add event listener to the container element of the FluentOverlay based on its absolute position (nearest non-static positioned ancestor element?) instead of the document element, then it makes the same semantic for the full screen FluentOverlay and can decouple the `FullScreen` property.

dvoituron:
> One last thought, for non-full-screen scenario, if we add event listener to the container element

Mmm. Let me try if it's possible (but using the FluentOverlay element, not using the absolute position).

dvoituron:
PR fixed (and description updated)

vnbaaij (approved):

PR #2585, opened 2024-08-26
[DataGrid] Add parameter to provide labels to resize UI

### LINKED ISSUE #2507: fix: buttons in ColumnResizeOptions are inaccessible
# 🐛 Bug Report

Please expose parameters we can pass to `ColumnResizeOptions` to make sure that the commit edit and reset buttons have aria labels.

## 💻 Repro or Code Sample

https://github.com/microsoft/fluentui-blazor/blob/7b747797a865ff46ee0f7a41771eef21520eb750/src/Core/Components/DataGrid/Columns/ColumnResizeOptions.razor#L29

These buttons do not have aria labels nor can we add them ourselves. This is an accessibility concern and prevent us from being able to use this.

## 🤔 Expected Behavior

NA, see above

## 😯 Current Behavior

NA, see above

## 💁 Possible Solution

My preferred solution would be to add settable parameters to `ColumnBase` and `FluentDataGrid` for `CommitButtonLabel` and `ResetButtonLabel` with defaults. If there is a desire to avoid cluttering `FluentDataGrid` with additional parameters, `ShowColumnOptionsAsync` could be modified to accept these parameters, which would then be set on `ColumnBase`. The downside of this approach would be that you would have to implement a header template yourself.   

## 🌍 Your Environment

vnbaaij:
Fix #2507 by adding a new `ColumnResizeLabels` record which contains the following properties:
- `DiscreteLabel`
- `ExactLabel`  
- `GrowAriaLabel`
- `ShrinkAriaLabel`
- `ResetAriaLabel` 
- `SubmitAriaLabel` 
 
All properties are assigned default values. A parameter of this type with name `ColumnResizeLabels` is added to DataGid. This can be used to supply custom strings for the mentioned properties which will then show up in the resize options UI

**NOTE**  the DataGrid parameter `ResizeLabel` has been superseded by this new parameter. If you were using a custom text for this, then change this to use the `DiscreteLabel` and/or `ExactLabel`  (depending on the `ResizeType` you are using)
 
image


src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -100,9 +100,41 @@ public partial class FluentDataGrid<TGridItem> : FluentComponentBase, IHandleEve
...
     public DataGridResizeType? ResizeType { get; set; }
 
+    /// <summary>
+    /// Label used for the text field in the resize options dialog.
+    /// Defaults to "Column width (in pixels)".
+    /// </summary>
     [Parameter]
     public string ResizeLabel { get; set; } = "Column width (in pixels)";
 
+    /// <summary>
```
dvoituron, line 110: To avoid these multiple parameters, and group all this information in a single object, how about using something like this :

```csharp
public record ResizeAria
{
    public string? GrowLabel { get; set; }
    public string? ShrinkLabel { get; set; }
    public string? ResetLabel { get; set; }
    public string? SubmitLabel { get; set; }

    public static ResizeAria Default => new ResizeAria();
}
```

The usage will be:

```xml
<FluentDataGrid ResizeAria="@(ResizeAria.Default with { GrowLabel = "...", ShrinkLabel = "..." } )">
```

Do you find it too complex ?

dvoituron (dismissed):

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
vnbaaij: I was thinking about something like that as well (but then with a dictionary of some sorts)

vnbaaij: A record has been added. See updates in PR message above

dvoituron: 1. I find the syntax for defining the record clearer using property declarations (as above). And using your syntax, we don't have XML comments / property documentation.

2. Why did you put this class in the `Infrastructure` folder and not in the `DataGrid/Columns` folder?

3. I think we need a static property `Default` to create an instance of this object. Thus, the developer can initialize the parameter using `ResizeAria.Default with { GrowLabel = “ ...”, ShrinkLabel = “ ...” }`

4. ⚠️ You can't force RPs to merge: without explicit approval, we can continue discussing.

vnbaaij: 1) By doing it this way, the developer does not have to do anything to use the default values.  I think we can put the comments about the fields of the record in the parameter declaration in the Grid
2) Yes, that is a mistake. Will move them
3) ~~I could not get your variant to work with implemented defaults. I don't want to force the developer to **have to** supply values in the default case~~ I figured it out. Will change it in the other PR
4) I know but I wanted this in because (and was too impatient). I need it in the other grid thing I'm working on (PR inbound) (mea culpa). We can still discuss and change because we are the end bosses.

PR #2586, opened 2024-08-27
[DataGrid] Provide new way to render column actions

vnbaaij:
**For now**, this new behavior is 'opt-in' and can be enabled by adding `HeaderCellAsButtonWithMenu="true"` in your data grid component tag.

Following the work that was done in Aspire, we are adding a new way to render column actions (sort, resize and options (filter)) by using a menu shown when clicking on a column header:

image

As can be seen in the image, the current sort direction is displayed on the menu item. The sort direction and filter state are shown on the header with icons.

Which items are displayed on a menu is dependent on the parameter settings on the grid. So when a column is not sortable, its menu won't show the 'Sort' menu item. If there is also no column options/filter specified, it will directly show the Resize UI (in this case because `ResizeType.Discrete` has been set on the grid):
image

When a DataGrid is not resizable and a column does not have any options ui (no filter) applied, the header button still functions as a sort button directly:
newcolumnheaderuinotresizable

The 'Bronze' column is not sortable and therefore shows a disabled cursor (not shown in the screengrab)

This PR also contains changes discussed in PR #2585. There a now records for resize, sort and options which use default values when a developer does not make any changes. `DataGridTypical.razor` contains an example of how the labels can be changed.


src/Core/Components/DataGrid/Columns/ColumnBase.razor.cs
```diff
@@ -218,4 +223,65 @@ public ColumnBase()
...
+        if (args.Key == "Enter" || args.Key == "NumpadEnter")
+        {
+            await Grid.ShowColumnResizeAsync(this);
+            _isMenuOpen = false;
+        }
+    }
+
+    private async Task HandleOptionsMenuKeyDownAsync(KeyboardEventArgs args)
+    {
+        if (args.Key == "Enter" || args.Key == "NumpadEnter")
```
dvoituron, line 264: Could you use a "constant" like this?
```csharp
private static string[] KEYBOARD_MENU_SELECT_KEYS = ["Enter", "NumpadEnter"];
```

src/Core/Components/DataGrid/Columns/ColumnBase.razor.cs
```diff
@@ -128,6 +128,11 @@ public abstract partial class ColumnBase<TGridItem>
     /// </summary>
     protected FluentDataGrid<TGridItem> Grid => InternalGridContext.Grid;
 
+    protected bool AnyColumnActionEnabled => Sortable is true || IsDefaultSortColumn || ColumnOptions != null || Grid.ResizableColumns;
+
+    private bool _isMenuOpen;
```
dvoituron, line 133: Can you put the "private" variable in the top of the file?

dvoituron (changes requested):
Great work! 

About the menu display, is it possible to avoid a menu that's too big (as can be seen in your demo)?

vnbaaij:
> Great work!
> 
> About the menu display, is it possible to avoid a menu that's too big (as can be seen in your demo)?

Yes. I was experimenting with some settings. Changing that back to default which will get you:
image

dvoituron (approved):

PR #2590, opened 2024-08-28
[General] Add .net9.0 as targeted framework

vnbaaij:
Also:
- Update NuGet packages / .NET SDK
- Update `contributing.md` and `using-lates-daily.md`
- Update publish workflows to get .NET 9 version deployed

### DIFF
--- a/.github/workflows/deploy_preview.yml
+++ b/.github/workflows/deploy_preview.yml
@@ -17,6 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Build and deploy Demo site
     env:
+      DOTNET_VERSION: "net9.0"
       DOTNET_CLI_TELEMETRY_OPTOUT: 1
       DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
       DOTNET_NOLOGO: true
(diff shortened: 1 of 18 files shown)


.github/workflows/deploy_preview.yml
```diff
@@ -30,20 +30,20 @@ jobs:
...
-     
-      - name: .NET Builld 
-        run: dotnet build ${{ env.PROJECT }} -c ${{ env.BUILD_CONFIG }} -o publish -f net8.0 -r linux-x64 --self-contained=true -p:BuildNumber=$BUILD_NUMBER -p:SourceRevisionId=$GITHUB_SHA -p:ContinuousIntegrationBuild=true
-      
-      - name: .NET Publish 
-        run: dotnet publish ${{ env.PROJECT }} -c ${{ env.BUILD_CONFIG }} -o publish -f net8.0 -r linux-x64 --self-contained=true -p:BuildNumber=$BUILD_NUMBER -p:SourceRevisionId=$GITHUB_SHA -p:ContinuousIntegrationBuild=true
-       
+            9.0.x
+
+      #- name: .NET Builld
```
dvoituron, line 41: Although the Publish will do the Build if it doesn't exist. I think it's better to keep a Build and then make a Publish (the bin will already be present, so it won't do this compilation again).

dvoituron (changes requested):
Why do you want to target .net9 in v4? As far as I'm concerned, only v5 will support .net9.

vnbaaij:
> Why do you want to target .net9 in v4? As far as I'm concerned, only v5 will support .net9.

v5 will **only** support .NET 9. 

For v4 we add this so devs can choose which .NET version they want to use. We are not adding or supporting any .NET 9 specific (Blazor) enhancements. Just compiling for it and including the DLL in the package. If they choose to use .NET 9, they will profit from the general Blazor enhancements (faster WASM start up, Server reconnect changes, etc) in the most optimal way.

dvoituron (approved):

PR #2438, opened 2024-07-23
FluentSlider: by using web components current-value instead of value

LuohuaRain:
# Pull Request

- By using web components current-value instead of value
- Update Tests and fixed a typo

### General

- [ ] I have added tests for my changes.
- [x] I have tested my changes.
- [ ] I have updated the project documentation to reflect my changes.
- [x] I have read the [CONTRIBUTING](https://github.com/microsoft/fluentui-blazor/blob/master/docs/contributing.md) documentation and followed the [standards](https://www.fast.design/docs/community/code-of-conduct/#our-standards) for this project.

### Component-specific

- [ ] I have added a new component
- [ ] I have added [Unit Tests](https://github.com/Microsoft/fluentui-blazor/blob/master/unit-tests.md) for my new component
- [] I have modified an existing component
- [x] I have validated the [Unit Tests](https://github.com/Microsoft/fluentui-blazor/blob/master/unit-tests.md) for an existing component

PR #2591, opened 2024-08-28
Revert "[FluentSlider] Use current-value instead of value attribute in web component"

vnbaaij:
Reverts microsoft/fluentui-blazor#2438

When using `current-value` instead of `value`, on Blazor Server the Slider will get stuck in an update loop when moving the tick around quickly. I've tested which other effects there could be with using `value` and haven't found any.


tests/Core/Slider/FluentSliderTests.razor
```diff
@@ -77,19 +77,4 @@
         // Assert
         cut.Verify();
     }
-
-    [Fact]
-    public void FluentSlider_Binding()
```
dvoituron, line 82: Why did you removed this test ?

dvoituron (changes requested):

tests/Core/Slider/FluentSliderTests.razor
vnbaaij: That was because of the revert action. I'll move this back in

dvoituron (approved):

LuohuaRain:
So, how can I change the value of the FluentSlider externally, rather than just setting its initial value?

Let me give an example: I want to change the value of the FluentSilder using the mouse wheel, and also be able to set the value by dragging the FluentSlider. Currently, after dragging it once, the pointer on the FluentSlider can no longer be changed with the mouse wheel.

Issue #1508, opened 2024-02-13
feat: allow cancelation of panel/dialog closing


rfrancisco:
# 🙋 Feature Request

Panels and Dialogs can be closed by explicit user actions and they can also be closed by pressing escape, by clicking on the backdrop, or the close button on the top-right corner. I believe it's a fairly common scenario to have forms inside these controls, and in these cases, it's useful to show a message when the user tries to close the panel/dialog and has made changes to the form.
Currently there's no way to intercept the close action to validate if the close action is allowed or not.

## 🤔 Expected Behavior

Be able to intercept close request and perform logic to determine if the closing the panel/dialog is allowed or not.

## 💁 Possible Solution

Allow setting a custom function that is called, instead of the default one, when the user tries to close the panel/dialog by clicking the close button, pressing escape, or clicking on the backdrop. Alternatively, it would also work to have a callback function that intercepts the close and allows the developer to implement custom logic (for example show a confirmation messagebox) to determine if the close should proceed or be canceled.

## 🔦 Context

Show a panel with forms. If the user tries to close the form with unsaved changes be able to request confirmation from the user and cancel the close event if the user decides not to lose the changes made.

vnbaaij:
You should be able to do what you want by hooking into the `OnDialogClosing` event callback.

rfrancisco:
> You should be able to do what you want by hooking into the `OnDialogClosing` event callback.

I don't see a way to cancel the closing using OnDialogClosing callback. I looked for usages of OnDialogClosing on the source code but could only find an example of using it to display animations. Is there an example on how it can be used to cancel the operation?

vnbaaij:
> @vnbaaij I don't see a way to cancel the closing using OnDialogClosing callback. I looked for usages of OnDialogClosing on the source code but could only find an example of using it to display animations. Is there an example on how it can be used to cancel the operation?

No, we don;t have an example of that
 
> Also, and I might be getting it wrong, it looks like this event is useful to run code from outside the dialog instance. In the case of dialogs with forms, the decision to prevent the close is made on the dialog itself because he is the one who knows if the form is dirty.

Ok, you might be right there. 

[...]

PR #2614, opened 2024-09-03
[Dialog] Add event before closing panel to allow validation of the data provided by the user

diegomodolo:
…ta. (#1508)

# Pull Request

## 📖 Description

I've added functionality to prevent the panel from closing if rules defined by the user don't pass.
I've added a func<bool> that will be called when the user clicks the button to commit the changes (won't be called on cancel or dismiss), and that func will return true or false, accordingly to the user validation rules.
image

### 🎫 Issues
#1508 

### General

### Component-specific


diegomodolo:
@microsoft-github-policy-service agree [company="{Nectar Sistemas}"]




[cid:c7339ad7-a11d-4ad0-a568-280eab5c3581]
________________________________

diegomodolo:
@microsoft-github-policy-service agree company="Nectar Sistemas"

[cid:c0a63e77-4275-4add-9ea1-e279fdc54b86]
________________________________

vnbaaij:
Nice work! We need to take a closer look of course, but on first quick glance this look good and simple.

We are on the brink of releasing a new version. If we approve and merge, this will go into the next one then.

dvoituron:
Interesting. 

Why not use `OnClosing`? Possibly by adding a `Cancelled` parameter. 

I also find the `On` suffix indicates that this is an `EventCallback`, which is not the case.

diegomodolo:
> Interesting.
> 
> Why not use `OnClosing`? Possibly by adding a `Cancelled` parameter.
> 
> I also find the `On` suffix indicates that this is an `EventCallback`, which is not the case.

I thought about the parameter, but I would have to add it to DialogInstance, and I thought it would be more confusing since the property would not relate to the context of the class.

About the On prefix, I can change it, no problem.

dvoituron:
You could add this parameter to a inherited class of DialogInstance (like `DialogInstanceClosing`). 

My question is more: why to have similar "events". A
`Func<bool>` like you did can solve this problem, but using a `Closing` event can be more intuitive, no? 🙂

dvoituron:
My reflection is more like used by the old (but excellent) **WinForms** 😁

https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.form.formclosing

dvoituron:
> I will abandon this PR, make the changes and create a new one.

No keep opened this PR until you find another solution. May be my idea is false and your PR will be better.

lsurma:
@diegomodolo 
Wouldn't it make more sense to have an asynchronous version of this 'OnDialogValidation'?
`Func<Task<bool>> OnDialogValidationAsync`
Most of the time, some validation logic will require async calls to DB, API etc, so having sync callbacks only seems not to be enough (at least for me).
Or am I just missing something and it's possible to call some async method right now?

> You could add this parameter to a inherited class of DialogInstance (like `DialogInstanceClosing`).
> 
> My question is more: why to have similar "events". A `Func<bool>` like you did can solve this problem, but using a `Closing` event can be more intuitive, no? 🙂

@dvoituron And I feel the same way. The OnDialogValidation event/callback seems to be a really specific use case (but I agree that it could be used a lot), and the components in the core library should stay in a more generic way, and then maybe some more specific use cases should be implemented using the composition approach. 
Because now, if I'm right, the validation behaviour could be achieved using the OnDialogClosing event callback, and if the closing should be stopped then inside that callback we should throw an exception (which is far from perfect, I guess, but it works).

diegomodolo:
Hi @vnbaaij , I would like to apologize for not being active at this thread. 
It is my first contribution at a plublic repository like this, so I did not know the right steps.
I agree with the modification proposed by @lsurma to make the event async and I will make the changes.
Regarding @dvoituron proposal, I agree that having everything on the same event would make more sense, I did not develop it that way because I did not want to "mess" with the original code.

Anyway, I would like to know the correct steps to amend my PR, if anyone could help me.

vnbaaij:
No need to apologize. Your contributions are always appreciated! 

Just do the changes you want to make in your branch, push them to this PR and let us know know if and when you want us to review stuff.

andreaslauwersTC:
Hi @vnbaaij, @dvoituron  & @diegomodolo,
I was wondering it there already is an update of when these changes will be released? 😊

vnbaaij:
No news from our side. Thin @diegomodolo stille wants to make some changes to this PR. So it is not ready yet.

We never give dates or put dates on releases. We release when we think a release is ready.

src/Core/Components/Dialog/Parameters/DialogParameters.cs
```diff
@@ -167,6 +167,14 @@ public virtual HorizontalAlignment Alignment
...
     /// </remarks>
     public EventCallback<DialogInstance> OnDialogOpened { get; set; } = default!;
+
+    /// <summary>
+    /// Function that is called and awaited before the dialog is closed.
+    /// </summary>
+    /// <remarks>
+    /// This is a suitable callback to use when you need to validate the data in the dialog <em>before</em> it closes.
+    /// </remarks>
+    public Func<Task<bool>> OnDialogValidation { get; set; } = default!;
```
vnbaaij, line 177: Please change the name to `ValidateDialogAsync`. As discussed it is not an `EventCallback` and Async needs to be added because it returns a Task<bool>, indicating that it is an asynchronous method.

vnbaaij (dismissed):

vnbaaij:
Hey @diegomodolo! Would be great if you can make the final changes to this PR. Thanks.

diegomodolo:
Hi @vnbaaij , I've just made the changes. 
Sorry for the delay.

vnbaaij (approved):

PR #2656, opened 2024-09-13
[dev-v5] Add Debounce

dvoituron:
# [dev-v5] Add Debounce

The **Debounce** dispatcher delays the invocation of an action until a predetermined interval has elapsed since the last call.
This ensures that the action is only invoked once after the calls have stopped for the specified duration.

I created 2 classes:
1. `DebounceAction` inspired from Microsoft.Toolkit.Uwp.UI.DispatcherQueueTimerExtensions and using `System.Timers.Timer`
2. `DeboundTask` inspired from blog posts and using `System.Task`

The usage is identical, but I propose to use `Debounce: DebounceTask`. With the option of upgrading to the second class if problems are detected (to avoid losing this code).
And in some weeks, we could delete the second class to keep only one.

```csharp
var debounce = new Debounce();

debounce.Run(50, async () =>
{
    actionCalled = true;
    await Task.CompletedTask;
});
```


dvoituron:
@vnbaaij Which class do you prefer? 
1. `DebounceAction` inspired from Microsoft.Toolkit.Uwp.UI.DispatcherQueueTimerExtensions and using `System.Timers.Timer`
2. `DeboundTask` inspired from blog posts and using `System.Task`

vnbaaij:
I would be inclined to go with a `System. Task` based implementation. No real strong reason for that though 😁

dvoituron:
Oh. I would have taken the Timer to follow UWP 😁

dvoituron:
The usage of these 2 classes is identical. I propose to use `Debounce: DebounceTask`. With the option of upgrading to the second class if problems are detected (to avoid losing this code).
And in some weeks, we could delete the second class to keep only one.

src/Core/Utilities/InternalDebounce/DebounceAction.cs
```diff
@@ -0,0 +1,61 @@
...
+// MIT License - Copyright (c) Microsoft Corporation. All rights reserved.
+// ------------------------------------------------------------------------
+
+namespace Microsoft.FluentUI.AspNetCore.Components.Utilities.InternalDebounce;
+
+/// <summary>
+/// The DebounceTask dispatcher delays the invocation of an action until a predetermined interval has elapsed since the last call.
+/// This ensures that the action is only invoked once after the calls have stopped for the specified duration.
+/// </summary>
+[Obsolete("Use Debounce, which inherits from DebounceTask.")]
```
vnbaaij, line 11: Why are we adding an obsolete class?

vnbaaij (changes requested):

src/Core/Utilities/InternalDebounce/DebounceAction.cs
dvoituron: It is an `internal` class. To use only the other class and avoid mistakes. But I would like to keep this class to validate the usage in future days. Next we will choose or delete this class

vnbaaij (approved):

PR #2664, opened 2024-09-16
[DataGrid] Alter rendering to use table elements

vnbaaij:
### Update ###
All work is done but some changes might be made based on the discussion in #2667

---
I'm swapping out the web components for standard HTML table elements. Reasons for doing this:
1) the web components will probably not be available in v3 as of the first release. We can't do a v5 release without a DataGrid component
2) using standard table elements is, in this case, good for performance. Loading the DataGrid examples page (which is still too heavy) is **2.2** seconds quicker with this new rendering applied. 

Because of some HTML/CSS trickery, it is still possible to use the `GridTemplateColumns` parameter on the grid or the `Width` property on the columns to set column widths. 

There are going to be some **breaking changes** with this...

- ~~The option to **choose** between using the new way of adding column actions (introduced in 4.10) or doing it the previous way, will disappear. The new way will be the standard moving forward. This is done because of code maintenance reasons but also because we achieve much better accessibility with this.~~ The option to choose is still available
- A lot of changes have been made to the structure of the rendered content and the class names used. If you have been overriding grid classes in your own code, these probably no longer work.
- When you have rows in a table with content that needs to span multiple lines, you now need to use and set the new `MultiLine` (bool) parameter of the `FluentDataGrid`. Previously this could be done by setting a `RowClass` on a column but that does not work anymore

### DIFF
--- a/examples/Demo/Shared/Microsoft.FluentUI.AspNetCore.Components.xml
+++ b/examples/Demo/Shared/Microsoft.FluentUI.AspNetCore.Components.xml
@@ -1185,7 +1185,7 @@
         </member>
         <member name="P:Microsoft.FluentUI.AspNetCore.Components.ColumnBase`1.Index">
             <summary>
-            Gets or sets the (1-based) index of the column
+            Gets or sets the index (1-based) of the column
             </summary>
         </member>
         <member name="P:Microsoft.FluentUI.AspNetCore.Components.ColumnBase`1.Class">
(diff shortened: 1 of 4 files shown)


src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -76,6 +76,14 @@ public partial class FluentDataGrid<TGridItem> : FluentComponentBase, IHandleEve
     [Parameter]
     public bool Virtualize { get; set; }
 
+    /// <summary>
+    /// This is applicable only when using <see cref="Virtualize"/>. It defines how many additional items will be rendered
+    /// before and after the visible region to reduce rendering frequency during scrolling. While higher values can improve
+    /// scroll smoothness by rendering more items off-screen, they can also increase initial load times. Finding a balance
+    /// based on your data set size and user experience requirements is recommended. The default value is 3.
+    /// </summary>
+    [Parameter] public int OverscanCount { get; set; } = 3;
```
dvoituron, line 85: Can you set the `Parameter` just above the property?

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -243,10 +245,18 @@ public partial class FluentDataGrid<TGridItem> : FluentComponentBase, IHandleEve
     [Parameter]
     public bool AutoFit { get; set; }
 
+    [Parameter]
+    public DataGridRowSize RowSize { get; set; } = DataGridRowSize.Medium;
+
+    [Parameter]
+    public bool MultiLine { get; set; } = false;
```
dvoituron, line 252: Can you add comments?

And why this property is on Table and not on each Column?

dvoituron (changes requested):

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
vnbaaij: With the way the grid is rendered now, every column's cell height  in a row needs to be set to 100% when MultiLine is true. We can't therefore set it on a column and it must be set on grid level.

Comments added

dvoituron (changes requested):

dvoituron (approved):

PR #2674, opened 2024-09-17
[DataGrid] Fix column options popup being blank

### LINKED ISSUE #2669: fix: Unexpected column options icon in FluentDataGrid
# 🐛 Bug Report

Adding `ResizeType="DataGridResizeType.Discrete" ` causes the column option button with an unexpected icon to appear in all the column headers.

## 💻 Repro or Code Sample

```
            <FluentDataGrid Items="Results" ResizableColumns="true" ResizeType="DataGridResizeType.Discrete" ShowHover="true">
                <PropertyColumn Property="@(p => p.Mapping)">
                    <ColumnOptions>
                        <div class="search-box">
                            <FluentSearch type="search" Autofocus=true @bind-Value="_mappingFilter"
                                          @oninput="@(args => _mappingFilter = args.Value?.ToString() ?? string.Empty)"/>
                        </div>
                    </ColumnOptions>
                </PropertyColumn>
                <PropertyColumn Property="@(p => p.ValueA)" Title="A"/>
            </FluentDataGrid>
```
## 🤔 Expected Behavior

image
on the columns with  `ColumnOptions`, nothing on the ones without.

## 😯 Current Behavior

image
Pressing it on the ones without `ColumnOptions` opens an empty popup.

## 🌍 Your Environment

* OS & Device: Windows 11, PC
* Browser Microsoft Edge
* .NET 8.0 
* Fluent UI Blazor library Version 4.10

vnbaaij:
- Fix #2669 
- Make Reset and set width methods public

With implementing the `HeaderCellAsButtonWithMenu` a couple of things got mixed up and that is why issue #2669 happened. This PR fixes that.

We added the option to do resize within the same popup as the filter earlier. When using the new `HeaderCellAsButtonWithMenu` parameter set to `true`, the different column action will be shown with a menu. If `false`, the resize action (either discrete or exact and the filter action will show in the same popup:
image

### DIFF
--- a/src/Core/Components/DataGrid/FluentDataGrid.razor.cs
+++ b/src/Core/Components/DataGrid/FluentDataGrid.razor.cs
@@ -765,7 +765,7 @@ public partial class FluentDataGrid<TGridItem> : FluentComponentBase, IHandleEve
         //return Task.CompletedTask;
     }
 
-    internal async Task SetColumnWidthDiscreteAsync(int? columnIndex, float widthChange)
+    public async Task SetColumnWidthDiscreteAsync(int? columnIndex, float widthChange)
     {
         if (_gridReference is not null && Module is not null)
         {
@@ -773,7 +773,7 @@ public partial class FluentDataGrid<TGridItem> : FluentComponentBase, IHandleEve
         }
     }
 
-    internal async Task SetColumnWidthExactAsync(int columnIndex, int width)
+    public async Task SetColumnWidthExactAsync(int columnIndex, int width)
     {
         if (_gridReference is not null && Module is not null)
         {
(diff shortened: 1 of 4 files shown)


src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -765,23 +765,23 @@ public async Task OnKeyDownAsync(FluentKeyCodeEventArgs args)
         //return Task.CompletedTask;
     }
 
-    internal async Task SetColumnWidthDiscreteAsync(int? columnIndex, float widthChange)
+    public async Task SetColumnWidthDiscreteAsync(int? columnIndex, float widthChange)
```
dvoituron, line 768: Why do you want to move these methods to `public` ?
If this is the case, we will need to support these features and you need to add comments.

dvoituron (changes requested):

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
vnbaaij: For Aspire, it is needed to call these methods from the outside. I'll document them.

vnbaaij: Done

dvoituron (approved):

PR #2717, opened 2024-09-25
[dev-v5] Documentation: Add `{{ INCLUDE }}` element

dvoituron:
#  Documentation: Add `{{ INCLUDE }}` element

Add {{ INLCUDE [file] }} to generate a "Migration to v5" page and to include a section "Migration from v4" at the end of each components

```
{{ INCLUDE MigrationFluentButton }}
```

{FD6394B7-1DEB-4969-978B-983CC11FA75C}

{B98C706D-E77C-49CF-8443-3621F5A34F65}


vnbaaij (approved):
We should name all of them {{ INCLUDE File=Migrate... }} but we can do that in one rename action after merge. Migration it too French 😉

PR #2730, opened 2024-09-27
[DataGrid] Support for multiple IAsyncQueryExecutor registrations

miguelhasse:
# Pull Request

## 📖 Description

AsyncQueryExecutorSupplier modified so it looks-up multiple IAsyncQueryExecutor registrations and returns the first one that supports the provided IQueryable<T> items source.

### DIFF
--- a/src/Core/Components/DataGrid/Infrastructure/AsyncQueryExecutorSupplier.cs
+++ b/src/Core/Components/DataGrid/Infrastructure/AsyncQueryExecutorSupplier.cs
@@ -25,9 +25,9 @@ internal static class AsyncQueryExecutorSupplier
     {
         if (queryable is not null)
         {
-            var executor = services.GetService<IAsyncQueryExecutor>();
+            var executors = services.GetServices<IAsyncQueryExecutor>();
 
-            if (executor is null)
+            if (executors is null)
             {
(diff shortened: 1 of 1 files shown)


vnbaaij (commented):
Can you give an example of how and why this is handy/better

miguelhasse:
@vnbaaij this change allows an application to use multiple different IQueryProvider implementations instead of a single one. Suppose for instance you want to use EF Core's DataContext and DataServiceQuery for OData queries.
I'm actually working on a IAsyncQueryExecutor adaptor for DataServiceQuery.

vnbaaij:
I'm totally not up to speed on EF so don't know what DataServiceQuery does. But judging from the code this change does no do any harm.
Let's give @dvoituron a chance to review as well.

dvoituron:
I don't have much experience of this either :-) 

On the other hand, it would be interesting to add a Unit Test. 
At the moment we don't have many tests for the FluentDataGrid, but we're working on the next version v5, which will require all features to be unit-tested. And as you have more experience than we do, adding a test right now would be great. Otherwise we will have to temporarily disable this feature :-(

miguelhasse:
@dvoituron sure I'll take care of adding unit tests for this feature.

PR #2739, opened 2024-09-30
[DataGrid] changes related to loading behavior

### LINKED ISSUE #2734: fix: Loading == true ignored when Items == null in DataGrid
# 🐛 Bug Report

`items == null` overrides `loading == true`, such that `No data to show!` is displayed instead of `Loading..`
prior to the IEnumerable having data.

## 💻 Repro or Code Sample

```razor
@rendermode @(new InteractiveAutoRenderMode(prerender: true))

				
		<FluentDataGrid Items="MyData?.AsQueryable()" Loading="@(MyData is null)">
			...
		</FluentDataGrid>
	

@code {

	[Parameter] public IEnumerable<MyEntity>? MyData { get; set; }
}
```

## 🤔 Expected Behavior

Collections are either null, empty, or full. I would expect the most common state where loading should display is when the data is null, during data loading. If it comes back empty that is when "No data" takes effect. Or in any case, Loading == true should display loading regardless of anything else.

## 😯 Current Behavior

When `Items` is empty, `Loading == true` takes effect and displays properly.

```razor
<FluentDataGrid Items="(RecentTreasuries ?? []).AsQueryable()" Loading="@(MyData is null)">
    ...
</FluentDataGrid>
```

## 💁 Possible Solution

Give precedence to `Loading == true`

## 🔦 Context

Just unexpected, and I guess I could throw that empty list in there but seems better the other way, just using the null parameter as is.

## 🌍 Your Environment

* OS & Device: Windows 11
* Browser Microsoft Edge
* .NET 9 Preview, Fluent UI Blazor 4.10.1

vnbaaij:
- Fix #2734
- Create separate page for testing issues
- Update version of Core and Icons packages in `Directory.Packages.props `

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -10,8 +10,8 @@
   </PropertyGroup>
   <ItemGroup>
     <!-- For Sample Apps -->
-    <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components" Version="4.10.0" />
-    <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components.Icons" version="4.10.0" />
+    <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components" Version="4.10.1" />
+    <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components.Icons" version="4.10.1" />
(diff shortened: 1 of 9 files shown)


vnbaaij:
Have confirmed it is working as expected against the repo supplied in the issue

dvoituron (approved):

PR #2758, opened 2024-10-04
[Docs] Fix - Nullable reference types not always shown in API documentation

oneolddev:
# Pull Request

## 📖 Description

API documentation did not always show nullable type for parameters or method return.

### 🎫 Issues

- Only the underlying type shown even though it was declared nullable. It does not match the [source code](https://github.com/microsoft/fluentui-blazor/blob/dev/src/Core/Components/Grid/FluentGridItem.razor.cs).

#### Before
image

#### After
image

## 👩‍💻 Reviewer Notes

Changes were made to the `ToTypeNameString` extensions instead of `APIDocumentation`. Only concern I have is that the original extensions were sourced from https://github.com/loxsmoke/DocXml.

## 📑 Test Plan

- I manually checked many of the components. 

### General

### DIFF
--- a/examples/Demo/Shared/ReflectionExtensions.cs
+++ b/examples/Demo/Shared/ReflectionExtensions.cs
@@ -181,9 +181,12 @@ public static class ReflectionExtensions
     public static string ToTypeNameString(this MethodInfo methodInfo, Func<Type, Queue<string>, string> typeNameConverter = null,
         bool invokeTypeNameConverterForGenericType = false)
     {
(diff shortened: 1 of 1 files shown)


vnbaaij (approved):

vnbaaij:
> Changes were made to the ToTypeNameString extensions instead of APIDocumentation. Only concern I have is that the original extensions were sourced from https://github.com/loxsmoke/DocXml.

Hi Gary (@oneolddev). Why does that concern you? We used that as the base back then but have made significant changes along the way. We are not going back anytime to copying the original code again, I think.

PR #2760, opened 2024-10-04
[AppBar] Make it work in horizontal orientation as well

vnbaaij:
Initial work to make AppBar work in horizontal orientation. More work needed before this can be released:

- [x]  Make Overflow script 'cut-off' value configurable
- [x] - Test with smaller bar heigh/icon sizes
- [x] - Unit Tests

Result so far:
AppBar-horizontal


src/Core/Components/AppBar/FluentAppBar.razor.cs
```diff
@@ -117,6 +131,11 @@ public async Task OverflowRaisedAsync(string value)
         await InvokeAsync(StateHasChanged);
     }
 
+    private async Task InitializeOverflowAsync()
+    {
+        await _jsModuleOverflow.InvokeVoidAsync("fluentOverflowInitialize", _dotNetHelper, Id, Orientation == Orientation.Horizontal, OVERFLOW_SELECTOR);
```
dvoituron, line 136: You should test `_jsModuleOverflow` before executing this method.
It is likely that this object will not be null, but there is no guarantee that a Blazor event or unit test will try to execute it.

src/Core/Components/AppBar/FluentAppBar.razor
```diff
@@ -23,7 +23,7 @@
                                  HorizontalAlignment="HorizontalAlignment.Center"
                                  VerticalAlignment="VerticalAlignment.Center"
                                  VerticalGap="0"
-                                 Style="width: 100%; --appbar-active-indicator-width: calc(var(--design-unit)* 0.5px); height: 56px;">
+                                 Style="--appbar-active-indicator-width: calc(var(--design-unit)* 0.5px); width: 56px; height: 56px;">
```
dvoituron, line 26: Why not to use `var(--appbar-item-height, 56px)` for width and height (like done in `FluentAppBar.razor.css` ?

examples/Demo/Shared/Pages/AppBar/Examples/AppBarDefault.razor
```diff
@@ -75,7 +84,7 @@
 
 
 @code {
-
+    private bool _orientation = true;
```
dvoituron, line 87: You should call that `bool _vertical = true`

dvoituron (changes requested):

src/Core/Components/AppBar/FluentAppBar.razor
vnbaaij: I'm creating a new CSS var for this: --appbar-item-size

src/Core/Components/Overflow/FluentOverflow.razor.js
```diff
@@ -1,7 +1,9 @@
 let resizeObserver;
 let observerAddRemove;
+let lastId;
```
dvoituron, line 3: Why did you need these variables?

`lastId` will be overwritten when several **FluentOverflow** are used at the same time, as it is a single variable for all components. Is this correct?

dvoituron (changes requested):

src/Core/Components/Overflow/FluentOverflow.razor.js
vnbaaij: I have tested and confirmed this is working as it should (added OverflowDefault example to AppBarDefault example). Is is only of interest if and when the orientation of an AppBar is changed which will typically not happen as frequently. In the example we now provide that option explicitly but that is not the case for regular usage, I think.

dvoituron (approved):

PR #2764, opened 2024-10-05
[Docs] Fixes missing method descriptions for APIs

oneolddev:
# Pull Request

## 📖 Description

Descriptions for methods on some APIs are mssing.

#### Missing description from `FluentGridItem.GetHiddenAttribute` and `ToastService`

image

image

#### After fix

image

image

## 👩‍💻 Reviewer Notes

The information is in the generated `CodeComments` file. The `key` is not being generated correctly because the method is either `static` or one of the parameters is an `EventCallback<T>`.

## 📑 Test Plan

I manually checked several components.

### General

### DIFF
--- a/examples/Demo/DocGenerator/CodeCommentsGenerator.cs
+++ b/examples/Demo/DocGenerator/CodeCommentsGenerator.cs
@@ -79,6 +79,8 @@ public class CodeCommentsGenerator : IIncrementalGenerator
         value = regex.Replace(value, "");
         regex = new("[P,T,M,F]:FluentUI\\.Demo\\.Shared\\.");
         value = regex.Replace(value, "");
+        regex = new("\\([^()]*\\)");
+        value = regex.Replace(value, "");
 
         return value;
     }
--- a/examples/Demo/DocGenerator/CodeCommentsGenerator.cs
+++ b/examples/Demo/DocGenerator/CodeCommentsGenerator.cs
@@ -79,7 +79,7 @@ public class CodeCommentsGenerator : IIncrementalGenerator
         value = regex.Replace(value, "");
         regex = new("[P,T,M,F]:FluentUI\\.Demo\\.Shared\\.");
         value = regex.Replace(value, "");
-        regex = new("\\([^()]*\\)");
+        regex = new("\\(.*\\)$");
         value = regex.Replace(value, "");
 
         return value;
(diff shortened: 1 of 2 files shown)


oneolddev:
Examinging the `CodeComments.g.cs` file, you find the following:
`["FluentGridItem.GetHiddenAttribute(System.Nullable{Microsoft.FluentUI.AspNetCore.Components.GridItemHidden})"] = "Returns the hidden attribute value for the specified <code>GridItemHidden</code> value.", `

The key being generated also has the following: `(System.Nullable{Microsoft.FluentUI.AspNetCore.Components.GridItemHidden})` which is not part of the method name.

vnbaaij (approved):

examples/Demo/DocGenerator/CodeCommentsGenerator.cs
```diff
@@ -79,6 +79,8 @@ private static string CleanupParamName(string value)
         value = regex.Replace(value, "");
         regex = new("[P,T,M,F]:FluentUI\\.Demo\\.Shared\\.");
         value = regex.Replace(value, "");
+        regex = new("\\([^()]*\\)");
```
dvoituron, line 82: If ChatGPT give me the correct answer: "The goal of this regular expression (\([^()]*\)) is to match any sequence of characters that is enclosed within parentheses, without any nested parentheses inside."

Is it possible to be more precise to avoid replacing s’ (abc) ’ written by ourselves in the documentation?

dvoituron (dismissed):

examples/Demo/DocGenerator/CodeCommentsGenerator.cs
oneolddev: I did some further testing after drafting and found a few more issues. This fix does not address overloaded methods. For example for `ToastService.ShowToast`, the fix is only returning the description for the first.

@dvoituron  I'm working on another fix which will reverse the last commit.

oneolddev: @dvoituron 

It appears that this may be the best approach. That is, to correct the `key` being generated in the `CodeCommentGenerator`.  The intention is the remove parameter description from the "method name" in the entry created in the generated `CodeComments` class.

``` text
["ToastService.ShowToast``2(Microsoft.FluentUI.AspNetCore.Components.ToastParameters{``1})"] = "Shows the toast with the component type", 
```
to
``` text
["ToastService.ShowToast``2"] = "Shows the toast with the component type",
```

Good catch on the regular expression 😆, It's been a while since I've used regular expressions so I used ChatGPT to generate the initial expression. I'm using `"\\(.*\\)$"` now.  It will remove anything between the first opening parenthes and last closing parentheses including the parentheses on any string ending with a closing parentheses. 🤞

oneolddev:
I got a better understanding of the xml documentation file after finding https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/#id-strings. I changed the description retrieval code to handle the name mangling of overloaded method names with generic parameters.

image

vnbaaij (approved):

PR #2793, opened 2024-10-11
[MenuProvider] Check whether the FluentMenuProvider is included

dvoituron:
# Check whether the `FluentMenuProvider` is included

We've added a new `FluentMenuProvider` to manage menus more simply.
By default, it is used but requires the presence of `<FluentMenuProvider />` in the **MainLayout**. 
If this is not the case and the developer does not explicitly request `UseMenuService=‘false’`, an error is triggered.

{BFE5C91C-1D07-4D9F-B136-F4E446B0FEFB}


vnbaaij (approved):
We need to do the same for the other providers

PR #2796, opened 2024-10-11
[Slider] Make label respond to `Disabled` state

### LINKED ISSUE #2794: fix: FluentSlider's FluentSliderLabels doesn't recover from Disabled=true
# 🐛 Bug Report

If a FluentSlider's `Disabled=true` by default, the FluentSliderLabels doesn't recover properly when setting `Disabled=false` during runtime. If `Disabled=false` by default and manually toggled during runtime, everything works fine.

**Current behavior:** Slider `Disabled=true` by default and was set to `false` during runtime (FluentSliderLabels are still washed out).
Image

**Expected behavior:** Slider `Disabled=false` by default and toggled during runtime (FluentSliderLabels are displayed properly).*
Image

## 💻 Repro or Code Sample

See [FluentBlazorTest](https://github.com/RobGess/FluentBlazorTest) or:

1. Create new fluent blazor project: `dotnet new fluentblazor --name FluentBlazorTest`
2. Add `@rendermode InteractiveServer` to Home.razor pager
3. Add FluentSlider, FluentButton and FluentLabel to toggle Disabled property and display current state:
```
<FluentStack Orientation="Orientation.Vertical" VerticalGap="30" Style="margin-top:2rem;">
    <FluentSlider Min="0" Max="100" Step="1" Disabled=@_disabled Style="width:50%">
        <FluentSliderLabel Position="0">0</FluentSliderLabel>
        <FluentSliderLabel Position="30">30</FluentSliderLabel>
        <FluentSliderLabel Position="50">50</FluentSliderLabel>
        <FluentSliderLabel Position="70">70</FluentSliderLabel>
        <FluentSliderLabel Position="90">90</FluentSliderLabel>
        <FluentSliderLabel Position="100">100 %</FluentSliderLabel>
    </FluentSlider>

    <FluentButton OnClick="ToggleDisabled">Toggle slider</FluentButton>
    <FluentLabel>Slider disabled is @_disabled</FluentLabel>
</FluentStack>
```

## 🤔 Expected Behavior

FluentSliderLabel's color should be set properly.

## 😯 Current Behavior

FluentSliderLabel's color is not set properly.

## 🌍 Your Environment

* OS: Windows 11 Enterprise
* Google Chrome Version 129.0.6668.101 (64-Bit), MS Edge Version 129.0.2792.89 (64-Bit)
* Microsoft.FluentUI.AspNetCore.Components 4.10.2

vnbaaij:
Fix #2794 by adding an owning slider cascading parameter and set label's disabled state to owner value.

After:
fix-issue-#2794


tests/Core/Microsoft.FluentUI.AspNetCore.Components.Tests.csproj
```diff
@@ -59,4 +59,10 @@
       <DependentUpon>%(ParentFile).razor</DependentUpon>
     </None>
   </ItemGroup>
+
+  <ItemGroup>
+    <None Update="Slider\FluentSliderLabelTests.FluentSliderLabel_Default.received.html">
```
dvoituron, line 64: I think this line can be removed.

src/Core/Components/Slider/FluentSliderLabel.razor.cs
```diff
@@ -38,10 +38,16 @@ public FluentSliderLabel()
     public bool? HideMark { get; set; }
 
     /// <summary>
-    /// Gets or sets disabled state of the label. This is generally controlled by the parent.
+    /// Gets the disabled state of the label. This is controlled by the owning <see cref="FluentSlider{TValue}"/>.
     /// </summary>
     [Parameter]
-    public bool? Disabled { get; set; }
+    public bool? Disabled {
```
dvoituron, line 44: That's a breaking change: previously it was possible to use `FluentSliderLabel` outside `FluentSlider` (I don't know why it was created that way, but it is :-)

So, probably you should manage the possibility to have `Owner is null` and keep an private Disabled value in this case.

```csharp
[Parameter]
public bool? Disabled
{
    get => Owner is null ? _disabled : Owner.Disabled;
    set => _disabled = Owner is null ? value : Owner.Disabled;
}
```

dvoituron (changes requested):

src/Core/Components/Slider/FluentSliderLabel.razor.cs
vnbaaij: I don't know why it was created like that either 🤔

Your change is indeed non-breaking so better. Will change it

dvoituron (approved):

PR #2812, opened 2024-10-14
[DataGrid] Add mechanism to not fire single click when a row is double clicked

### LINKED ISSUE #2789: DataGrid RowFocus DoubleClick
# 🙋 Feature Request

I know this was raised previously under https://github.com/microsoft/fluentui-blazor/issues/1953 and you've closed integrating into Multiselect .
However that doesn't seem to allow a double-click as it doesn't pass the row details - we use rowfocus for that which only support single clicks  - am I missing something ?

vnbaaij:
Add delay on single click to prevent it from firing when double clicked. Remote data example has been updated to show the working of the new feature. Fixes #2789 

After:
fix-issue-#2789

### DIFF
--- a/examples/Demo/Shared/Microsoft.FluentUI.AspNetCore.Components.xml
+++ b/examples/Demo/Shared/Microsoft.FluentUI.AspNetCore.Components.xml
@@ -3517,6 +3517,7 @@
             Gets or sets the dialog position:
             left (full height), right (full height)
             or screen middle (using Width and Height properties).
+            HorizontalAlignment.Stretch is not supported for this property.
             </summary>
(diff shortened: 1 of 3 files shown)


dvoituron:
Why do we have to do this? This is the default behavior in HTML/JavaScript. You have the same "click - click - dblclick" sequence with the `div` tag. And with this PR, the click will not happen immediately :-(

```html
<div onclick="console.log('click')" ondblclick="console.log('double')">
	HERE
</div>
```

peek_1

vnbaaij:
Closing this because of it being a standard HTML/JavaScript behavior

PR #2855, opened 2024-10-23
[ComboBox] Fix OnValueChanged being called multiple times

### LINKED ISSUE #2851: FluentCombobox OnValueChanged is called twice when the item selected has more than one space in the middle of the items string or an extra space at the beginning or end of the string.
# 🐛 Bug Report

FluentCombobox OnValueChanged is called twice when the item selected has more than one space in the middle of the items string or an extra space at the beginning or end of the string.

When OnValueChanged is called twice, the function is reentered a second time prior to the first call being completed.

This cannot be fixed in user code without elaborate workarounds when data comes from a database or other source.

## 💻 Repro or Code Sample

```
@page "/test"

<FluentCombobox Items="@_items" ValueChanged="@OnValueChanged" />

@code
{
    private List<string> _items = ["good", "good good", "bad  bad", "bad ", " bad"];

    private async Task OnValueChanged(string? value)
    {
        Debug.WriteLine($"Start: '{value}')");

        await Task.Delay(100);

        Debug.WriteLine($"End: '{value}')");
    }
}
```

## 🤔 Expected Behavior

    When you select 'good' you get this expected output in the Output window:
    Start: 'good')
    End: 'good')

    When you select 'good good' you get this expected output in the Output window:
    Start: 'good good')
    End: 'good good')

## 😯 Current Behavior

    When you select 'bad  bad' you get this unexpected output in the Output window:
    Start: 'bad bad')
    Start: 'bad  bad')
    End: 'bad  bad')
    End: 'bad bad')

    When you select 'bad ' you get this unexpected output in the Output window:
    Start: 'bad')
    Start: 'bad ')
    End: 'bad ')
    End: 'bad')

    When you select ' bad' you get this unexpected output in the Output window:
    Start: 'bad')
    Start: ' bad')
    End: 'bad')
    End: ' bad')

## 🌍 Your Environment

.NET 8
Fluent UI Blazor library Version 4.10.2
Blazor Server
Microsoft Edge for Business Version 130.0.2849.52 (Official build) (64-bit)

vnbaaij:
Fix #2851 by not calling `base.ChangeHandlerAsync` when no item selected. This could happen if the  Value and the Text of an option differed.

See original issue for before output. Now all test situations render as expected.

### DIFF
--- a/src/Core/Components/List/FluentCombobox.razor.cs
+++ b/src/Core/Components/List/FluentCombobox.razor.cs
@@ -138,15 +138,19 @@ public partial class FluentCombobox<TOption> : ListComponentBase<TOption> where 
             if (item is null)
             {
                 SelectedOption = default;
-                await base.ChangeHandlerAsync(e);
             }
             else
             {
                 await OnSelectedItemChangedHandlerAsync(item);
             }
-
         }
     }
+
+    protected override async Task InputHandlerAsync(ChangeEventArgs e)
+    {
+        await base.ChangeHandlerAsync(e);
+    }
+
 
     protected override string? GetOptionValue(TOption? item)
     {
--- a/src/Core/Components/List/FluentCombobox.razor.cs
+++ b/src/Core/Components/List/FluentCombobox.razor.cs
@@ -143,14 +143,13 @@ public partial class FluentCombobox<TOption> : ListComponentBase<TOption> where 
             {
                 await OnSelectedItemChangedHandlerAsync(item);
             }
+
+            if (Value != value)
+            {
+                await base.ChangeHandlerAsync(e);
+            }
         }
     }
-
-    protected override async Task InputHandlerAsync(ChangeEventArgs e)
-    {
-        await base.ChangeHandlerAsync(e);
-    }
-
 
     protected override string? GetOptionValue(TOption? item)
     {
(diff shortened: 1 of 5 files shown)


src/Core/Components/List/FluentCombobox.razor.cs
```diff
@@ -138,16 +138,20 @@ protected override async Task ChangeHandlerAsync(ChangeEventArgs e)
...
             {
                 await OnSelectedItemChangedHandlerAsync(item);
             }
-
         }
     }
 
+    protected override async Task InputHandlerAsync(ChangeEventArgs e)
+    {
+        await base.ChangeHandlerAsync(e);
```
dvoituron, line 151: Why did you remove the usage of `Immediate`?
The method `FluentInputBase.InputHandlerAsync` will not used.

dvoituron (changes requested):

src/Core/Components/List/FluentCombobox.razor.cs
vnbaaij: Fixed in last commit

vnbaaij:
`Immediate` (and `ImmediateDelay`) now work again for ComboBox. The double calling of `ChangeHadlerAsync` is fixed by using a specific JS custom event handler. The web conponent was returning a incorrect value in the case of leading/trailing and double spaces. We now look at the actual value in the option.

Image below shows output from the test code. At first green arrow, text is being deleted from the text input with `Immediate` set to false. On pressing enter the value is then handled. At second green arrow, `Immediate` is set to true. You then see every typed character appear in the value

image

dvoituron (approved):

PR #2865, opened 2024-10-25
[FileInput] Fix uploading the same file twice by drag&drop

### LINKED ISSUE #2849: FluentInputFile: second+ drop of same file (same file name) does not trigger OnUploadCompletedAsync
exactly in the first example here https://fluentui-blazor.net/InputFile

vnbaaij:
Fix #2849. This PR changes the way drag & drop are detected. Now based on JS instead of doing it from the Blazor side. JS code based on https://www.meziantou.net/upload-files-with-drag-drop-or-paste-from-clipboard-in-blazor.htm

### DIFF
--- a/src/Core/Components/InputFile/FluentInputFile.razor
+++ b/src/Core/Components/InputFile/FluentInputFile.razor
@@ -7,7 +7,7 @@
      disabled="@Disabled"
      @ref="@_containerElement">
 
-    <div class="inputfile-content" style="@($"z-index: {ZIndex.InputFileDropZone + 1} ")">
+    <div class="inputfile-content" style="@($"z-index: {ZIndex.InputFileDropZone + 1}")">
 
         @ChildContent
 
--- a/src/Core/Components/InputFile/FluentInputFile.razor
+++ b/src/Core/Components/InputFile/FluentInputFile.razor
@@ -2,12 +2,12 @@
 @using Microsoft.AspNetCore.Components.Forms
 @inherits FluentComponentBase
 
-<div class="@ClassValue" 
+<div class="@ClassValue"
      style="@StyleValue"
      disabled="@Disabled"
      @ref="@_containerElement">
 
-    <div class="inputfile-content">
+    <div class="inputfile-content" style="@($"z-index: {ZIndex.InputFileDropZone + 1} ")">
 
         @ChildContent
 
--- a/src/Core/Components/InputFile/FluentInputFile.razor
+++ b/src/Core/Components/InputFile/FluentInputFile.razor
@@ -2,19 +2,12 @@
 @using Microsoft.AspNetCore.Components.Forms
 @inherits FluentComponentBase
 
-<div class="@ClassValue"
+<div class="@ClassValue" 
      style="@StyleValue"
-     drop-files="@DropOver"
      disabled="@Disabled"
-     @ondragenter="@(e => DropOver = true)"
-     @ondragenter:stopPropagation
-     @ondragover="@(e => DropOver = true)"
-     @ondragover:stopPropagation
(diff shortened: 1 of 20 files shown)


src/Core/Components/InputFile/FluentInputFile.razor
```diff
@@ -2,19 +2,12 @@
...
-     @ondragenter:stopPropagation
-     @ondragover="@(e => DropOver = true)"
-     @ondragover:stopPropagation
-     @ondragleave="@(e => DropOver = false)"
-     @ondragleave:stopPropagation
-     @ondragend="@(e => DropOver = false)">
+     @ref="@_containerElement">
 
-    <div class="inputfile-content" style="z-index: @(DropOver ? null : 999)">
+    <div class="inputfile-content">
```
dvoituron, line 10: To verify but this `z-index` is important to have the possibility to click on panel links.

src/Core/Components/InputFile/FluentInputFile.razor.js
```diff
@@ -21,3 +21,51 @@ export function previewImage(inputElem, index, imgElem) {
...
+
+    function onPaste(e) {
+        // Set the files property of the input element and raise the change event
+        inputFile.files = e.clipboardData.files;
+        const event = new Event('change', { bubbles: true });
+        inputFile.dispatchEvent(event);
+    }
+
+    // Register all events
+    containerElement.addEventListener("dragenter", onDragHover);
```
dvoituron, line 55: I prefer to keep the `div` attributes to avoid to create event listener. What do you think?

dvoituron (changes requested):

src/Core/Components/InputFile/FluentInputFile.razor.js
vnbaaij: This is needed to fix the fact that the attributes don't trigger anything when a file is dropped a second tim. With JS this does happen

src/Core/Components/InputFile/FluentInputFile.razor.js
```diff
@@ -21,3 +21,51 @@ export function previewImage(inputElem, index, imgElem) {
...
+    containerElement.addEventListener("dragover", onDragHover);
+    containerElement.addEventListener("dragleave", onDragLeave);
+    containerElement.addEventListener("drop", onDrop);
+    containerElement.addEventListener('paste', onPaste);
```
dvoituron, line 59: I don't understand how this ‘paste’ function can work with our component.
1. Because, by default, when the user clicks on the `FluentInputFile`, the `Open File` dialog box opens and the user cannot paste the file.
2. This functionality must be removed when the component is deactivated (see the last example in the demo site).

tests/Core/InputFile/FluentInputFileTests.FluentInputFile_Default.verified.razor.html
```diff
@@ -1,6 +1,6 @@
 
 <div class="fluent-inputfile-container" blazor:ondragenter="1" blazor:ondragenter:stoppropagation="" blazor:ondragover="2" blazor:ondragover:stoppropagation="" blazor:ondragleave="3" blazor:ondragleave:stoppropagation="" blazor:ondragend="4" b-72uj7bp8pi="">
-  <div class="inputfile-content" style="z-index: 999" b-72uj7bp8pi="">
+  <div class="inputfile-content" b-72uj7bp8pi="">
```
dvoituron, line 3: With you last commit, the unit tests must be updated :-)

dvoituron (changes requested):

dvoituron (approved):

PR #2921, opened 2024-11-11
Add Support to the DataGrid Control for Complex Layout Rendering. Wit…

Christian-Schl:
# Pull Request

## 📖 Description

Currently the FluentDataGrid has problem to collect all the columns if the rendering structure is nested with some wrapper component. The Target of this PR is to unbound the Grid from internal rendering logic of blazor by using the well known apis.
On bigger projects the dev always try to wrap components to integrate there own extensions an logics. 

### 🎫 Issues

I found no issue for this topic

## 👩‍💻 Reviewer Notes

Reviewer can check the updated DataGrid example page. There is now a Example which generate the columns in a nested way

## 📑 Test Plan

I think the wohle DataGrid Control need to be tested. I have dont only the minimal change to dont intercept other functions but i am not sure about side effects

### General

### Component-specific

### DIFF
--- a/src/Core/Components/DataGrid/Columns/ColumnBase.razor
+++ b/src/Core/Components/DataGrid/Columns/ColumnBase.razor
@@ -2,9 +2,8 @@
 @using Microsoft.FluentUI.AspNetCore.Components.DataGrid.Infrastructure
 @namespace Microsoft.FluentUI.AspNetCore.Components
 @typeparam TGridItem
-@{
-    Grid.AddColumn(this, InitialSortDirection, IsDefaultSortColumn);
-}
+@implements IDisposable
+
 @code
 {
     private void RenderDefaultHeaderContent(RenderTreeBuilder __builder)
@@ -171,4 +170,18 @@
             @PlaceholderTemplate(placeholderContext)
         }
     }
+
+    protected override void OnParametersSet()
+    {
+        base.OnParametersSet();
+        Grid.AddColumn(this, InitialSortDirection, IsDefaultSortColumn);
+    }
+
+    public void Dispose()
+    {
+        if (Grid != null)
+            Grid.RemoveColumn(this);
+
+    }
+
 }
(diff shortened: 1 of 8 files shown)


Christian-Schl:

src/Core/Components/DataGrid/Columns/ColumnBase.razor
```diff
@@ -171,4 +170,18 @@
             @PlaceholderTemplate(placeholderContext)
         }
     }
+
+    protected override void OnParametersSet()
+    {
+        base.OnParametersSet();
+        Grid.AddColumn(this, InitialSortDirection, IsDefaultSortColumn);
```
dvoituron, line 177: I don't think you can add this `AddColumn` to the `OnParametersSet` method.
This method will be triggered several times: each time a parameter is updated.

dvoituron (changes requested):
I think there are too many changes applied to the base classes and we don't yet have enough unit tests to check that the code isn't regressed. It would be preferable to add this functionality in the next v5 version, which will bring more changes to the code and unit tests.

vnbaaij:
I agree with @dvoituron. This is not something that we want to merge at the moment. 

I don't think that we have had any issues being created on what you have put in the description. So not sure if there is a problem we are solving with this or if it is more refactoring for the refactoring. 

I'll close this PR now. We **do** appreciate the effort but for the next time perhaps first open a discussion item to see if what you have in mind is something we and the users need/want.

Christian-Schl:
Sorry for the late response, I’ve been pretty busy.

It’s absolutely fine if you don’t want to merge the PR—I understand it might look like a critical change for you. The reason I didn’t create an issue for it is that sometimes it’s easier to create a PR and discuss it directly rather than working from an abstract issue. After all, we’re all developers. 😊

The problem I described in the PR description does exist, so perhaps this PR can serve as a starting point when the issue becomes a higher priority.

Thanks for reviewing my PR! I really appreciate all the work you’re doing to build this great component library.

PR #2932, opened 2024-11-14
[dev-v5] Change to .NET 9 GA version all the way. 

vnbaaij:
Remove .NET 8 packages. settings, etc

### DIFF
--- a/.github/workflows/build-core-lib.yml
+++ b/.github/workflows/build-core-lib.yml
@@ -42,16 +42,11 @@ jobs:
      - name: Checkout source
        uses: actions/checkout@v4
 
-     - name: Setup .NET 8.0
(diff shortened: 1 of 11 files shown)


dvoituron:
Why do you want to remove NET8? Because NET9 os not LTS and will probably be restricted for some companies?

vnbaaij:
This is how we have done it before as well. Our major releases target the latest .NET version so we can fully use that version's latest enhancements.

v3 supports .NET 6 and 7 (and is not supported anymore as of now)
v4 supports .NET 8 and 9
v5 support >NET 9 and 10
:
:
image

If someone does not want to use that specific .NET version, they can continue using our previous version.

dvoituron-msft (approved):

PR #2935, opened 2024-11-15
[dev-v5] Add FluentLabel component

vnbaaij:
Label is now exclusively being used for labeling input fields. 

For a more v4 compatible component to show text using Fluent's opinions on typography, we will add a new Text component.

Includes Unit Tests
image


src/Core/Microsoft.FluentUI.AspNetCore.Components.csproj
```diff
@@ -115,5 +115,8 @@
       <LastGenOutput>%(ParentFile).Designer.cs</LastGenOutput>
     </EmbeddedResource>
   </ItemGroup>
+  <ItemGroup>
+    <Folder Include="Components\TextInput\" />
```
dvoituron, line 119: Can you remove this line?

tests/Core/Components/Label/FluentLabelTests.razor
```diff
@@ -0,0 +1,129 @@
...
+    public FluentLabelTests()
+    {
+        //Services.AddSingleton<LibraryConfiguration>();
+    }
+
+    [Fact]
+    public void FluentLabel_Default()
+    {
+        // Arrange
+        using var id = Identifier.SequentialContext();
```
dvoituron, line 16: You don't need this line

tests/Core/Components/Label/FluentLabelTests.razor
```diff
@@ -0,0 +1,129 @@
...
+
+        // Assert
+        cut.Verify();
+    }
+
+    [Fact]
+    public void FluentLabel_Required()
+    {
+        // Arrange
+        using var id = Identifier.SequentialContext();
```
dvoituron, line 67: You don't need this line

src/Core/Components/Label/FluentLabel.razor.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// Gets or sets wether the label show a required marking (red star).
+    /// </summary>
+    [Parameter]
+    public bool Required { get; set; }
+
+    /// <summary>
+    /// Gets or sets the size of the label.
+    /// </summary>
+    [Parameter]
+    public LabelSize Size { get; set; } = LabelSize.Medium;
```
dvoituron, line 34: Did you tried using a nullable property, to verify is the HTML code contains the attribute or not?
If not, probably better to allow nullable property. The result will be shorter.

examples/Demo/FluentUI.Demo.Client/Documentation/GetStarted/Migration/MigrationFluentLabel.md
```diff
@@ -0,0 +1,25 @@
...
+- `Size`, sets the label font size.
+- `Required`, sets required field styling by adding an asterisk 
+
+### Changed properties 🔃
+- `Weight`, now used to determine if the label text is shown regular or semibold
+
+### Removed properties💥
+
+- `Alignment`
+- `Color`
```
dvoituron, line 18: Can we add `Color` and `CustomColor`? So the dev don't need to add a CSS style to do that (and we keep more compatibility with Lib v4)?

Same for `MarginBlock`

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Label/Examples/LabelDisabled.razor
```diff
@@ -0,0 +1,3 @@
+﻿<div style="display: flex; gap: 12px; margin-bottom: 24px;">
```
dvoituron, line 1: This is a comment I've had: we're using ‘too many’ styles to get the right result.
I think we need to use examples with only the FluentUI Blazor components.

If something is missing to get the expected result, it's because we've forgotten something: property, other component, etc.

What is your opinion?

dvoituron (changes requested):

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Label/Examples/LabelDisabled.razor
vnbaaij: Yes, we need to make this more 'simple'

examples/Demo/FluentUI.Demo.Client/Documentation/GetStarted/Migration/MigrationFluentLabel.md
vnbaaij: No, not for the Label component. We do want to add those to the Text component

src/Core/Components/Label/FluentLabel.razor.cs
vnbaaij: Web Component always renders size and weight attributes

dvoituron: I'm not talking about Web Components, but this code gives the ‘default’ values and is shorter. So it's better for the exchange between server and client: avoid including ‘extra attributes’.
In the WebComponents v3 documentation, these attributes are not mandatory either

```xml
<fluent-label>Label</fluent-label>
```

vnbaaij: Working on it!

dvoituron (approved):

PR #2936, opened 2024-11-15
[dev-v5] Add FluentText component

vnbaaij:
Adds the FluentText component. This is a replacement for the v4 Label component.

image

Unit Test added with 100% line coverage

### DIFF
--- a/src/Core/Components/Text/FluentText.razor.cs
+++ b/src/Core/Components/Text/FluentText.razor.cs
@@ -1,0 +1,138 @@
+// ------------------------------------------------------------------------
+// MIT License - Copyright (c) Microsoft Corporation. All rights reserved.
+// ------------------------------------------------------------------------
+using Microsoft.AspNetCore.Components;
+using Microsoft.FluentUI.AspNetCore.Components.Extensions;
+using Microsoft.FluentUI.AspNetCore.Components.Utilities;
+
+namespace Microsoft.FluentUI.AspNetCore.Components;
+
+/// <summary>
+/// FluentText component, codifies Fluent's opinions on typography to make them easy to use and standardize across products.
+/// </summary>
+public partial class FluentText : FluentComponentBase
+{
+    /// <summary />
+    protected string? ClassValue => new CssBuilder(Class)
+        .Build();
+
+    /// <summary />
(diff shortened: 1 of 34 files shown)


src/Core/Microsoft.FluentUI.AspNetCore.Components.csproj
```diff
@@ -115,5 +115,8 @@
       <LastGenOutput>%(ParentFile).Designer.cs</LastGenOutput>
     </EmbeddedResource>
   </ItemGroup>
+  <ItemGroup>
+    <Folder Include="Components\Text\" />
```
dvoituron, line 119: Can you remove this line?

src/Core/Components/Text/FluentText.razor.cs
```diff
@@ -0,0 +1,138 @@
...
+    /// Gets or sets the font family of the text.
+    /// </summary>
+    [Parameter]
+    public TextFont? Font { get; set; }
+
+    /// <summary>
+    /// Gets or sets if the texts is set to no-wrap.
+    /// </summary>
+    [Parameter]
+    public bool Nowrap { get; set; }
```
dvoituron, line 58: Can you avoid the "negation" of properties?
And to use "Wrap" like did by [React](https://react.fluentui.dev/?path=/docs/components-text--docs)
(and with default value to `True`)

tests/Core/Microsoft.FluentUI.AspNetCore.Components.Tests.csproj
```diff
@@ -58,4 +58,8 @@
       <DependentUpon>%(ParentFile).razor</DependentUpon>
     </None>
   </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="Components\Text\" />
```
dvoituron, line 63: Can you remove this line?

dvoituron (changes requested):

src/Core/Components/Text/FluentText.razor.cs
vnbaaij: This is the name of the attribute in the web componnets. Not a fan either but think it is better to keep it the same

dvoituron (approved):

PR #2990, opened 2024-11-29
[Icons] Update to Fluent UI System Icons 1.1.266

vnbaaij:
# [Icons Emojis] Split to multiple DLLs in a single NuGet package

Update to Fluent UI System Icons 1.1.266

We were running into the following error when compiling the Icons project: 
> Combined length of user strings used by the program exceeds allowed limit. Try to decrease use of string literals.

Solution we chose to implement was to create a project for each icon variant and change the core Icon project to reference these new projects and include the generated DLLs in the Icons package.

To do this, subclasses have been transformed into namespace (with the same name).
It is also necessary to centralize the `Icons` class (renamed to `IconsExtensions`) in the Components library used by Icons and Emojis.
> **Breaking changes**
>
> - `Icons.GetInstance(icon)` will be `icon.GetInstance()` or `IconsExtension.GetInstance()`
> - `Icons.GetAllIcons()` will be `IconsExtension.GetAllIcons()`
> - `Icons.AllIcons` will be `IconsExtension.AllIcons`


eng/pipelines/build-all-lib.yml
```diff
@@ -5,10 +5,12 @@ trigger: none    # Disable dev and main branches.
 pr: none         # Disable pull request triggers.
 
 parameters:
-  - name: Projects  	    # List of projects to build #**/Microsoft.FluentUI.AspNetCore.Components.Emoji.csproj
+  - name: Projects  	    # List of projects to build 
     default: |
       **/Microsoft.FluentUI.AspNetCore.Components.csproj
       **/Microsoft.FluentUI.AspNetCore.Components.Icons.csproj
```
dvoituron, line 11: I think you need to remove the previous `Icons.csproj` line
And maybe to add the new `ODataAdapter` project?

dvoituron (changes requested):

src/Core/Components/Emojis/EmojiExtensions.cs
```diff
@@ -21,7 +25,7 @@ public static CustomEmoji GetInstance(EmojiInfo emoji)
     {
```
dvoituron, line 25: My bad, I forgot to add `this`. Can you do that?
```csharp
GetInstance(this EmojiInfo emoji)
```

dvoituron (approved):

Issue #2691, opened 2024-09-20
Retrieve IToastService trough IServiceProvider


nullreferencez:
I'm attempting to retrieve the IToastService from a singleton class through IServiceProvider using the code below. However, although I do obtain an instance of IToastService , it does not display the toast.

```c#


  private readonly IServiceProvider _serviceProvider;
  
  public TaskService(IServiceProvider serviceProvider)
  {
          _serviceProvider = serviceProvider;
  }
  
  private async Task ProcessCompletedTask(PendingTask task)
  {
      //other code above
      using (var scope = _serviceProvider.CreateScope())
      {
          var toastService = scope.ServiceProvider.GetRequiredService<IToastService>();
          toastService?.ShowSuccess($"Uploaded and processed {document.Name}");
      }
   }
```

vnbaaij:
HI,

Please supply us with ready-to-run reproduction code in the form of something we can copy/paste, a (zipped) project structure or a GitHub repository.

We do **not** have capacity to craft or compose a reproduction for every issue that gets raised. 

If no code or repository is provided, this issue will be closed in **3** days

Help us to help you. Thanks.

Hwiet:
`IToastService` is injected as a scoped service if you look at the implementation of [AddFluentUIComponents](https://github.com/microsoft/fluentui-blazor/blob/b0f62ea2f11f42548ed4f2ec8821a81f022630f1/src/Core/Extensions/ServiceCollectionExtensions.cs#L18). As is, not cannot be injected into a singleton service.

In addition, from my experience (with FluentUI and even with MudBlazor), I had to re-inject the ToastService as a singleton class in order for it to be injected into a class that is not a Blazor component, even if the class was scoped or transient.

vnbaaij:
Good catch @Hwiet!

Closing this as it is an unsupported scenario.

nullreferencez:
> `IToastService` is injected as a scoped service if you look at the implementation of [AddFluentUIComponents](https://github.com/microsoft/fluentui-blazor/blob/b0f62ea2f11f42548ed4f2ec8821a81f022630f1/src/Core/Extensions/ServiceCollectionExtensions.cs#L18). As is, not cannot be injected into a singleton service.
> 
> In addition, from my experience (with FluentUI and even with MudBlazor), I had to re-inject the ToastService as a singleton class in order for it to be injected into a class that is not a Blazor component, even if the class was scoped or transient.

You are correct brother.

[...]

PR #2991, opened 2024-11-29
[General] Ability to change service lifetime

svrooij:
# Pull Request

## 📖 Description

There are situations where you might want to access the built-in services from a singleton service. This pull requests allows the developer to configure the lifetime if needed. Normal operation is not changed

### 🎫 Issues

* #2691

## 👩‍💻 Reviewer Notes

Normal operation is not changed, only the ability to register the services as Singleton

### General

### Component-specific


dvoituron:
To be clear:
- **AddScoped**: This option creates a new instance of the service for each Blazor circuit. A circuit represents a user connection, so each user gets its own instance of the service. This is useful for services that need to maintain a user-specific state.

- **AddSingleton**: This option creates a single instance of the service that is shared between all users and sessions. This is appropriate for services that do not maintain user state, or for services that need to share data between users.

In general, including for Blazor WebAssembly, it is recommended to use `AddScoped` for services that interact with the user interface or manage user-specific data, as this ensures that each user has his or her own instance of the service. `AddSingleton` can be used for static services or services that share data between users.

I have 3 comments about this PR:

1. Why do you want to change the default `AddFluentUIComponents` method, when **you can** successively call the `services.AddSingleton<IToastService, ToastService>();` methods you want to inject the services of your choice?
   Example, the standard [AddRazorComponents](https://github.com/dotnet/aspnetcore/blob/main/src/Components/Endpoints/src/DependencyInjection/RazorComponentsServiceCollectionExtensions.cs) doesn't contain an argument like that. 

   The same problem will occur if a developer want to inject a single service in Singleton and the others in Scoped. Why inject everything in a single mode?
   
2. I'm afraid that offering this possibility **by default** in the library will lead to developers making mistakes: using a Singleton is not recommended for all services and variable storage in memory. This could cause undesirable side-effects that we can't yet fully detect.

3. If we really want to add such a configuration, I think it would be better to use the `LibraryConfiguration` service configuration object. And why refuse th `Transient`?

svrooij:
I taught that when you have WebAssembly using scoped is useless, does not exists and only stops you from using the services in Singleton services. And it seems this is in the [documentation](https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/dependency-injection?view=aspnetcore-9.0#service-lifetime) as well. You always have the option to do your own service registering, but that will lead to people forgetting new services.

Giving the option to use the default registration while still allowing to change the lifetime seems like a nice way for developers.

Off-course this should **NEVER** be used when using this server side, because there you have a scope per request which makes far more sense.

dvoituron:
> I taught that when you have WebAssembly using scoped is useless, does not exists and only stops you from using the services in Singleton services

Yes and no :-)
Like explained in the [doc you referenced](https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/dependency-injection?view=aspnetcore-9.0#service-lifetime):
> `Scoped`-registered services behave like `Singleton` services.

So, the services injected using `AddScope` will behave like `Singleton` services. But it is possible to inject a Scoped service into a WASM application (which is what we're using in the demo site). And that's no problem. The only difference is that you need to retrieve it as a “Scoped” service (and not Singleton) if you need it later.

dvoituron:
Now I understand. You want to retrieve a FluentUI service from a **Singleton** service. 

It's really a special case. And I don't think we need to manage it from the library. You can inject these services manually to use the Singleton without calling `AdsFluentUIComponents`.

Or via a `ServiceLifetime` property of the `ServiceConfiguration` object.

svrooij:
> You want to retrieve a FluentUI service from a Singleton service.

No I don't want that, did not have a use case for this yet. But I understand the question from #2691 and though I would fix some of the long standing issues.

dvoituron:
By adapting your PR to put this Flag in the **LibraryConfiguration** object, I think it would be better adapted.

dvoituron (approved):

src/Core/Infrastructure/LibraryConfiguration.cs
vnbaaij, line 1: You need to add `using Microsoft.Extensions.DependencyInjection;`
to fix compilation error

vnbaaij (changes requested):

src/Core/Infrastructure/LibraryConfiguration.cs
svrooij: You're right

vnbaaij (approved):

PR #2992, opened 2024-12-01
[Icons Emijis] Split to multiple DLLs in a single NuGet package

dvoituron:
# [Icons Emijis] Split to multiple DLLs in a single NuGet package

We need to split the Icons library into several DLLs. Currently, all icons are in a single file and compiled at once. Unfortunately, .NET and C# have a maximum number of strings available in a single C# project. We therefore decided to split .cs files into several DLLs.

To do this, subclasses must be transformed into namespace (with the same name).
It is also necessary to centralize the `Icons` class (renamed to `IconsExtensions`) in the Components library used by Icons and Emojis.
> **Breaking changes**
>
> - `Icons.GetInstance(icon)` will be `icon.GetInstance()` or `IconsExtension.GetInstance()`
> - `Icons.GetAllIcons()` will be `IconsExtension.GetAllIcons()`
> - `Icons.AllIcons` will be `IconsExtension.AllIcons`

## Build process

To build the NuGet package, we need to build each DLLs using commands like
```bash
dotnet build -c release -p:ICONS_BUILD=Regular -o ./bin/Regular
dotnet build -c release -p:ICONS_BUILD=Filled  -o ./bin/Filled
````

Next a package could be creating using a `csproj` similar to [this one](https://github.com/dotnet/roslyn-analyzers/blob/main/nuget/Microsoft.CodeAnalysis.ResxSourceGenerator/Microsoft.CodeAnalysis.ResxSourceGenerator.Package.csproj).


dvoituron:
Duplicated #2990... Closing

PR #2665, opened 2024-09-16
[Slider] Fix FluentSlider two-way binding issue [#2609}

oneolddev:
# Pull Request

## 📖 Description

Fixes two way binding issue.

### 🎫 Issues

- setting value does not cause the slider thumbnail to redraw

## 👩‍💻 Reviewer Notes

- see comments in #2609 and #2612

### General

- [ ] I have added tests for my changes.
- [X] I have tested my changes.
- [ ] I have updated the project documentation to reflect my changes.
- [X] I have read the [CONTRIBUTING](https://github.com/microsoft/fluentui-blazor/blob/master/docs/contributing.md) documentation and followed the [standards](https://www.fast.design/docs/community/code-of-conduct/#our-standards) for this project.

### Component-specific

- [ ] I have added a new component
- [ ] I have added [Unit Tests](https://github.com/Microsoft/fluentui-blazor/blob/master/unit-tests.md) for my new component
- [X] I have modified an existing component
- [ ] I have validated the [Unit Tests](https://github.com/Microsoft/fluentui-blazor/blob/master/unit-tests.md) for an existing component

PR #3077, opened 2024-12-21
[FluentSlider] [FluentNumberField] Fix #2948

### LINKED ISSUE #2609: [Bug] FluentSlider two-way binding issue

### LINKED ISSUE #2948: fix: desync of FluentNumberField and FluentSlider in 4.10.4
# 🐛 Bug Report

Seeming regressions of #1531 and #1450 

There are 2 Problems

### FluentNumberField

This test demonstrates that the FluentNumberField's two-way data binding is broken in version 4.10.4. This appears to be a regression.

In my web app, I have a component that takes a numeric input and divies a portion to a wallet and a portion to a stash. After a number is selected, a button triggers a command and this command also resets the value of the numeric field to 0.

Unfortunately, the numeric field does not respect the behavior of being reset to 0 by the button command.

Go ahead and select a number like 2, which will enable the reset button. The expected behavior is for the reset button to set the value back to 0, and for the number input field to reflect that value.

You can see in the console logs that the data model's value is reset to 0, however the number field's value is still non-zero both visibly and under the hood. This implies that it isn't respecting the binding.

### FluentSlider

This test demonstrates that the FluentSlider's two-way data binding can desync from its bound value in version 4.10.4. This may be a regression.

While you can see the behavior working for the Reset button, the behavior does not always work for the "Change by 2" button.

Go ahead and set the slider to 3, which will enable the reset button (only enabled at 3). Reset. You see that it works.

Go ahead and set the slider to 1, then keep pressing the "Change by 2" button. In addition to the visual weirdness, by checking the console logs, you will see that it desyncs from the actual value in the data.

When I run it, the console logs always have an accurate data value for the "Change by 2" button's usage. The underlying slider value on blazor side is one step behind. And the visual only updates every 3 clicks of the button.

A second test scenario is: 1. Refresh the page. 2. Set the slider to 1. 3. Click "Change by 2". 4. Click Reset. This will break the visual synchronization with the "Reset" button, which is otherwise usually working correctly.

## 💻 Repro or Code Sample

https://github.com/arazni/blazor-fluent-ui-slider-bug/tree/main

## 🤔 Expected Behavior

When changing a 2-way bound value outside of the component, the visual and component's value should stay synchronized.

## 😯 Current Behavior

There are scenarios where the component does not match the underlying data. See the repro steps on the repo.

## 🔦 Context

Same as  #1531 and #1450, I have buttons that change values nearby fields and sliders that take those changes. See https://arazni.github.io/blades-in-the-sheets/demo for context. You may need to enable sliders over checkboxes in the accessibility tab. Then go back to the demo page, The fund section with the pencil icon enabled is where some of these issues come up.

## 🌍 Your Environment

* At least Windows
* At least Mozilla FireFox
* .NET 9 and Fluent UI Blazor library Version 4.10.4

oneolddev:
# Pull Request

## 📖 Description

Fixes issues reported in #2948. Two issue reported both related to changing the value of the control programatically.

### 🎫 Issues

The underlying issue is that the web components being wrapped were designed to have there states only changed interactively.  `bind-value` does not behave as expected. Inward binding works only on for initialization. Outward binding works as expected.

This gives the appearance of a *sync* issue.

## 👩‍💻 Reviewer Notes

`FluentSlider` - #2665 did not fully fix #2609. The original fix was to prevent the [flickering when the slider was moved rapidly](https://github.com/microsoft/fluentui-blazor/issues/2609#issuecomment-2326339496) originally reported. I was not able to reproduce this problem with the current implementation of `FluentDataGrid`. I did end up adding debouncing with the `DebounceAction`.  I believe I did this correctly.

`FluentNumberField` - The fix was to ensure the underlying web component reflected the correct value. I could not find a way of distinguishing if the value was changed by manipulating the control or if it was changed programatically. I opted to use some javascript to force the value in the web component. I also added a `DisposeAsync`. It was missing.

I've also included an `issue-tester` page as part of this draft

### General

### Component-specific

### DIFF
--- a/src/Core/Components/NumberField/FluentNumberField.razor.cs
+++ b/src/Core/Components/NumberField/FluentNumberField.razor.cs
@@ -160,6 +160,17 @@ public partial class FluentNumberField<TValue> : FluentInputBase<TValue>, IAsync
         base.OnParametersSet();
     }
 
+    /// <summary>
+    /// set to true when value is changed by the fluent-number-field either by input or by the spin buttons.
+    /// </summary>
+    private bool inputChanged = false;
+
+    protected override Task InputHandlerAsync(ChangeEventArgs e)
+    {
+        inputChanged = true;
+        return base.InputHandlerAsync(e);
+    }
+
     protected override async Task OnAfterRenderAsync(bool firstRender)
     {
         await base.OnAfterRenderAsync(firstRender);
@@ -175,8 +186,15 @@ public partial class FluentNumberField<TValue> : FluentInputBase<TValue>, IAsync
         }
         else
         {
-            Module ??= await JSRuntime.InvokeAsync<IJSObjectReference>("import", JAVASCRIPT_FILE.FormatCollocatedUrl(LibraryConfiguration));
-            await Module.InvokeVoidAsync("setNumberFieldValue", Element, Value!.ToString());
+            if (inputChanged)
+            {
+                inputChanged = false;
+            }
+            else
+            {
+                Module ??= await JSRuntime.InvokeAsync<IJSObjectReference>("import", JAVASCRIPT_FILE.FormatCollocatedUrl(LibraryConfiguration));
+                await Module.InvokeVoidAsync("setNumberFieldValue", Element, Value!.ToString());
+            }
         }
     }
 
--- a/src/Core/Components/NumberField/FluentNumberField.razor.cs
+++ b/src/Core/Components/NumberField/FluentNumberField.razor.cs
@@ -160,41 +160,20 @@ public partial class FluentNumberField<TValue> : FluentInputBase<TValue>, IAsync
         base.OnParametersSet();
     }
 
-    /// <summary>
-    /// set to true when value is changed by the fluent-number-field either by input or by the spin buttons.
-    /// </summary>
-    private bool inputChanged = false;
-
-    protected override Task InputHandlerAsync(ChangeEventArgs e)
-    {
-        inputChanged = true;
-        return base.InputHandlerAsync(e);
-    }
-
     protected override async Task OnAfterRenderAsync(bool firstRender)
     {
         await base.OnAfterRenderAsync(firstRender);
 
         if (firstRender)
         {
+            Module ??= await JSRuntime.InvokeAsync<IJSObjectReference>("import", JAVASCRIPT_FILE.FormatCollocatedUrl(LibraryConfiguration));
+            await Module.InvokeVoidAsync("ensureCurrentValueMatch", Element);
+
             if (AutoComplete != null && !string.IsNullOrEmpty(Id))
             {
-                Module ??= await JSRuntime.InvokeAsync<IJSObjectReference>("import", JAVASCRIPT_FILE.FormatCollocatedUrl(LibraryConfiguration));
                 await Module.InvokeVoidAsync("setControlAttribute", Id, "autocomplete", AutoComplete);
             }
 
-        }
-        else
-        {
-            if (inputChanged)
-            {
-                inputChanged = false;
-            }
-            else
-            {
-                Module ??= await JSRuntime.InvokeAsync<IJSObjectReference>("import", JAVASCRIPT_FILE.FormatCollocatedUrl(LibraryConfiguration));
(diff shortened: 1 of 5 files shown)


examples/Demo/Shared/Pages/Lab/IssueTesterPage.razor
vnbaaij, line 1: For now this is fine but we have the  lssueTesterPage as the 'container' (so you do not need to mess with the @page and title, etc.) and the IssueTester as the actual pungent to put the code in.

vnbaaij (commented):
Will test this soop

examples/Demo/Shared/Pages/Lab/IssueTesterPage.razor
oneolddev: I left it in for testing purposes only. was planning to remove it 😄

oneolddev:
@vnbaaij 

I've fixed the unit test by easing the BUnit JSInterop mode. I'm not sure if that breaks any testing conventions.

Should this be split into two PRs? One for `FluentSlider` and one for `FluentNumberField`.

oneolddev:
Found another issue with my fix for `FluentNumberField` 😞

src/Core/Components/NumberField/FluentNumberField.razor.cs
```diff
@@ -173,5 +184,34 @@ protected override async Task OnAfterRenderAsync(bool firstRender)
             }
 
         }
+        else
```
dvoituron, line 187: I don't like calling a JS method every time a render is raised :-(

Why not use a JS Observable to apply the `value` attribute to the `value` JS property?
Like this: https://github.com/microsoft/fluentui-blazor/blob/dev-v5/src/Core/Components/TextInput/FluentTextInput.razor.ts

dvoituron (changes requested):

src/Core/Components/NumberField/FluentNumberField.razor.cs
oneolddev: I'll take a look at that approach 🙏

oneolddev: @dvoituron 

This was a great suggestion 😄.  It's definitely better than my original approach.

dvoituron (approved):
Next time, please create one PR to fix one problem. This makes it easier to maintain the code.
Thank you :-)

vnbaaij (approved):

PR #3105, opened 2024-12-27
[ListComponentBase] Fix #3029 invoke `ValueChanged` when initially selecting an option

### LINKED ISSUE #3029: fix: Page not responding in combobox from a list of Option<T> items
# 🐛 Bug Report

If you enter a free text value into a combobox from a list of Option<T> items the page will become unresponsive.

## 💻 Repro or Code Sample

On the demo site, change the text in the combobox to one that does not exist, and then click out of the combobox. The page will become unresponsive
https://www.fluentui-blazor.net/Combobox#fromalistofoptiontit
Image

## 🤔 Expected Behavior

The page should not become unresponsive. I guess that it should either accept the inputted value as plaintext or reset to the previous value?

## 😯 Current Behavior

The page becomes unresponsive and requires killing by the browser

## 💁 Possible Solution

My guess would be that it cannot find the option which matches the input and goes into some kind of infinite loop

## 🔦 Context

I was trying to find an autocomplete-like component which only allows a single selection and has an async search function. I was trying out the behaviour of a few other components to see if any other worked well for this.

## 🌍 Your Environment

* OS & Device: Windows 11
* Browser Firefox 133.0
* .NET and Fluent UI Blazor library Version: No idea, whatever the demo site is using

pk9r:
# Pull Request

## 📖 Description

Fix 3029, Combobox no longer causes the page to become unresponsive.
Image

https://github.com/microsoft/fluentui-blazor/issues/3029#issuecomment-2563476494

### 🎫 Issues

#3029 

### General

### Component-specific

## ⏭ Next Steps

The displayed value of the combobox is not consistent with its value. I have no idea how to solve this.
peek_2

### DIFF
--- a/src/Core/Components/List/ListComponentBase.razor.cs
+++ b/src/Core/Components/List/ListComponentBase.razor.cs
@@ -364,8 +364,10 @@ public abstract partial class ListComponentBase<TOption> : FluentInputBase<strin
         {
             if (SelectedOption == null && Items != null && OptionSelected != null)
             {
-                TOption? item = Items.FirstOrDefault(i => OptionSelected.Invoke(i));
-                InternalValue = GetOptionValue(item);
+                var item = Items.FirstOrDefault(i => OptionSelected.Invoke(i));
+                var value = GetOptionValue(item);
+                InternalValue = value;
+                ValueChanged.InvokeAsync(value);
             }
         }
 
--- a/src/Core/Components/List/ListComponentBase.razor.cs
+++ b/src/Core/Components/List/ListComponentBase.razor.cs
@@ -367,10 +367,12 @@ public abstract partial class ListComponentBase<TOption> : FluentInputBase<strin
                 var item = Items.FirstOrDefault(i => OptionSelected.Invoke(i));
                 var value = GetOptionValue(item);
                 InternalValue = value;
-                ValueChanged.InvokeAsync(value);
+                if (value != Value && ValueChanged.HasDelegate)
+                {
+                    ValueChanged.InvokeAsync(value);
+                }
             }
         }
-
     }
 
     /// <inheritdoc />


src/Core/Components/List/ListComponentBase.razor.cs
```diff
@@ -364,8 +364,10 @@ protected override void OnParametersSet()
         {
             if (SelectedOption == null && Items != null && OptionSelected != null)
             {
-                TOption? item = Items.FirstOrDefault(i => OptionSelected.Invoke(i));
-                InternalValue = GetOptionValue(item);
+                var item = Items.FirstOrDefault(i => OptionSelected.Invoke(i));
+                var value = GetOptionValue(item);
+                InternalValue = value;
```
dvoituron, line 369: Please note that this `ListComponentBase` class is used by all lists. A change here will affect all other lists, not just the **Combobox**. Have you checked that other components are not affected?

src/Core/Components/List/ListComponentBase.razor.cs
```diff
@@ -364,8 +364,10 @@ protected override void OnParametersSet()
...
+                ValueChanged.InvokeAsync(value);
```
dvoituron, line 370: I think you should check that the `value != Value` before raising the `ValueChanged`.
In addition, you should check that the user has requested to be subscribed via `if (ValueChanged.HasDelegate)`.

dvoituron (changes requested):

src/Core/Components/List/ListComponentBase.razor.cs
pk9r: I checked other components including Fluent Select, Fluent Autocomplete, Fluent Listbox. The examples on the demo page still work as before.

dvoituron (approved):

vnbaaij (approved):

PR #3112, opened 2024-12-30
[dev-v5] [Dialog Service] Add the FluentDialog and services

dvoituron:
# [dev-v5] [Dialog Service] Add the FluentDialog and services

## Dialog
**FluentDialog** is a window overlaid on either the primary window or another dialog window. Windows under a modal dialog are inert. That is, users cannot interact with content outside an active dialog window. Inert content outside an active dialog is typically visually obscured or dimmed so it is difficult to discern, and in some implementations, attempts to interact with the inert content cause the dialog to close

## Usage
The simplest way is to use the **DialogService** to display a dialog box. By injecting this service, you have **ShowDialogAsync** methods at your disposal. You can pass the type of Dialog to be displayed and the options to be passed to that window.

Once the user closes the dialog window, the **ShowDialogAsync** method returns a **DialogResult** object containing the return data.

> Any Blazor component can be used as a dialog type.

```csharp

var result = await DialogService.ShowDialogAsync<SimpleDialog>(options =>
{
    // Options
});

if (result.Cancelled == false)
{
    ...
}
```

The **SimpleDialog** window can inherit from the **FluentDialogInstance** class to have access to the **DialogInstance** property, which contains all parameters defined when calling **ShowDialogAsync**, as well as the **CloseAsync** and **CancelAsync** methods. Using **CloseAsync** you can pass a return object to the parent component. It is then retrieved in the **DialogResult** like in the example below.

```xml
@inherits FluentDialogInstance

<FluentDialogBody>
    Content dialog
</FluentDialogBody>
```

## Data exchange between components

You can easily send data from your main component to the dialog box by using the **options.Parameters.Add()** method in the **ShowDialogAsync** method options.

Parameters is a dictionary that allows you to send any type of data to the dialog box. the key is the name of the property in the dialog box, and the value is the data to be sent.

```csharp
PersonDetails John = new() { Age = "20" };

var result = await DialogService.ShowDialogAsync<CustomizedDialog>(options =>
{
    options.Parameters.Add(nameof(SimpleDialog.Name), "John");  // Simple type
    options.Parameters.Add(nameof(SimpleDialog.Person), John);  // Updatable object
});
```

## MessageBox

A **MessageBox** is a dialog that is used to display information with a specific intent to the user. It uses the **DialogService** to display the dialog.

The **DialogService** is an injected service that can be used into any component. It exposes methods to show a dialog. For working with a **MessageBox**, the following methods are available:

- **ShowSuccessAsync**: Shows a dialog with a success (green) icon, a message and an OK button.

- **ShowWarningAsync**: Shows a dialog with a warning (orange) icon, a message and an OK button.

- **ShowInfoAsync**: Shows a dialog with an information (gray) icon, a message and an OK button.

- **ShowErrorAsync**: Shows a dialog with an error (red) icon, a message and an OK button.

- **ShowConfirmationAsync**: Shows a dialog with a confirmation icon, a message and a Yes/No buttons. This method returns a DialogResult object where Cancelled = false if Yes is pressed and Cancelled = true if No is pressed.

- **ShowMessageBoxAsync**: Shows a dialog with the specified options.

## Live Demo

See the [Dialog](https://fluentui-blazor-v5.azurewebsites.net/Dialog) page,
and the [MessageBox](https://fluentui-blazor-v5.azurewebsites.net/MessageBox) page.

## Unit Tests

{BE0299BF-30F2-4201-8462-1EC89C04F7F3}


examples/Demo/FluentUI.Demo.Client/Documentation/Components/Dialog/Examples/DialogServiceCustomized.razor
```diff
@@ -0,0 +1,41 @@
+﻿@inject IDialogService DialogService
+
+<FluentButton OnClick="@(e => OpenDialogAsync(DialogAlignment.End))" Appearance="ButtonAppearance.Primary">Open Panel</FluentButton>
```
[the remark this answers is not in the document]
dvoituron: In our library, a dialog can be opened left/right (Panel).
But I agree with you that we could delete the example, retaining this functionality... and in a new PR we could “replace” the Panel using the new `fluent-drawer`.

Is that what your comment means?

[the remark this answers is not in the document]
vnbaaij: Yes, that is what I mean

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Dialog/FluentDialog.md
```diff
@@ -0,0 +1,214 @@
+---
+title: Dialog
+route: /Dialog
+---
+
+# Dialog
+
+`FluentDialog` is a window overlaid on either the primary window or another dialog window.
```
vnbaaij, line 8: This is a bit in contradiction with line 24 below. Maybe use the text from fluent2 site:

A dialog is a supplemental surface that can provide helpful interactions or require someone to take an action before they can continue their task, like confirming a deletion.

Dialogs are often interruptions, so use them for important actions. If you need to give someone an update on an action they just took but that they don't need to act on, try a toast.

dvoituron: I copied the text from https://web-components.fluentui.dev/?path=/docs/components-dialog-dialog-body--docs
But make sense. I will change that.

src/Core.Scripts/src/Components/Dialog/FluentDialog.ts
```diff
@@ -0,0 +1,20 @@
...
+   */
+  export function Show(id: string): void {
+    const dialog = document.getElementById(id) as any;
+    dialog?.show();
+  }
+
+  /**
+ * Display the fluent-dialog with the given id
+ * @param id The id of the fluent-dialog to display
+ */
```
vnbaaij, line 15: Hide instead of display

src/Core/Components/Base/FluentServiceProviderException.cs
```diff
@@ -0,0 +1,20 @@
...
+/// <summary>
+/// Exception thrown when a service provider is not available.
+/// </summary>
+public class FluentServiceProviderException<TProvider> : Exception
+{
+    /// <summary>
+    /// Creates a new instance of the <see cref="FluentServiceProviderException{TProvider}"/> class.
+    /// </summary>
+    public FluentServiceProviderException()
+        : base($"{typeof(TProvider).Name} needs to be added to the main layout of your application/site.")
```
vnbaaij, line 16: Not every service is added to MainLayout only. Should we make it a bit more general? Maybe something like
"$"{typeof(TProvider).Name} needs to be added to the page/component hierarchy of your application/site. Usually this will be 'MainLayout' but depending on your setup it could be at a different location."

tests/Core/Components/Dialog/Data/ShortCutsData.cs
```diff
@@ -0,0 +1,98 @@
...
+// MIT License - Copyright (c) Microsoft Corporation. All rights reserved.
+// ------------------------------------------------------------------------
+
+using System.Collections;
+using Microsoft.AspNetCore.Components.Web;
+using Microsoft.FluentUI.AspNetCore.Components.Tests.Components.Dialog.Templates;
+
+namespace Microsoft.FluentUI.AspNetCore.Components.Tests.Components.Dialog.Data;
+
+public class ShortcutsData : IEnumerable<object[]>
```
vnbaaij, line 11: Think 'ShortcutData' (singular) is better

tests/Core/Components/Dialog/Data/ShortCutsData.cs
```diff
@@ -0,0 +1,98 @@
...
+        {
+            Key = key,
+            CtrlKey = ctrlKey ?? false,
+            ShiftKey = shiftKey ?? false,
+            AltKey = altKey ?? false
+        };
+    }
+}
+
+public record ShortcutsDataItem(
```
vnbaaij, line 91: Also use singular here

vnbaaij (changes requested):
Approved once comments are resolved (if I approve now, it will get merged immediately)

vnbaaij (approved):

PR #3174, opened 2025-01-15
Fix FluentCombobox option value

denispaluca:
# Pull Request

## 📖 Description

When using `FluentCombobox` and providing the parameters `OptionText` and `OptionValue`, the `Value` of the combobox is derived from `OptionText` and not `OptionValue`.

The expected behavior would be, that for the `Value` of the `FluentCombobox`, `OptionValue` would be prioritized when both `OptionText` and `OptionValue` are provided.

Fixed it by changing the order in method `GetOptionValue`.

### General

### Component-specific


vnbaaij:
This is by design. People are free to change the option text in a combobox so we need to look at that first

denispaluca:
@vnbaaij Is there no way to limit the user so that only one of the given values (one of `Items) is valid? What I want is a Select with search/autocomplete.

vnbaaij:
> @vnbaaij Is there no way to limit the user so that only one of the given values (one of `Items) is valid? 

No, with a combobox that is not possible. By default in a combobox a user can **always** input his/her own text.

> What I want is a Select with search/autocomplete.

Sounds to me like you want https://www.fluentui-blazor.net/Autocomplete with a MaximumSelectedOptions set to 1

denispaluca:
@vnbaaij Ok, thanks for the help 💯

PR #3248, opened 2025-01-26
[dev-v5] FluentStack

dvoituron:
# [dev-v5] FluentStack

Add identical component created in v4.

### Updated properties
  `HorizontalGap` and `VerticalGap` has been update to a `string?` type.   In v4, the type was `int?` (in pixels).
  This allows the use of other units proposed by the CSS standard.
  If this value is an integer, the unit `px` is automatically added.

  This should not generate any breaking changes.

### New properties
  `Height` is a new property.

## Unit Tests

image


dvoituron:
Do we need to rename de two css classes to `fluent-stack-horizontal` and `fluent-stack-vertical` ? To avoid conflict with potential other frameworks.

vnbaaij:
If it is not scoped anymore, yes then we should add a prefix

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Slack/FluentSlack.md
```diff
@@ -0,0 +1,73 @@
...
+  and the `VerticalGap` has been set to `20 pixels`.
+- The second, nested, `FluentStack` has its `Orientation` parameter
+  set to `Orientation.Horizontal`. The `HorizontalGap` has been set
+  to `4 pixels`. Its first element contains a forced break.
+  No height has been set, so the container height adjusts to the height
+  of the highest element.
+
+The alignment of the contents of both `FluentStack` can be changed
+by selecting one the different options from each of the lists
+(which are placed in a FluentStack themselves as well).
```
vnbaaij, line 63: (which themselves are placed and positioned using a `FluentStack` as well).

vnbaaij (changes requested):

vnbaaij (approved):

PR #3267, opened 2025-01-29
[dev-v5] Add FluentField component

dvoituron:
# [dev-v5] Add `FluentField` component

Field adds a label, validation message, and hint text to a control.
> ⚠️This component will be used with all Fluent Input components.

## Label

The `Label` associated with the field.
This text is usually displayed above the component to describe it to the user.
You can position it on the left or right using `LabelPosition`.
In this case, the attribute `LabelWidth` can be used to set an identical label width
to multiple `FluentField`.

This label can be fully customized (bold, italic, icons, ...) using `LabelTemplate`.

## Validation Message

The `Message` is used to give the user feedback about the value entered.
The field can validate itself using the rules defined in `MessageCondition`.
It can be used to signal the result of form validation.

This message can be fully customized (bold, italic, icons, ...) using `MessageTemplate`.

The `MessageState` affects the behavior and appearance of the message:
- `Error` - The validation message has the red text `An error occured` with a red error icon.
- `Success` - The validation message has gray text `Valid data` with a green checkmark icon.
- `Warning` - The validation message has gray text `Please, check this value` with a yellow exclamation icon.
- `<null>` - The validation message must be defined using `Message` or `MessageTemplate`.

These default messages can be located using the [localization](/localization) service.

Optionally, `Message` and `MessageIcon` attributes can be used to override
the default text and icon (or add an icon in the case of `MessageState = null`).

## Message Conditions

The `MessageCondition` attribute specifies when to display the message
at the bottom of the `FluentField`.
This condition is a lambda expression (or method) whose input parameter
is the current `IFluentField` and output is `true` to display the message
or `false` to hide the message.

Examples: `@(field => field.FocusLost == true)` will display the message
when the user will leave the component.

To help you implement these rules, you can use the `IFluentField.Where`
method associated with the `Display` method. You must finalize the rule
definition using the method `Build()` (to start the validation of the rules).

Rules are evaluated in order of creation: the first rule checked applies
the associated `Display` action

```csharp
field => field.When(() => MyValue.Length <= 1)
                   .Display("Less than 1", MessageState.Success)
              .When(() => MyValue.Length <= 3)
                   .Display("Less than 3", MessageState.Error)
              .When(() => MyValue.Length <= 5)
                   .Display("Less than 5", new Icons.Regular.Size16.Image())
              .When(() => true)
                   .Display(MessageState.Warning)
              .Build())
```

## Usage in other Input components

Example for the **FluentTextInput**:
```razor
<FluentField InputComponent="@this" ForId="@Id" Class="@ClassValue" Style="@StyleValue">
    <fluent-text-input ... />
</FluentField>
```

Using `InputComponent=‘@this’` allows the **FluentField** to retrieve all the parameters of `IFluentField`, like
`Label`, `LabelTemplate`, `LabelPosition`, `LabelWidth`, `Required`, `Disabled`, `Message`, `MessageIcon`, `MessageTemplate`, and `MessageCondition`.

peek_4

## Examples

See the documentation and live samples: https://fluentui-blazor-v5.azurewebsites.net/Field

## Unit Tests

image


src/Core/Enums/LabelPosition.cs
vnbaaij, line 1: Should we prefix the enum with `Field`. I can imagine this being used in other places as well but with different values then (we've seen that before)

src/Core/Enums/MessageState.cs
vnbaaij, line 1: Think `MessageIntent` is a better name. We can re-use v4 enum values (also info and custom)

vnbaaij (changes requested):

src/Core/Enums/LabelPosition.cs
dvoituron: No actual **Enumeration** are prefixed. I don't think that could be a problem because we are using the namespace `Microsoft.FluentUI.AspNetCore.Components` to avoid conflicts with other frameworks.

image

src/Core/Enums/MessageState.cs
dvoituron: I used the suffix `State` from React: https://react.fluentui.dev/?path=/docs/components-field--docs#validation-message
> The **validationState** affects the behavior and appearance of the message.

src/Core/Enums/LabelPosition.cs
vnbaaij: I mean being reused within the library. As we have now:
image

dvoituron: Oh I did not understand :-) Yes, we can have that... so `FieldLabelPosition`?

vnbaaij: Exactly!

vnbaaij (approved):

PR #3284, opened 2025-02-01
[Tabs] Fix Overflow getting out of sync with tab collection

### LINKED ISSUE #3268: fix: TabsOverflow not being properly disposed when List<items> is changed in FluentTabs
# 🐛 Bug Report

I have a component that builds makes a tab for each item in a list.  When the user filters the list down to a subset of items the number of tabs is correctly updated in the UI but it leaves behind an orphaned badge of overflow items that previously existed.

## 💻 Repro or Code Sample

```    
<FluentSwitch @bind-Value=hideInactive >Hide Inactive Items</FluentSwitch>
<FluentTabs @bind-ActiveTabId="@activeid" OnTabChange="HandleOnTabChange" Data="@Items">
@foreach (var item in Items)
{
     <FluentTab Id="@($"tab-{item.Name}")" @bind-Label="@item.Name" Data="@item" > item.Content </FluentTab>
}
</FluentTabs>

@code 
{
     private bool hideInactive;
     List<Items> => Projects.Where(n => n.Active == hideInactive).ToList();
}
```

## 🤔 Expected Behavior

If the number of items in the list changed such that there is no longer an overflow, those items should be properly disposed of from TabsOverflow.

## 😯 Current Behavior

When I change the slider to hide inactive projects, the number of tabs is reduced to the expected amount.  If there was an overflow prior to reducing the list of items then the overflow badge remains along with any tab-ids that were in the overflow prior to the item list change.

## 🌍 Your Environment

* OS & Device: Windows 11 on  PC
* Browser: Microsoft Edge
* .NET SDK Version: 9.0.101
* Fluent UI Blazor library Version: 4.11.3
* Project Target: net8.0

vnbaaij:
Fix #3268

Replace Dictionary of tabs with a List and fix overflow keeping closed/unregistered items. Thanks @rpodevns!

### DIFF
--- a/src/Core/Components/Tabs/FluentTab.razor.cs
+++ b/src/Core/Components/Tabs/FluentTab.razor.cs
@@ -1,3 +1,7 @@
+// ------------------------------------------------------------------------
+// MIT License - Copyright (c) Microsoft Corporation. All rights reserved.
+// ------------------------------------------------------------------------
+
 using Microsoft.AspNetCore.Components;
 using Microsoft.FluentUI.AspNetCore.Components.Extensions;
 using Microsoft.FluentUI.AspNetCore.Components.Utilities;
@@ -148,7 +152,7 @@ public partial class FluentTab : FluentComponentBase
     /// <summary />
     protected virtual Task CloseClickedAsync()
     {
-        return Owner!.UnregisterTabAsync(this);
+        return Owner!.UnregisterTabAsync(Id!);
     }
 
     /// <summary />
--- a/src/Core/Components/Tabs/FluentTabs.razor.cs
+++ b/src/Core/Components/Tabs/FluentTabs.razor.cs
@@ -2,12 +2,12 @@
 // MIT License - Copyright (c) Microsoft Corporation. All rights reserved.
 // ------------------------------------------------------------------------
 
+using System.Diagnostics.CodeAnalysis;
+using System.Text.Json;
 using Microsoft.AspNetCore.Components;
 using Microsoft.FluentUI.AspNetCore.Components.Extensions;
 using Microsoft.FluentUI.AspNetCore.Components.Utilities;
 using Microsoft.JSInterop;
-using System.Diagnostics.CodeAnalysis;
-using System.Text.Json;
 
 namespace Microsoft.FluentUI.AspNetCore.Components;
 
@@ -16,7 +16,7 @@ public partial class FluentTabs : FluentComponentBase
     private const string JAVASCRIPT_FILE = "./_content/Microsoft.FluentUI.AspNetCore.Components/Components/Overflow/FluentOverflow.razor.js";
 
     private const string FLUENT_TAB_TAG = "fluent-tab";
-    private readonly Dictionary<string, FluentTab> _tabs = [];
+    private readonly List<FluentTab> _tabs = [];
     //private string _activeId = string.Empty;
     private DotNetObjectReference<FluentTabs>? _dotNetHelper = null;
     private IJSObjectReference _jsModuleOverflow = default!;
@@ -98,7 +98,7 @@ public partial class FluentTabs : FluentComponentBase
     /// <summary>
     /// Gets the active selected tab.
     /// </summary>
-    public FluentTab ActiveTab => _tabs.FirstOrDefault(i => i.Key == ActiveTabId).Value ?? _tabs.First().Value;
+    public FluentTab ActiveTab => _tabs.FirstOrDefault(t => t.Id == ActiveTabId) ?? _tabs.First();
 
     [Parameter]
     public string ActiveTabId { get; set; } = default!;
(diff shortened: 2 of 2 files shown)


src/Core/Components/Tabs/FluentTabs.razor.cs
```diff
@@ -154,41 +154,44 @@ protected override async Task OnAfterRenderAsync(bool firstRender)
...
     {
         var tabId = args?.ActiveId;
-        if (tabId is not null && _tabs.TryGetValue(tabId, out FluentTab? tab))
+        var tab = _tabs.FirstOrDefault(i => i.Id == tabId);
+
+        if (tab is not null && _tabs.Contains(tab))
         {
             await OnTabChange.InvokeAsync(tab);
-            ActiveTabId = tabId;
+            ActiveTabId = tabId!;
```
dvoituron, line 169: You should test if `tabId` is null before this code. If the compiler informs you that it's possible to have a null value, you should test it.

src/Core/Components/Tabs/FluentTabs.razor.cs
```diff
@@ -154,41 +154,44 @@ protected override async Task OnAfterRenderAsync(bool firstRender)
...
 
     internal int RegisterTab(FluentTab tab)
     {
-        _ = _tabs.TryAdd(tab.Id!, tab);
+        _tabs.Add(tab);
         return _tabs.Count - 1;
     }
 
-    internal async Task UnregisterTabAsync(FluentTab tab)
+    internal async Task UnregisterTabAsync(string id)
```
dvoituron, line 180: Why did you change this signature ? Using the Tab object could be more upgradable. And the Tab ID can be set to `null`.

src/Core/Components/Tabs/FluentTabs.razor.cs
```diff
@@ -154,41 +154,44 @@ protected override async Task OnAfterRenderAsync(bool firstRender)
...
         if (OnTabClose.HasDelegate)
         {
+            var tab = _tabs.FirstOrDefault(t => t.Id == id);
             await OnTabClose.InvokeAsync(tab);
         }
 
         if (_tabs.Count > 0)
         {
-            _tabs.Remove(tab.Id!);
+            _tabs.RemoveAt(_tabs.Count - 1);
```
dvoituron, line 190: Why to remove the last one and not the Tab ID?

src/Core/Components/Tabs/FluentTab.razor.cs
```diff
@@ -148,7 +152,7 @@ protected override async Task OnAfterRenderAsync(bool firstRender)
     /// <summary />
     protected virtual Task CloseClickedAsync()
     {
-        return Owner!.UnregisterTabAsync(this);
+        return Owner!.UnregisterTabAsync(Id!);
```
dvoituron, line 155: Same as my other comment : `Id` is nullable.

dvoituron (changes requested):

src/Core/Components/Tabs/FluentTabs.razor.cs
dvoituron: > Blazor is recycling HashCodes so when a FluentTab is removed, the next tab is assigned the same HashCode (and memory pointer?) and the UnregisterTabAsync method starts acting on the wrong tab.

Blazor itself does not recycle hash codes.

In my opinion, the problem probably comes from somewhere else, where the dictionary reorganization is not adapted to this component.

I understand the advantage of using an ID that avoids passing the wrong object reference. On the other hand, it may trigger another problem, as `Id` is nullable, so it's possible to have no identifier.

vnbaaij: Id is indeed nullable but in our use cases it can't be null as it is assigned a value in the FluentTab constructor.

src/Core/Components/Tabs/FluentTabs.razor.cs
dvoituron: This doesn't explain why calling the “UnregisterTabAsync(string id)” method removes the last item from the tab list :-)

src/Core/Components/Tabs/FluentTabs.razor.cs
vnbaaij: Yes, will add a check (although it can't be null in our situation)

src/Core/Components/Tabs/FluentTab.razor.cs
vnbaaij: Same comment for me as well 😁

src/Core/Components/Tabs/FluentTabs.razor.cs
dvoituron: Even if the constructor assigns a default value to ID, nothing prevents the developer from defining `Id=“@null”` :-)

vnbaaij: > Even if the constructor assigns a default value to ID, nothing prevents the developer from defining `Id=“@null”` :-)

That is why I said 'in our use cases' 😉 Developers will always get themselves in trouble if we let them.

vnbaaij: As discussed internally. We'll go for the quick fix now. We can revisit when we work on the v5 component

src/Core/Components/Tabs/FluentTabs.razor.cs
vnbaaij: We've confirmed that for this fix, it is doing the right thing

dvoituron (approved):

PR #3294, opened 2025-02-03
[Lists] Trigger field changed when in multiple mode

### LINKED ISSUE #3115: feat: add Validation to/in Autocomplete control
# 🙋 Feature Request

Using the Autocomplete control, we are dynamically setting required to "true" when a different control is populated.
This properly sets the requirement; you cannot submit the form when the control is empty. However, the field remains required and will not validate successfully, even after you select a value. It would be nice for the validation to properly accept a value and allow for submission.

This was [reported as a Bug](https://github.com/microsoft/fluentui-blazor/issues/3025), but marked 'working as designed'. So I'm converting it to a feature request for future consideration.

## 🤔 Expected Behavior

When we select a value in the autocomplete control, it should no longer say that the field is required, and allow the form to submit.

## 😯 Current Behavior

Field fails validation, even after a selection is made.
Image

## 🔦 Context

We would like to have a an autocomplete control be required when another field is selected. The control should require a selection and validate properly.

## 💻 Examples
Examples provided in referenced Bug ticket.
Image

joriverm:
# Pull Request

## 📖 Description

autocomplete doesnt trigger the fieldchanged as the default ChangeHandler, which handles the validation/fieldchanged event of the editform, this was fixed by always triggering the event on selection changes for both single and multiple mode.

i also noticed that the list base didnt implement the binding expression property, so the fieldidentifier was always wrong.

### 🎫 Issues
Fix #3115 

## 👩‍💻 Reviewer Notes

this was tested using the form page from the demo project. i am currently going through the tests to check out if i missed anything

## 📑 Test Plan

none, i dont see how to put this in a test without adding a binding the bunit test and rendering a validator at the same time. not sure if its worth it?

### General

### Component-specific

## ⏭ Next Steps
N/A

### DIFF
--- a/src/Core/Components/List/ListComponentBase.razor.cs
+++ b/src/Core/Components/List/ListComponentBase.razor.cs
@@ -527,19 +527,8 @@ public abstract partial class ListComponentBase<TOption> : FluentInputBase<strin
         {
             if (!Equals(item, SelectedOption))
             {
-                var value = GetOptionValue(item);
-
-                if (this is FluentListbox<TOption> ||
-                    this is FluentCombobox<TOption> ||
-                    (this is FluentSelect<TOption> && Value is null))
-                {
-                    await base.ChangeHandlerAsync(new ChangeEventArgs() { Value = value });
-                }
-
                 SelectedOption = item;
-
-                //InternalValue = Value = value;
-                InternalValue = value;
+                InternalValue = GetOptionValue(item);
                 await RaiseChangedEventsAsync();
             }
         }
@@ -562,6 +551,8 @@ public abstract partial class ListComponentBase<TOption> : FluentInputBase<strin
                 await SelectedOptionChanged.InvokeAsync(SelectedOption);
             }
         }
+
+        await base.ChangeHandlerAsync(new ChangeEventArgs() { Value = InternalValue });
     }
 
     protected virtual async Task OnKeydownHandlerAsync(KeyboardEventArgs e)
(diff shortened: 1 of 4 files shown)


src/Core/Components/List/ListComponentBase.razor.cs
```diff
@@ -291,13 +306,35 @@ public override async Task SetParametersAsync(ParameterView parameters)
             }
         }
 
+        if (Multiple)
+        {
+            foreach (var parameter in parameters)
```
dvoituron, line 311: It's probably left over from a test :-)

src/Core/Components/List/ListComponentBase.razor.cs
```diff
@@ -527,19 +581,8 @@ protected virtual async Task OnSelectedItemChangedHandlerAsync(TOption? item)
         {
             if (!Equals(item, SelectedOption))
             {
-                var value = GetOptionValue(item);
-
-                if (this is FluentListbox<TOption> ||
```
dvoituron, line 532: Why did you remove these lines?

joriverm: wouldnt it have caused `base.ChangeHandlerAsync` to be called twice for single options controls? once from here and another from the `ChangeHandlerAsync` in `RaiseChangedEventAsync` below?

dvoituron: Probably. To check :-)

joriverm: just checked, it indeed gets called from the `RaiseChangedEventAsync` below :)

dvoituron (approved):

vnbaaij (approved):

PR #3319, opened 2025-02-07
[DataGrid] Move down column action popup

vnbaaij:
To prevent z-index issues, the popup is now aligned to the bottom of the column headers (instead of being partially on top of the header).

### DIFF
--- a/src/Core/Components/DataGrid/FluentDataGridCell.razor.cs
+++ b/src/Core/Components/DataGrid/FluentDataGridCell.razor.cs
@@ -80,7 +80,7 @@ public partial class FluentDataGridCell<TGridItem> : FluentComponentBase
         .AddStyle("height", "100%", Grid.MultiLine)
         .AddStyle("min-height", "44px", Owner.RowType != DataGridRowType.Default)
         .AddStyle("display", "flex", ShouldHaveDisplayFlex())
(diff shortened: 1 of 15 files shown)


src/Core/Components/DataGrid/FluentDataGridCell.razor.cs
```diff
@@ -80,7 +80,7 @@ public partial class FluentDataGridCell<TGridItem> : FluentComponentBase
         .AddStyle("height", "100%", Grid.MultiLine)
         .AddStyle("min-height", "44px", Owner.RowType != DataGridRowType.Default)
         .AddStyle("display", "flex", ShouldHaveDisplayFlex())
-        .AddStyle("z-index", (Grid._columns.Count + 1 - Column?.Index).ToString(), CellType == DataGridCellType.ColumnHeader && Grid._columns.Count > 0)
+        .AddStyle("z-index", "1", CellType == DataGridCellType.ColumnHeader && Grid._columns.Count > 0)
```
dvoituron, line 83: You could create a new value in the `ZIndex.cs` class to centralize these values

dvoituron (changes requested):

dvoituron (approved):

PR #3322, opened 2025-02-08
[dev-v5] Refactoring the generator of API documentation

dvoituron:
# [dev-v5] Refactoring the generator of API documentation

The current project `FluentUI.Demo.Generators.csproj` is an [Analyzer](https://github.com/dotnet/roslyn-analyzers/blob/main/docs/rules/RS1041.md) project which had to be compiled to **netstandard2.0**.

The documentation file generated by Visual Studio is an XML file containing some tags like `<inheritdoc />` to automatically read the comment from the parent element. To manage these flows, we need to use a library [LoxSmoke.DocXml](https://www.nuget.org/packages/LoxSmoke.DocXml) and to load the `Microsoft.FluentUI.AspNetCore.Components` library developed with .NET9. These different .NET versions are not compatible.

So, we created a new project `FluentUI.Demo.DocApiGen` to generate a JSON file `api-comments.json` that is read read by the Demo app to populate the API tables.
This `DocApiGen` is a console app that is already configured for the developer using the `launchSettings.json` file
```
Usage: DocApiGen --xml <xml_file>
                 --dll <dll_file>
                 --output <generated_file>
                 --format <csharp|json>
```

## Developer usage

Without actions, the demo application will launch without errors, but the API tables will be empty (only the Description column).

To populate this column, the developer must run the `DocApiGen` project (right click -> Debug -> Start without debugging) and the JSON file will be saved in the `FluentUI.Demo.Client/wwwroot` folder.
{A91619C2-6370-44A6-A3AC-060A3C6AFCD4}

## Build pipeline and publication

This command must be included in the YML file to generate the JSON file in the `wwwroot` folder.


vnbaaij (approved):
👏

Issue #2882, opened 2024-10-29
fix: Missing support of unsigned numerics in `FluentNumberField`


Eagle3386:
# 🐛 Bug Report
I chose "bug" over "feature request" since a component named `FluentNumberField` should have supported the unsigned equivalents of `int`, `long`, `sbyte` & `short` right from the start.
Because I can't imagine a reason to not enable developers to limit sliders, steppers, etc. to positive values only (talking mostly age, RGB color & range pickers here, but applies to customer ID fields as well).

On a more personal note, I'm _really_ having a hard time understanding why (C#) developers still prefer `int CustomerId` over `uint CustomerId` since such a property never really contains a negative integer.
But I assume it's just like with `decimal` which - besides scientific calculations - should always be preferred over `float` & `double` (AFAIK, even the C# docs explicitly state this).. 😅

## 💻 Repro or Code Sample
```razor
@* // … code left out for brevity… *@
<FluentNumberField HideStep="true" Immediate="true" Label="Customer #:" Max="999999" MaxLength="6"
                   Min="100000" MinLength="6" Required="true" @bind-Value="Account.CustomerId" />
@* // … code left out for brevity… *@
@code
{
  private AccountData Account { get; } = new();
  // … code left out for brevity…
  public record AccountData
  {
    internal uint? CustomerId { get; set; }
    // … code left out for brevity…
  }
}
```

## 🤔 Expected Behavior
As I explicitly declared `HideStep="true"` above, the following method shouldn't even get called.
However, if it's necessary for some reason, this code inside `FluentNumberField.razor.cs`:
```csharp
private static string GetStepAttributeValue()
{
    // Unwrap Nullable<T>, because InputBase already deals with the Nullable aspect
    // of it for us. We will only get asked to parse the T for nonempty inputs.
    var targetType = Nullable.GetUnderlyingType(typeof(TValue)) ?? typeof(TValue);
    if (targetType == typeof(sbyte) ||
        targetType == typeof(int) || typeof(by)
        targetType == typeof(long) ||
        targetType == typeof(short) ||
        targetType == typeof(float) ||
        targetType == typeof(double) ||
        targetType == typeof(decimal))
    {
        return "1";
    }
    else
    {
        throw new InvalidOperationException($"The type '{targetType}' is not a supported numeric type.");
    }
}
```
should include the 4 unsigned numerics, too.

## 😯 Current Behavior
A `InvalidOperationException` is thrown:
```log
[...]

PR #3373, opened 2025-02-16
[NumberField] Unsigned integer values and fixes

PascalVorwerk:
# Pull Request

## 📖 Description

I came across #2882 and thought this was an interesting proposal. In my own applications i have felt the need to set the minimum to 0 without allowing the user to go below it! Now to my supprise, the current implementation actually has some bugs (in my eyes):
- If one does not provide a min or max property to the numberfield, you can go over or below the minimum and max values for this type. So for example a short value could over its max:

image

This makes type safety not really a thing.. So my proposal would be to 'enforce' the minimum and maximum values of each type by using the min/max values of the type as a default parameter. @vnbaaij already made a nice helper method to get these minimum and maximum values for each type.

- The old code would not validate the minimum and maximum value parameters if one of them was null, but i think this is wrong behaviour. I would say a 'Typed' number should always have a minimum and maximum value, even if that value is unbelievably large or small. So i think if either one of them is set to null by a user, the system should throw an exception.

### 🎫 Issues

#2882 

## 👩‍💻 Reviewer Notes

I've had to custom check for the unsigned integers and parse them separately because the BindConverter of AspNetCore does not implement extension methods for these types. I could not extend the BindConvert because for some reason it gave me a sealed error, so i am manually checking for the types, after first trying it with the BindConvert. Let me know if you guys think this is a wrong approach!

## 📑 Test Plan

I have added some tests and examples to the projects. The tests validate the new Min/max parameters. I've had to regenerate the verified files because now each fluent-number-input component will always have a min/max by default generated.

### General


PascalVorwerk:
Im sorry i am not sure why the commits of my previous pull request are included as commits... Can i somehow undo that? Or is it not a problem?

src/Core/Components/NumberField/FluentNumberField.razor.cs
```diff
@@ -61,13 +61,13 @@ public partial class FluentNumberField<TValue> : FluentInputBase<TValue>, IAsync
     /// Gets or sets the maximum value.
     /// </summary>
     [Parameter]
-    public string? Max { get; set; }
+    public string? Max { get; set; } = InputHelpers<TValue>.GetMaxValue();
```
dvoituron, line 64: You cannot set default values for `Min` and `Max`. With these default values, the generated html **always** contains `min` and `max` attributes, and we don't want that.

dvoituron (changes requested):

src/Core/Components/NumberField/FluentNumberField.razor.cs
PascalVorwerk: Why do we not want that?

vnbaaij: Attributes in html are not required so we should not make them required either. It is also technically (not talking about logically here) not required to have them.

PascalVorwerk: We're not making them required, just filled by default to the underlying minimum and maximum values of their type! You could still supply a null value if you really do not want a minimum value.

I understand that technically they are not required because in javascript it is not needed, but converting these values to a typed system is seems logical to me to impose some minimal/maximum values? But right now the current implementation can take larger and smaller numbers in for example shorts.. like shown in the picture

PascalVorwerk: This way, you actually enforce some type safety and equality from the javascript input element to the c# data type behind it

PascalVorwerk: Hmm i did make them required by adding the check for null values in the ValidateInputParameters method.. but that is something else then supplying default values. Now if someone sets the 'min' property to null, an exception will be thrown. I can understand it from that point, but i don't see why the default values are an issue. It makes life easier for anyone using a number input with their respective types.

PascalVorwerk: image
We are doing the same for other parameters btw..

dvoituron: These values must remain `Null` because 
1. As Vincent says, it is not required.
2. By applying these values, the Html output generated will be larger, which can impact network performance.
3. This is no longer compatible with the current result. In fact, you've had to modify almost all the unit tests in this respect. This could be breaking changes for some devs at the moment.
4. Nothing prevents you from managing private properties that return these values by default, when Min/Max is null.

I don't see any reason to change this.

PascalVorwerk: So the reason that one can by default now enter larger numbers then the underlying data type is not reason enough? Cause that is the behaviour right now. The types do not do anything to enforce the inputs in the component.

PascalVorwerk: image 

> These values must remain `Null` because
> 
> 1. As Vincent says, it is not required.
> 2. By applying these values, the Html output generated will be larger, which can impact network performance.
> 3. This is no longer compatible with the current result. In fact, you've had to modify almost all the unit tests in this respect. This could be breaking changes for some devs at the moment.
> 4. Nothing prevents you from managing private properties that return these values by default, when Min/Max is null.
> 
> I don't see any reason to change this.

1. Supplying a default value to a nullable type does not mean it makes the property a required property?
2. I would suggest this is very neglible..
3. Indeed, but the only usecase someone would have a 'breaking' change is if they were going over or under the limits of their underlying data types.. now they won't be able to do that anymore (unless they set the min/max property to null)
4. What would this be different from my current change? The unit tests would still generate those private properties then.. 

Sorry but i dont understand, it's not like i am making these parameters a 'required' thing to fill in now? People can still use a FluentNumberField without having to worry about the Min/Max properties... but now it will work as expected to the underlying data type? E.g. a short data type will never be higher then...

PascalVorwerk: Okay I hope I have found a middleground for this issue. I've added a opt-in boolean to allow users to specify if they want the input to automatically constrain the input to the types min/max values. 

If a user specifically sets a Min or Max property next to this boolean, it will always use the explicitly set Min/Max property. 

This way the feature remains backwards compatibility and only introduces extra html/parameters when requested by the user.

The boolean parameter is ofcourse defaulting to false, so any existing users will not have any breaking changes or changes to behaviour.

rpodevns:
Another issue with default `Min` and `Max` is that it could break existing `CSS`, such as:

```css
fluent-number-field:has([min]) {
    border: 1px solid red;
}
```

`MinLength = 1` likely ensures no blank values and `MaxLength = 14` will prevent excessively long numbers because the `fluent-number-field` web component is `type="text".

src/Core/Components/NumberField/FluentNumberField.razor.cs
vnbaaij: > Supplying a default value to a nullable type does not mean it makes the property a required property?

What we mean by this is that once you supply a default value to a property, it will always show up in the HTML as an attribute. That is not the case when using the web component by itself and we do not want to change that default behavior.

Have to look at your latests change. Seems like a good middleground indeed

src/Core/Components/NumberField/FluentNumberField.razor.cs
```diff
@@ -86,26 +86,70 @@ public partial class FluentNumberField<TValue> : FluentInputBase<TValue>, IAsync
...
+
+            return Max;
+        }
+        set
+        {
+            Max = value;
+        }
+    }
+
+   private static string GetStepAttributeValue()
```
dvoituron, line 142: There is a space missing in front of the ‘private’.

tests/Core/NumberField/FluentNumberFieldTests.FluentNumberField_AdditionalParameter.verified.html
```diff
@@ -1,2 +1,2 @@
 
-<fluent-number-field maxlength="14" minlength="1" size="20" step="1" id="xxx" value="100" appearance="outline" blazor:onchange="1" blazor:oninput="2" additional-parameter-name="additional-parameter-value" blazor:elementreference="xxx">100</fluent-number-field>
```
dvoituron, line 2: All these unit tests should not be modified. Why do you want to remove the return from the bottom of the file?
Can you put back all the previous unit tests? To avoid telling Git that a change has been made when it hasn't.

Only your new tests should be present.

src/Core/Components/NumberField/FluentNumberField.razor.cs
```diff
@@ -86,26 +86,70 @@ public partial class FluentNumberField<TValue> : FluentInputBase<TValue>, IAsync
...
+    /// unless an explicit value for Min or Max is provided.
+    /// </summary>
+    [Parameter]
+    public bool UseTypeConstraints { get; set; }
+
     private static readonly string _stepAttributeValue = GetStepAttributeValue();
 
-    private static string GetStepAttributeValue()
+    // If type constraints is true and min is null, set min to the minimum value of TValue.
+    private string? MinValue
```
dvoituron, line 107: You don't need the `set` property.
In my opinion, only this is sufficient:
```csharp
private string? MinValue => (UseTypeConstraints && Min == null) ? InputHelpers<TValue>.GetMinValue() : Min;
private string? MaxValue => (UseTypeConstraints && Max == null) ? InputHelpers<TValue>.GetMaxValue() : Max;
```

src/Core/Components/NumberField/FluentNumberField.razor.cs
```diff
@@ -118,42 +162,52 @@ private static string GetStepAttributeValue()
...
+                validationErrorMessage = null;
+                return true;
+            }
+        }
+        catch (ArgumentException)
+        {
+            result = default!;
+        }
+
+
```
dvoituron, line 206: Avoid multiple blank lines

dvoituron (changes requested):

tests/Core/NumberField/FluentNumberFieldTests.FluentNumberField_AdditionalParameter.verified.html
PascalVorwerk: The return from the bottom file removal was not intended, i tried to revert back to the old situation, but it still has some changes. I am pulling the files from development branch as they were, except for my own added tests!

src/Core/Components/NumberField/FluentNumberField.razor.cs
PascalVorwerk: Fixed

tests/Core/NumberField/FluentNumberFieldTests.FluentNumberField_AdditionalParameter.verified.html
PascalVorwerk: Okay so appearantly my Rider IDE autoremoves these trailing endlines.. :)

dvoituron: The `.editorconfig` file and this line should force this line to be present.

{7533E2BE-DD8C-4408-8EB4-1556D4DCDF41}

dvoituron: You can also use [Visual Studio](https://visualstudio.microsoft.com/vs/community) #joke ;-)

src/Core/Components/Base/InputHelpers.cs
```diff
@@ -14,12 +14,12 @@ public static string GetMaxValue()
             TypeCode.Double => double.MaxValue.ToString(CultureInfo.InvariantCulture),
             TypeCode.Int16 => short.MaxValue.ToString(),
             TypeCode.Int32 => int.MaxValue.ToString(),
-            TypeCode.Int64 => "999999999999",
+            TypeCode.Int64 => long.MaxValue.ToString(),
```
vnbaaij, line 17: I'm pretty sure that this value was there because the web components can't take more than 12 digits. We need to make sure that is checked on the wc side. Sending a real maxvalue that way will lead to problems then.

Same goes for UInt64 and for the minvalues

vnbaaij: See https://github.com/microsoft/fast/issues/6713

PascalVorwerk: I see what you mean. trying to enter in a higher number results in this: 
image

So any higher number then the one previously stated will be bugged and not go higher? Although.. when entering a realllly really large number, it will go to a scientific notation:

image

Shall i revert the changes for Long values?

PascalVorwerk: The same happens for the other values: 
image

I think the component changes anything after 10 digits to a 'scientific number' and in result truncating any numbers after it to a 0. If the number is 'small' enough to still be visible in the component, the scientific notation will be left out.

What would we like to do for the other types?

PascalVorwerk: There are some more weird things happening.. For some reason, when entering a float number, my current implementation is triggering the OnChangeEvent args twice... Resulting in weird behaviour..

https://github.com/user-attachments/assets/2507ff11-23ca-4df0-8a09-6043092182dc

PascalVorwerk: > There are some more weird things happening.. For some reason, when entering a float number, my current implementation is triggering the OnChangeEvent args twice... Resulting in weird behaviour..
> 
>  doubleEvents.mp4

Okay i have this fixed locally now, waiting with commit for answer about the double and floats. They are being truncated aswell, would you like me to add a maximum of "9999999999" there aswell?

vnbaaij: Yes, that seems to be the only choice we have for that.

Appreciate your efforts here. I did warn/complain before that the Web Component implementation is, how do I say that nicely, ...not very polished. I wouldn't try to fix/adapt/change too much on our side anymore after this one.

PascalVorwerk: It does seem to defeat the point of having these different types for the input, but ah well.. 😂

PascalVorwerk: I realize this might all be not really worth it especially with the new components coming up.. but atleast i am learning alot about how this library is set up 😄

vnbaaij: It is not like v4 will not be around anymore anytime soon, so definitely not waisted effort!

vnbaaij (approved):

PR #3391, opened 2025-02-18
[dev-v5] Add FluentSpinner component

dvoituron:
# [dev-v5] Add FluentSpinner component

A spinner alerts a user that content is being loaded or processed and they should wait for the activity to complete.

> The component **FluentProgressRing** has been renamed to **FluentSpinner** to be coherent with the Web Components and React naming conventions.
>
> The new component `FluentSpinner` cannot be paused. It must always be spinning.
> So, the `Paused`, `Min`, `Max`, `Value` properties has been removed.

The spinner can have multiple sizes, the default size is `Medium`.
```xml
<FluentSpinner Size="@SpinnerSize.Huge" />
```

Spinner-1

In some situations, you may need to display a spinner on a dark background.
```xml
<FluentSpinner AppearanceInverted="true" />
```

Spinner-2

### DIFF
--- a/examples/Demo/FluentUI.Demo.Client/Documentation/Components/Progress/Examples/SpinnerInverted.razor
+++ b/examples/Demo/FluentUI.Demo.Client/Documentation/Components/Progress/Examples/SpinnerInverted.razor
@@ -1,0 +1,22 @@
+﻿<FluentStack HorizontalAlignment="HorizontalAlignment.Center"
+             VerticalAlignment="VerticalAlignment.Center"
+             Padding="@Padding.All2">
+    <FluentSpinner />
+    <FluentText Margin="@Margin.Horizontal3">Primary spinner</FluentText>
+</FluentStack>
+
+
+<FluentStack HorizontalAlignment="HorizontalAlignment.Center"
+             VerticalAlignment="VerticalAlignment.Center"
+             Class="inverted-colors"
+             Padding="@Padding.All2">
+    <FluentSpinner AppearanceInverted="true" />
+    <FluentText Margin="@Margin.Horizontal3">Inverted spinner</FluentText>
+</FluentStack>
+
+<style>
+    .inverted-colors {
+        color: var(--colorNeutralForegroundOnBrand);
+        background-color: var(--colorBrandBackgroundHover);
+    }
+</style>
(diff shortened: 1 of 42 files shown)


examples/Demo/FluentUI.Demo.Client/Documentation/Components/Progress/Examples/SpinnerInverted.razor
vnbaaij, line 1: Should we not do the inverted stuff through a parameter and StyleBuilder?

vnbaaij (changes requested):

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Progress/Examples/SpinnerInverted.razor
dvoituron: What do you mean? We have a parameter linked to the same webcomponent parameter.

src/Core/Enums/SpinnerSize.cs
vnbaaij, line 1: The default value for the web component is 'medium'. Think that needs to be the first value in the enum then as well. So when people define a non-nullable variable of this type. the default value will be correct.

tests/Core/Components/Progress/FluentSpinnerTests.razor
```diff
@@ -0,0 +1,72 @@
...
+    [Theory]
+    [InlineData(SpinnerSize.Tiny)]
+    [InlineData(SpinnerSize.ExtraSmall)]
+    [InlineData(SpinnerSize.Small)]
+    [InlineData(SpinnerSize.Medium)]
+    [InlineData(SpinnerSize.Large)]
+    [InlineData(SpinnerSize.ExtraLarge)]
+    [InlineData(SpinnerSize.Huge)]
+    [InlineData((SpinnerSize)999)]
+
```
vnbaaij, line 44: Should there not be a InlineData `null` test as well

vnbaaij (changes requested):

src/Core/Enums/SpinnerSize.cs
dvoituron: Interesting :-)
Done in next commit

tests/Core/Components/Progress/FluentSpinnerTests.razor
dvoituron: Yes :-)
Done

vnbaaij (approved):

PR #3402, opened 2025-02-19
[dev-v5] Add FluentSlider component

tossnet:
# [dev-v5] Add FluentSlider component

The **FluentSlider** component is an input where the user selects a value from within a given range. Sliders typically have a slider thumb that can be moved along a bar, rail, or track to change the value of the slider.

**Default** settings are `Min="0"`, `Max="100"` and `Step="1"`.

**Value** can be one of the following **Numeric Type**:
`byte`, `sbyte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `float`, `double`, `decimal`.

Example

```razor
<FluentStack Orientation="Orientation.Vertical">
    <FluentSlider @bind-Value="@value">Hello</FluentSlider>
    <FluentLabel>Value: @value</FluentLabel>
</FluentStack>

@code
{
    int value = 50;
}
```
{74FD95D0-DE0A-4310-BB9E-A1DEDE4D8505}

## Unit tests

{4E4C229E-7BE2-479B-80A2-9F9621EF79D0}


src/Core/Components/Slider/FluentSlider.razor.cs
```diff
@@ -0,0 +1,124 @@
...
+using Microsoft.AspNetCore.Components.Web;
+using Microsoft.FluentUI.AspNetCore.Components.Extensions;
+
+namespace Microsoft.FluentUI.AspNetCore.Components;
+
+/// <summary>
+/// FluentSlider component, a slider control that allows users to select from a range of values.    
+/// </summary>
+/// <typeparam name="TValue"></typeparam>
+public partial class FluentSlider<TValue> : FluentInputBase<TValue>
```
dvoituron, line 17: You can add a constraint: `where TValue : struct, IComparable<TValue>`

src/Core/Components/Slider/FluentSlider.razor.cs
```diff
@@ -0,0 +1,124 @@
...
+            throw new InvalidOperationException("FluentSlider only supports numeric types.");
+        }
+    }
+
+    /// <summary>
+    /// Formats the value as a string. Derived classes can override this to determine the formatting used for <c>CurrentValueAsString</c>.
+    /// </summary>
+    /// <param name = "value">The value to format.</param>
+    /// <returns>A string representation of the value.</returns>
+    protected override string? FormatValueAsString(TValue? value)
```
dvoituron, line 84: Due to the generic constraint (`where`), the override method accepts only a non-nullable TValue:
```csharp
protected override string? FormatValueAsString(TValue value)
{
    return Convert.ToString(value, System.Globalization.CultureInfo.InvariantCulture);
}
```

src/Core/Components/Slider/FluentSlider.razor.cs
```diff
@@ -0,0 +1,124 @@
...
+    /// <param name="result">The result to inject into the Value.</param>
+    /// <param name="validationErrorMessage">If the value could not be parsed, provides a validation error message.</param>
+    /// <returns>True if the value could be parsed; otherwise false.</returns>
+    protected override bool TryParseValueFromString(string? value, [MaybeNullWhen(false)] out TValue result, [NotNullWhen(false)] out string? validationErrorMessage)
+    {
+        return this.TryParseSelectableValueFromString(value, out result, out validationErrorMessage);
+    }
+
+    // Only for Unit Tests
+    internal string? ValidateFormatValueAsString(TValue? value) => FormatValueAsString(value);
```
dvoituron, line 123: You can replace this method with this one to be used in the razor file when a null value can be returned.
```csharp
internal string? FormatValueAsStringOrNull(TValue? value)
{
    if (value is null)
    {
        return null;
    }

    return FormatValueAsString(value.Value);
}
```

src/Core/Components/Slider/FluentSlider.razor
```diff
@@ -0,0 +1,28 @@
+﻿@namespace Microsoft.FluentUI.AspNetCore.Components
+@using Microsoft.FluentUI.AspNetCore.Components.Extensions
+@inherits FluentInputBase<TValue>
+@typeparam TValue
+
+<FluentField InputComponent="@this" ForId="@Id" Class="@ClassValue" Style="@StyleValue">
+	<fluent-slider slot="input"
+	               id="@Id"
+	               disabled="@Disabled"
+	               min="@FormatValueAsString(Min)"
```
dvoituron, line 10: These attributes must return a `null` value if the parameter is null, using the new method `FormatValueAsStringOrNull`
```razor
min="@FormatValueAsStringOrNull(Min)"
max="@FormatValueAsStringOrNull(Max)"
value="@FormatValueAsString(Value)"
step="@FormatValueAsStringOrNull(Step)"
```

dvoituron (changes requested):

src/Core/Components/Slider/FluentSlider.razor.cs
```diff
@@ -0,0 +1,124 @@
...
+    /// <summary>
+    /// Gets or sets the slider's minimal value.
+    /// </summary>
+    [Parameter]
+    public TValue? Min { get; set; }
+
+    /// <summary>
+    /// Gets or sets the slider's maximum value.
+    /// </summary>
+    [Parameter, EditorRequired]
```
dvoituron, line 34: It cannot be a mandatory parameter

dvoituron (changes requested):

src/Core/Components/Slider/FluentSlider.razor
```diff
@@ -0,0 +1,28 @@
...
+	               step="@FormatValueAsStringOrNull(Step)"
+	               size="@Size.ToAttributeValue()"
+	               orientation="@(Orientation == Orientation.Horizontal ? null : Orientation.ToAttributeValue())"
+	               mode="@Mode.ToAttributeValue()"
+	               name="@Name"
+	               readonly="@ReadOnly"
+	               aria-label="@AriaLabel"
+	               autofocus="@Autofocus"
+	               required="@Required"
+	               tabindex="@GetTabIndexValue"
```
dvoituron, line 22: Either you use the verb "get" and that must be a method with parenthesis: `GetTabIndexValue()`
Either you don't use a verb and that must be a property `TabIndexValue`

dvoituron: PS: I prefer to use a method :-)

src/Core/Components/Slider/FluentSlider.razor
```diff
@@ -0,0 +1,28 @@
...
+	               @onfocusout="@FocusOutHandlerAsync"
+	               @onchange="@ChangeHandlerAsync"
+	               @attributes="AdditionalAttributes">
```
dvoituron, line 25: Please add `@` to specify that is a C# element: `@AdditionalAttributes`

src/Core/Components/Slider/FluentSlider.razor.cs
```diff
@@ -0,0 +1,130 @@
...
+    where TValue : struct, IComparable<TValue>
+{
+    /// <summary>
+    /// Gets or sets the size for the slider.
+    /// </summary>
+    [Parameter]
+    public SliderSize? Size { get; set; }
+
+    /// <summary>
+    /// Gets or sets the slider's minimal value.
```
dvoituron, line 26: Please, add `Default is 0`

src/Core/Components/Slider/FluentSlider.razor.cs
```diff
@@ -0,0 +1,130 @@
...
+    /// </summary>
+    [Parameter]
+    public TValue? Min { get; set; }
+
+    /// <summary>
+    /// Gets or sets the slider's maximum value.
```
dvoituron, line 32: Please, add `Default is 100`

src/Core/Components/Slider/FluentSlider.razor.cs
```diff
@@ -0,0 +1,130 @@
...
+    /// </summary>
+    [Parameter]
+    public TValue? Max { get; set; }
+
+    /// <summary>
+    /// Gets or sets the slider's step value.
+    /// </summary>
+    [Parameter]
+    public TValue? Step { get; set; }
```
dvoituron, line 41: Please, add `Default is 1`

src/Core/Components/Slider/FluentSlider.razor.cs
```diff
@@ -0,0 +1,130 @@
...
+
+    /// <summary>
+    /// Gets or sets the orientation of the stacked components.
```
dvoituron, line 44: Please, add `Default is Horizontal`

src/Core/Components/Slider/FluentSlider.razor.cs
```diff
@@ -0,0 +1,130 @@
...
+    /// </summary>
+    [Parameter]
+    public Orientation Orientation { get; set; }
+
+    /// <summary>
+    /// Gets or sets the selection mode.
+    /// </summary>
+    [Parameter]
+    public SliderMode? Mode { get; set; }
```
dvoituron, line 53: Can you remove this parameter? There's no point in having a property that isn't configurable :-)
If later, we will have a second option, we will add this property.

src/Core/Components/Slider/FluentSlider.razor.cs
```diff
@@ -0,0 +1,130 @@
...
+
+    /// <summary>
+    /// Gets or sets the content to be rendered inside the component.
+    /// </summary>
+    [Parameter]
+    public RenderFragment? ChildContent { get; set; }
```
dvoituron, line 59: What is the goal of this property? I tried this code, but without success.
```xml
<fluent-slider>Hello</fluent-slider>
```

I will remove this property **ChildContent**.

src/Core/Components/Slider/FluentSlider.razor.cs
```diff
@@ -0,0 +1,130 @@
...
+
+    /// <summary>
+    /// Gets or sets the value of the slider.
+    /// </summary>
+    protected override void OnParametersSet()
+    {
+        base.OnParametersSet();
+
+        if (typeof(TValue) != typeof(byte) &&
```
dvoituron, line 68: It is not consistent with the list of supported types in the documentation (1 missing here... `short`?).

tests/Core/Components/Base/InputBaseTests.cs
```diff
@@ -38,7 +38,8 @@ public class InputBaseTests : TestContext
     private static readonly Dictionary<Type, Func<Type, Type>> ComponentInitializer = new()
     {
         // { typeof(FluentIcon<>), type => type.MakeGenericType(typeof(Samples.Icons.Samples.Info)) }
-        { typeof(FluentSelect<>), type => type.MakeGenericType(typeof(string)) }    // FluentSelect<string>
+        { typeof(FluentSelect<>), type => type.MakeGenericType(typeof(string)) },// FluentSelect<string>
```
dvoituron, line 41: Don't remove the space before the comment `//`

tests/Core/Components/Slider/FluentSliderTests.razor
```diff
@@ -0,0 +1,212 @@
...
+	{
+		JSInterop.Mode = JSRuntimeMode.Loose;
+		Services.AddFluentUIComponents();
+	}
+
+	[Fact]
+	public void FluentSlider_Default()
+	{
+		// Arrange && Act
+		var cut = Render(@<FluentSlider Value=10>fluent-Slider</FluentSlider>);
```
dvoituron, line 18: Can you use double quote with all attributes?

tests/Core/Components/Slider/FluentSliderTests.razor
```diff
@@ -0,0 +1,212 @@
...
+
+		// Assert
+		cut.Verify();
+	}
+
+	[Fact]
+	public void FluentSlider_Min()
+	{
+		// Arrange && Act
+		var cut = Render(@<FluentSlider Min="2" Value=10 />);
```
dvoituron, line 28: Can you use double quote with all attributes?

tests/Core/Components/Slider/FluentSliderTests.razor
```diff
@@ -0,0 +1,212 @@
...
+
+		// Assert
+		cut.Verify();
+	}
+
+	[Theory]
+	[InlineData(SliderSize.Medium)]
+	[InlineData(SliderSize.Small)]
+	[InlineData((SliderSize)999)]
```
dvoituron, line 37: You need to test the `null` value

tests/Core/Components/Slider/FluentSliderTests.razor
```diff
@@ -0,0 +1,212 @@
...
+	[InlineData(true)]
+	[InlineData(false)]
+	public void FluentSlider_DisabledAttribute(bool disabled)
+	{
+		// Arrange && Act
+		var cut = Render(@<FluentSlider Value=10 Disabled="@disabled" />);
+
+		// Assert
+		var actual = cut.Find("fluent-slider").HasAttribute("disabled");
+		Assert.Equal(disabled, actual);
```
dvoituron, line 57: Can you remove the unused ".verified.html" files (here :  `...disabled...` file)?

tests/Core/Components/Slider/FluentSliderTests.razor
```diff
@@ -0,0 +1,212 @@
...
+		// Arrange && Act
+		byte value = byte.MaxValue;
+		var cut = Render(@<FluentSlider @bind-Value="@value" />);
+
+		// Assert
+		cut.Verify();
+	}
+
+ //    [Fact]
+ //    public void FluentSlider_Value_String()
```
dvoituron, line 105: Please, remove the unused code

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Slider/FluentSlider.md
```diff
@@ -0,0 +1,54 @@
...
+## Orientation
+
+{{ SliderOrientation }}
+
+## Size
+
+{{ SliderSizes }}
+
+## Thumbs
+
```
dvoituron, line 39: Can you explain what `slot=“ thumb”` is and when it might be interesting to use it?

dvoituron (changes requested):

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Slider/Examples/SliderOrientation.razor
vnbaaij, line 1: This should display all (in this case both) options next to each other.

src/Core/Components/Slider/FluentSlider.razor
vnbaaij, line 1: This should display all possible options (int this case sizes) next to each other.

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Slider/Examples/SliderDefault.razor
vnbaaij, line 1: When the slider is set to `ReadOnly` the value can still be changed with the keyboard (arrow left/right, home, end)

vnbaaij: Upstream issue opened: https://github.com/microsoft/fluentui/issues/33936

tests/Core/Components/Slider/FluentSliderTests.razor
```diff
@@ -0,0 +1,199 @@
...
+
+		// Assert
+		cut.Verify();
+	}
+
+	[Theory]
+	[InlineData(Orientation.Horizontal)]
+	[InlineData(Orientation.Vertical)]
+	[InlineData((Orientation)999)]
+	public void FluentSlider_Orientation(Orientation orientation)
```
vnbaaij, line 140: You need to test for null value

vnbaaij (changes requested):

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Slider/Examples/SliderDefault.razor
vnbaaij: Response received: 

Nnative HTML sliders do not support a readonly attribute, only inputs that accept text entry may be readonly, as per the HTML spec:

>"Only text controls can be made read-only, since for other controls (such as checkboxes and buttons) there is no useful distinction between being read-only and being disabled, so the [readonly](https://html.spec.whatwg.org/multipage/input.html#attr-input-readonly) attribute [does not apply](https://html.spec.whatwg.org/multipage/input.html#do-not-apply)."

from the HTML spec: https://html.spec.whatwg.org/multipage/input.html#the-readonly-attribute

@dvoituron we need to look at all input components to see whether they can support ReadOnly or not

vnbaaij: JS code like this can be used:

```
const slider = document.getElementById('slider1') as Slider;

if (slider.getAttribute('readonly') == 'true')
{
    var x = slider.handleKeydown(event);
    slider.handleKeydown = function (event) { return false; }
}
```

CSS code like this can be used:
```
<style>
  fluent-slider[readonly="true"]{
    pointer-events: none;
    user-select: none;
  }
</style>
```

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Slider/Examples/SliderOrientation.razor
dvoituron: This is an example for vertical orientation, horizontal orientation already being displayed in all the other examples.

dvoituron (approved):

PR #3410, opened 2025-02-19
[dev-v5] Add FluentRatingDisplay component

AClerbois:
# Add FluentRatingDisplay component

RatingDisplay is used to communicate user sentiment. By default, it shows rating as filled stars out of 5, as well as a text displaying the average value and the aggregate number of ratings.

## Default implementation

image

## Count

You can specify the total number of ratings being displayed with the Count. 

image

## Compact Mode

image

## Change max count 

image

## Shape 

image

## Appearances 

`Size` and `Color`

image

# Documentation

- Best practices
- Default
- Value
- Count
- Compact
- Max
- Shape
- Appearances
- API FluentRatingDisplay

# Coverage & Testing

image


src/Core/Components/RatingDisplay/FluentRatingDisplay.razor.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// </summary>
+    [Parameter]
+    public double? Count { get; set; }
+
+    /// <summary>
+    /// Gets or sets the max value of the rating.
+    /// This controls the number of rating items displayed. Must be a whole number greater than 1
+    /// </summary>
+    [Parameter]
+    public double? Max { get; set; }
```
vnbaaij, line 47: This is the number of 'elements' displayed, right? Should this not just be an integral whole number so `int` or even `byte`

src/Core/Components/RatingDisplay/FluentRatingDisplay.razor
```diff
@@ -0,0 +1,21 @@
+﻿@namespace Microsoft.FluentUI.AspNetCore.Components
+@using Microsoft.FluentUI.AspNetCore.Components.Extensions
+@inherits FluentComponentBase
+
+<fluent-rating-display id="@Id"
+                       class="@ClassValue"
+                       style="@StyleValue"
+                       count="@Count?.ToString("0.00", System.Globalization.CultureInfo.InvariantCulture)"
+                       compact="@Compact"
+                       max="@Max?.ToString("0.00", System.Globalization.CultureInfo.InvariantCulture)"
```
vnbaaij, line 10: See remark about type in the `.cs` file

src/Core/Components/RatingDisplay/FluentRatingDisplay.razor
```diff
@@ -0,0 +1,21 @@
...
+                       size="@Size.ToAttributeValue()"
+                       color="@Color.ToAttributeValue()"
+                       value="@Value?.ToString("0.00", System.Globalization.CultureInfo.InvariantCulture)"
```
vnbaaij, line 13: Can the web component handle the rounding? And will it then always be displayed in x.5 even if regional settings would use a ','

src/Core/Enums/RatingDisplayColor.cs
vnbaaij, line 1: The default value for the web component is 'marigold'. Think that needs to be the first value in the enum here then as well. So when people define a non-nullable variable of this type. the default value will be correct.

src/Core/Enums/RatingSize.cs
vnbaaij, line 1: The default value for the web component is 'medium'. Think that needs to be the first value in the enum here then as well. So when people define a non-nullable variable of this type. the default value will be correct.

vnbaaij (changes requested):

src/Core/Components/RatingDisplay/FluentRatingDisplay.razor
AClerbois: @vnbaaij It tested on my computer, if you use a comma, the value is not accepted and don't displayed.

```html
<fluent-rating-display value="3.5"></fluent-rating-display>
<br />
<fluent-rating-display value="3,5"></fluent-rating-display>
```
image

dvoituron (approved):

dvoituron (approved):

vnbaaij (approved):

PR #3414, opened 2025-02-20
[Menu] Add CheckedChanged EventCallback (fix 3390)

### LINKED ISSUE #3390: feat: Add two-way databinding for Checked on FluentMenuItems
# 🙋 Feature Request

`FluentMenuItem` component has no two way databinding when using `@bind-Checked`.

## 🤔 Expected Behavior

The user should be able to control the `Checked` state of the `FluentMenuItem`.

## 😯 Current Behavior

You can only set the `Checked` property in the beginning then you have not information what state the checked flag is for a certain menu item.

## 💁 Possible Solution

Add the `CheckedChanged` event callback.

## 🔦 Context

I want to know which items are checked or not, and how that is changing.

rpodevns:
# Pull Request

## 📖 Description

Add CheckChanged EventCallback to FluentMenuItem and FluentMenu. Fix #3390 

## 👩‍💻 Reviewer Notes

Not sure if this is the best approach so needing feedback for sure.

### General

### Component-specific


src/Core/Components/Menu/FluentMenu.razor.cs
```diff
@@ -316,6 +322,24 @@ internal async Task SetOpenAsync(bool value)
         }
     }
```
rpodevns, line 324: I think there is an issue here.  The `fluent-menu` component break out of the for loop when a separator is encountered so this de-selection below of radio items is too global -- with respect to the `FluentMenu`, I think.

https://github.com/aaronburro/fast/blob/b78c921ec4e49ec9d7ec980f079ec114045df42e/packages/web-components/fast-foundation/src/menu/menu.ts#L244

rpodevns: Thanks for the link. I understood the same.  Would the `Dictionary` of menu items in `FluentMenu` need to be converted to an ordered list since Dictionaries are not guaranteed to be ordered and therefor no way to be certain which radio items belong to specific radio groups.

```csharp
private readonly Dictionary<string, FluentMenuItem> items = [];
```

Or, should we not try for 2-way binding and instead return the checked/unchecked item in the `FluentMenu` `CheckedChanged` callback?

The checkbox menu items were straightforward so I assumed I was finished .. and then I tested radio items :)

rpodevns: 2-way with checkbox items will work,  however,  the divider is not stored with the items in FluentMenu so no easy way to determine which radio items are grouped that I can see.  I'll give it some more thought.

vnbaaij: As this is new functionality, we can make it opt-in (with a parameter) and require users to add some extra information perhaps?

rpodevns: I'm thinking that it may 'just work' but can still add an opt-in parameter. Another hurdle is that all menu items are at the same level in the `items` `Dictionary` so I'm loading the full DOM hierarchy for the menu into an internal structure on first render (`FluentMenu.razor.cs`), and bindings are refreshed in the new `@onchange="@OnChangeHandlerAsync"` event added to `FluentMenuItem.razor.cs`.

Still more work to do.  Is this an okay approach?

```csharp
public class FluentMenuNode
{
    public string Id { get; set; } = "";
    public string Role { get; set; } = "";
    public string Tag { get; set; } = "";
    public string Label { get; set; } = "";
    public string Group { get; set; } = "";
    public bool Checked { get; set; } = false;
    public int HorizontalLevel { get; set; } = 0;
    public int VerticalLevel { get; set; } = 0;
    public List<FluentMenuNode> Children { get; set; } = [];
}
```

vnbaaij: Yes, looks solid!

src/Core/Components/Menu/FluentMenu.razor.js
```diff
@@ -1,4 +1,4 @@
-﻿// Add Left click event to open the PowerMenu
+// Add Left click event to open the PowerMenu
```
vnbaaij, line 1: nittpcik: change PowerMenu -> FluentMenu

vnbaaij (changes requested):

src/Core/Components/Menu/FluentMenu.razor.js
rpodevns: I wonder why that's showing as a diff.  I didn't add the Left/Right click features.

vnbaaij (approved):

vnbaaij:
@rpodevns Is this ready to go?

rpodevns:
The manual tests I did seem to work fine.  Did you have a change to try it?

Issue #1066, opened 2023-12-04
Data Grid with Virtualize and ItemsProvider causes double loading of data set


StephenOTT:
Using latest release.

If you create a data grid with a ItemsProvider and virtualize = true, it causes double requests of the underlying provider.

Taking the code for Remote:

https://www.fluentui-blazor.net/DataGrid#remotedata

and setting virtualized seems to cause the code GridItemsProvider to be called twice.

Is virtualize and ItemsProvider supposed to work together? or must code similar to the manual code for `Items` in the Virtualized section be used?

vnbaaij:
Hi,
While we recognize this might be a valid issue, there are no reprodcution steps provided. We have too much things going on to completely start from the top with every issue we receive.
As you are raising in issue (which we appreciate!), you clearly have an interest in making the library better for everyone. Please help us doing that by creating a clear and complete reproduction scenario. If needed, accompanied with a ready to run repository. Thanks!

StephenOTT:
@vnbaaij in general should itemsProvider work with virtualize with a single api call in initial load?  If yes, I am happy to provide a reproducer

vnbaaij:
I really don't know that from the top of my head. Would need to dive into the code for that. But it is working and used together in the sample you mentioned. 
I'm assuming you have tried it locally? If so, was that on Client or Server project?
If server, than it might just be a case of prerendering taking place and the mehod is called twice because of that. 
If client, let me know where in the sample I need to put a breakpoint to see it happen. I tried to set one on line 32 in `DataGridRemoteData.razor` but only saw that one being called once

StephenOTT:
Understood. Will setup a reproducer for you.

vnbaaij:
Closing due to inactivity. Can be re-opened once reproduction is available.

diego81b:
Hi,
i have the same problem so i take a look on code and at first glance, the problem seems to be that the `SetParametersAsync` method on `ColumnsCollectedNotifier` is invoked, which in turn invokes `RefreshDataCoreAsync`, but at that moment `_virtualizeComponent` (which should contain the reference to `Virtualize`) is not initialised.

As I understand it, the reference set occurs on `OnAfterRenderAsync` and that is when the second refresh is triggered as soon as Virtualize component is rendered.

I will try to investigate further but at the moment i didn't find a solution

StephenOTT:
Snippet to reproduce the error:

```razor

[...]

PR #3423, opened 2025-02-21
[DataGrid] Provide alternate way to refresh when working with remote data

EvandroMoSou:
# Pull Request

## 📖 Description

Using ItemsProvider and Virtualize like the sample in [DataGrid Remote Data](https://www.fluentui-blazor.net/datagrid-remote-data) the API is called 6 or 7 times. Although it has a CancellationToken in the request this is not enough to mitigate the API's resource consumption.

You can test using the code below on the screen, or seeing the calls to the api in the output:

    @inject HttpClient Http
    @inject NavigationManager NavManager
    
    <div style="height: 434px; overflow:auto;" tabindex="-1">
        <FluentDataGrid
                        ItemsProvider="foodRecallProvider"
                        OnRowDoubleClick="@(()=>DemoLogger.WriteLine("Row double clicked!"))"
                        Virtualize="true"
                        ItemSize="46"
                        GenerateHeader="GenerateHeaderOption.Sticky"
                        TGridItem="FoodRecall" >
            <PropertyColumn Title="ID" Property="@(c => c!.Event_Id)" />
            <PropertyColumn Property="@(c => c!.State)" Style="color: #af5f00 ;" />
            <PropertyColumn Property="@(c => c!.City)" />
            <PropertyColumn Title="Company" Property="@(c => c!.Recalling_Firm)" Tooltip="true"/>
            <PropertyColumn Property="@(c => c!.Status)" />
            <TemplateColumn Title="Actions" Align="@Align.End">
                <FluentButton aria-label="Edit item" IconEnd="@(new Icons.Regular.Size16.Edit())" OnClick="@(() => DemoLogger.WriteLine("Edit clicked"))" />
                <FluentButton aria-label="Delete item" IconEnd="@(new Icons.Regular.Size16.Delete())" OnClick="@(() => DemoLogger.WriteLine("Delete clicked"))" />
            </TemplateColumn>
        </FluentDataGrid>
    </div>
    
    <p>Total: <strong>@numResults results found</strong></p>
    <p>Call Times: <strong>@qtdCall</strong></p>
    
    
    @code {
        GridItemsProvider<FoodRecall> foodRecallProvider = default!;
        int? numResults;
	    int qtdCall;
    
        protected override async Task OnInitializedAsync()
        {
            // Define the GridRowsDataProvider. Its job is to convert QuickGrid's GridRowsDataProviderRequest into a query against
            // an arbitrary data soure. In this example, we need to translate query parameters into the particular URL format
            // supported by the external JSON API. It's only possible to perform whatever sorting/filtering/etc is supported
            // by the external API.
            foodRecallProvider = async req =>
            {
                qtdCall++;
    
                var url = NavManager.GetUriWithQueryParameters("https://api.fda.gov/food/enforcement.json", new Dictionary<string, object?>
                        {
                    { "skip", req.StartIndex },
                    { "limit", req.Count },
                        });
    
                var response = await Http.GetFromJsonAsync<FoodRecallQueryResult>(url, req.CancellationToken);
    
                // Simulate a slow data retrieval process
                if (req.Count is null)
                {
                    await Task.Delay(2500);
                }
                return GridItemsProviderResult.From(
                    items: response!.Results,
                    totalItemCount: response!.Meta.Results.Total);
            };
    
            // Display the number of results just for information. This is completely separate from the grid.
            numResults = (await Http.GetFromJsonAsync<FoodRecallQueryResult>("https://api.fda.gov/food/enforcement.json"))!.Meta.Results.Total;
        }
    }

After several failed attempts due to multiples refresh by QuickGrid and loss of sequence due to cancellation tokens. I decided to deal with the problem with a different approch.

Normally the API itself takes care of filtering, sorting and pagination, so a new type of grid update was created the 'RefreshItems'.

Basically you can define a method that will call the API.
This method is called once and only once when there is a change in the request (initial, pagination, ordering).
You must update the 'Items' and the 'SetTotalItemCountAsync'.

Along with the necessary commit, the page 'DataGridRemoteData.razor' was updated so you can see how it works.
The display of records, pagination and sorting are working.

Additionally, the option to force the datagrid refresh was added, so that if there is an API filter on the screen that has been modified, you can treat this refresh as a change in the requset.
And a new property in the column 'SortName' to facilitate the assembly of the request for the api.

### 🎫 Issues

[#1066](https://github.com/microsoft/fluentui-blazor/issues/1066)

## 👩‍💻 Reviewer Notes

Start the application and go to "/datagrid-remote-data" you can test here.

## 📑 Test Plan

I did it in a way that the old datagrid functions will not be affected if you have not defined the 'RefreshItems'.
Refresh force is optional and is set to false, so nothing changes for old things either.
And I believe that the new property in the column does not affect anything

### General

- [X ] I have tested my changes.

### Component-specific

## ⏭ Next Steps

Check the interaction of the grid's internal filtering with the new method


EvandroMoSou:
I agree with the CLA

EvandroMoSou:

vnbaaij (changes requested):
Thanks for your contribution! Sorry it took a while to review.

There is basically only 1 change I want to see: Create a new page for the altered DataGridRemoteData sample and leave the 'old' one as-is. Now that we have separate grid sample pages it is easy to have (and good) to have more. People can compare the code of both samples to find the differences. And maybe you can also explain what the differences are  in the sample page itself.

examples/Demo/Shared/SampleData/FoodRecall.cs
```diff
@@ -10,6 +10,7 @@ public class FoodRecall
     public string City { get; set; }
     public string State { get; set; }
     public string Recalling_Firm { get; set; }
+    public string Termination_Date { get; set; }
```
vnbaaij, line 13: What is the purpose of adding this date? If you do not need it for running your sample then please leave as-is

src/Core/Components/DataGrid/Columns/ColumnBase.razor.cs
```diff
@@ -93,6 +93,9 @@ public abstract partial class ColumnBase<TGridItem>
     [Parameter]
     public RenderFragment? ColumnOptions { get; set; }
```
vnbaaij, line 95: You need to add valid XML comment for this (it shows up in the demo site in the API documentation)

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -53,6 +53,9 @@ public partial class FluentDataGrid<TGridItem> : FluentComponentBase, IHandleEve
     [Parameter]
     public IQueryable<TGridItem>? Items { get; set; }
 
+    [Parameter]
```
vnbaaij, line 56: Also add valid summary comment here

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -691,24 +698,39 @@ private async Task RefreshDataCoreAsync()
             // (2) We won't know what slice of data to query for
             await _virtualizeComponent.RefreshDataAsync();
             _pendingDataLoadCancellationTokenSource = null;
+
+            StateHasChanged();
```
vnbaaij, line 702: Are you absolutely sure this `StateHasChanged` is needed here?

vnbaaij (changes requested):

examples/Demo/Shared/SampleData/FoodRecall.cs
EvandroMoSou: I use this field to demonstrate the ordering in my example.

src/Core/Components/DataGrid/Columns/ColumnBase.razor.cs
EvandroMoSou: Added

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
EvandroMoSou: Added

EvandroMoSou:
I reverted the old page, added the new example and fulfilled the requests mentioned.

If you have any other points to adjust, I am at your disposal.

vnbaaij (approved):

vnbaaij:
Thanks for a great contribution!

Just a heads-up, your `SortName` addition will get changed through the word done in #3460.

Issue #3430, opened 2025-02-22
fix: Deletion allowed in FluentSearch that is ReadOnly or Disabled


ASchweers:
# 🐛 Bug Report
In a `<FluentSearch>`-component, where the `Disabled` or `ReadOnly` parameter is set to true, the 'X'-button on the right hand side disappears. But the button is only hidden and still clickable, resulting in deletion of the value.

## 💻 Repro or Code Sample
See the *Search*-page of the demo and try the *States*-sample:
- type something in one of the first three `<FluentSearch>`-components
- now you can delete the value, if you click on the far right edge of the readonly or disabled `<FluentSearch>`-component.

## 🤔 Expected Behavior
Deletion shouldn't be possible in this scenario.

PascalVorwerk:
I have this fixed locally by adding a disabled attribute to the button if the Disabled or Readonly property on the components are either true. Setting a readonly attribute on the button didn't fix it for the readonly attribute.. cause a readonly button still handles clicks.

Let me know if you want me to open a PR for this!

https://github.com/user-attachments/assets/ecc5d4e9-df45-47c3-9c6e-bdc2a31203c8

I still need to add some tests aswell, but waiting to hear first

vnbaaij:
Yes, please create a PR for this. We can fix the unit tests. There are other components where we are adding attributes as well.

PR #3433, opened 2025-02-23
[FluentSearch] FluentSearch fix deletion if Disabled/Readonly

PascalVorwerk:
# Pull Request

## 📖 Description

I've added a javascript function that will add/remove the disabled option on the underlying shadow dom element that controls the clear button. Because the parent component can change this value, I was not sure if this check should be done each render or only on the first render, if so let me know!

For testing I found that the FluentSearch did not have any tests yet, so i have added a few. Sadly, the change i have made was hard to test because the property is only rendered in the shadowdom.. Some other attributes like 'AutoComplete' were also hard to test.

### 🎫 Issues

#3430 

## 👩‍💻 Reviewer Notes

So a few questions for reviewers:
- Should the disabled attribute be added/removed each render, or only on the first render? Right now the JSModule is imported each render.
- Is there any way to include tests for the newly added change? If so, i would love to learn how 😄 

## 📑 Test Plan

Run the project and fill in some value in on the FluentSearch page at the states section. Before you could delete any filled value by clicking where the 'button would be', but now you can't anymore.

### General


dvoituron:
Have you ever checked whether a CSS `pointer-event: none;` fixed this problem? If so, maybe it's simpler and avoids a JS call.

PascalVorwerk:
> Have you ever checked whether a CSS `pointer-event: none;` fixed this problem (on the "clear" button depending of a `fluent-search` attribute)? If so, maybe it's simpler and avoids a JS call.

Oh no i haven't... let me try try out! Although, when reading the docs i am wondering if a point event is equal to a onclick event.

PascalVorwerk:
> Have you ever checked whether a CSS `pointer-event: none;` fixed this problem (on the "clear" button depending of a `fluent-search` attribute)? If so, maybe it's simpler and avoids a JS call.

Okay so point-events: none does work, but how would i go about adding this conditionally in the css if it should only be added when disabled or readonly is set to true? Cause if we put it always, you can't clear anymore if the input is not disabled or readonly

dvoituron:
> > Have you ever checked whether a CSS `pointer-event: none;` fixed this problem (on the "clear" button depending of a `fluent-search` attribute)? If so, maybe it's simpler and avoids a JS call.
> 
> Okay so point-events: none does work, but how would i go about adding this conditionally in the css if it should only be added when disabled or readonly is set to true? Cause if we put it always, you can't clear anymore if the input is not disabled or readonly

Something like this...
```css
fluent-search[readonly]::part(clear-button),
fluent-search[disabled]::part(clear-button) {
    pointer-events: none;
}
```

PascalVorwerk:
> > > Have you ever checked whether a CSS `pointer-event: none;` fixed this problem (on the "clear" button depending of a `fluent-search` attribute)? If so, maybe it's simpler and avoids a JS call.
> > 
> > 
> > Okay so point-events: none does work, but how would i go about adding this conditionally in the css if it should only be added when disabled or readonly is set to true? Cause if we put it always, you can't clear anymore if the input is not disabled or readonly
> 
> Something like this...
> 
> ```css
> fluent-search[readonly]::part(clear-button),
> fluent-search[disabled]::part(clear-button) {
>     pointer-events: none;
> }
> ```

Thanks! It worked and I have pushed the changes.

src/Core/Components/Search/FluentSearch.razor.js
```diff
@@ -5,7 +5,6 @@ export function addAriaHidden(id) {
         fieldElement?.setAttribute("aria-hidden", "true");
     }
 }
-
```
dvoituron, line 8: If there are not change in this file, Can you undo this line suppression?

dvoituron (approved):

dvoituron (approved):

PR #3434, opened 2025-02-23
[dev-v5] Add FluentSwitch component

agriffard:
# [dev-v5] Add FluentSwitch component  

A **FluentSwitch** component represents a physical switch that allows someone to choose between two mutually exclusive options.  
For example, "On/Off" and "Show/Hide". Choosing an option should produce an immediate result.

image

## Best practices

**Layout**

When people need to perform extra steps for changes to take effect, use a check box instead. For example, if they must click a "Submit", "Next", or "OK" button to apply changes, use a check box.

**Content**

Only replace the On/Off labels if there are more specific labels for the setting. For example, you might use Show/Hide if the setting is "Show images".
Keep descriptive text short and concise—two to four words; preferably nouns. For example, "Focused inbox" or "WiFi".


agriffard:
Note: The LabelPosition example has a strange behaviour : The switches bound to the same value lose their synchronization, depending on which one is clicked.


https://github.com/user-attachments/assets/2011ef69-950f-44aa-b521-ceef837db609

dvoituron:
Indeed. You need to find why. 😀

tests/Core/Components.Tests.csproj
```diff
@@ -72,4 +72,31 @@
       <DependentUpon>FluentLocalizer.resx</DependentUpon>
     </Compile>
   </ItemGroup>
+
```
dvoituron, line 75: Can you remove all of these lines? Already include using a global pattern.

tests/Core/Components/Switch/FluentSwitchTests.razor
```diff
@@ -0,0 +1,80 @@
...
+@code
+{
+    public FluentSwitchTests()
+    {
+        JSInterop.Mode = JSRuntimeMode.Loose;
+        Services.AddFluentUIComponents();
+    }
+
+    [Fact]
+    public void FluentSwitch_Default()
```
dvoituron, line 17: Can you add a "default" test containing only the component (without parameter): `<FluentSwitch  />`
Mainly yo verify that non extra html attributes are present.

tests/Core/Components/Switch/FluentSwitchTests.razor
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    [Theory]
+    [InlineData(false, true)]
+    [InlineData(true, false)]
+    public void FluentSwitch_OnClick(bool initialValue, bool expectedValue)
+    {
+        // Arrange
+        var value = initialValue;
+        var cut = Render(@<FluentSwitch @bind-value="@value" />);
```
dvoituron, line 72: You need to use the PascalCase for the parameter: `@bind-Value`

src/Core/Components/Switch/FluentSwitch.razor.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Reflection.Metadata;
+using Microsoft.AspNetCore.Components;
+using Microsoft.AspNetCore.Components.Web;
+
+namespace Microsoft.FluentUI.AspNetCore.Components;
+
+/// <summary>
+/// The FluentSwitch component is used to render a Switch
+/// </summary>
+public partial class FluentSwitch : FluentInputBase<bool>, IFluentComponentElementBase
```
dvoituron, line 15: Why do you need to implement `IFluentComponentElementBase`. The `Element` parameter must be included only if needed (e.g. JS calls)

src/Core/Components/Switch/FluentSwitch.razor
```diff
@@ -0,0 +1,19 @@
+﻿@namespace Microsoft.FluentUI.AspNetCore.Components
+@using Microsoft.FluentUI.AspNetCore.Components.Extensions
+@inherits FluentInputBase<bool>
+
+<FluentField InputComponent="@this" ForId="@Id" Class="@ClassValue" Style="@StyleValue">
+    <fluent-switch @ref="@Element"
+                     aria-label="@AriaLabel"
```
dvoituron, line 7: Can you align all attributes with previous?

src/Core/Components/Switch/FluentSwitch.razor
```diff
@@ -0,0 +1,19 @@
...
+                     autofocus="@Autofocus"
+                     checked="@(CurrentValue)"
```
dvoituron, line 9: Why are you added parentheses?

src/Core/Components/Switch/FluentSwitch.razor.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    /// <summary>
+    ///
```
dvoituron, line 18: Can you add a comment?

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Switch/FluentSwitch.md
```diff
@@ -0,0 +1,29 @@
+---
+title: Switch
+route: /Switch
+---
+
+# Switch
+
+A **FluentSwitch** component represents a physical switch that allows someone to choose between two mutually exclusive options.  
+For example, "On/Off" and "Show/Hide". Choosing an option should produce an immediate result.
+
```
dvoituron, line 10: Can you add the "Best practices" from the React page, adapted to our Blazor components?
https://react.fluentui.dev/?path=/docs/components-switch--docs

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Switch/Examples/SwitchDisabled.razor
```diff
@@ -0,0 +1,7 @@
+﻿<FluentSwitch Label="Checked and disabled" Disabled="true" @bind-Value="@value1" />
+<FluentSwitch Label="Unchecked and disabled" Disabled="true" @bind-Value="@value2" />
+
```
dvoituron, line 3: Is it possible to use `ReadOnly`?

dvoituron (changes requested):

tests/Core/Components.Tests.csproj
agriffard: ok, lines removed from .csproj

tests/Core/Components/Switch/FluentSwitchTests.razor
agriffard: Simpler default test : `<FluentSwitch />`

tests/Core/Components/Switch/FluentSwitchTests.razor
agriffard: ok, changed to `@bind-Value`

src/Core/Components/Switch/FluentSwitch.razor
agriffard: ok, attributes aligned.

src/Core/Components/Switch/FluentSwitch.razor
agriffard: ok, parentheses removed.

src/Core/Components/Switch/FluentSwitch.razor.cs
agriffard: ok, comment added on ctor.

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Switch/FluentSwitch.md
agriffard: ok, "Best practices" added in doc.

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Switch/Examples/SwitchDisabled.razor
agriffard: I can add a Readonly example and tests but it does not prevent from switching it, it is just a little bit greyer.

image

Compare with disabled:

image

dvoituron: With ReadOnly the component cannot be changed. You need to find a way to do that. Maybe using a CSS style `pointer-events: none;` and 
`tabindex="-1"` (only if the `AdditionalAttributes` don't contain this key.

agriffard: `pointer-events: none;` added if `ReadOnly` is true (prevents from clicking on it).

src/Core/Components/Switch/FluentSwitch.razor
```diff
@@ -0,0 +1,18 @@
+﻿@namespace Microsoft.FluentUI.AspNetCore.Components
+@using Microsoft.FluentUI.AspNetCore.Components.Extensions
+@inherits FluentInputBase<bool>
+
+<FluentField InputComponent="@this" ForId="@Id" Class="@ClassValue" Style="@((ReadOnly ? "pointer-events: none;" : "") + StyleValue)">
```
dvoituron, line 5: You can't hard-code your style: you have to use the `DefaultStyleBuilder` as we have done in other components (ex. FluentButton).
But in this case, this style is pre-defined" ans always the same: you need to add a  FluentSwitch.razor.css` containing `fluent-switch[readonly] { }`.

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Switch/FluentSwitch.md
dvoituron: We don't do that in other .md files: please don't use `###` but `**Layout**` and  `**Content**`.

src/Core/Components/Switch/FluentSwitch.razor
```diff
@@ -0,0 +1,18 @@
...
+    <fluent-switch aria-label="@AriaLabel"
+                   autofocus="@Autofocus"
+                   checked="@CurrentValue"
+                   disabled="@Disabled"
+                   id="@Id"
+                   name="@Name"
```
dvoituron, line 11: You need to add a `tabindex="-1"` if `ReadOnly=true` and `!AdditionalAttributes?.ContainsKey("tabindex")`.
If this is not the case, the user can navigate to the component using the keyboard.

src/Core/Components/Switch/FluentSwitch.razor
```diff
@@ -2,7 +2,7 @@
 @using Microsoft.FluentUI.AspNetCore.Components.Extensions
 @inherits FluentInputBase<bool>
 
-<FluentField InputComponent="@this" ForId="@Id" Class="@ClassValue" Style="@((ReadOnly ? "pointer-events: none;" : "") + StyleValue)">
```
dvoituron: `AdditionalAttributes` is a the `fluent-switch` attribute. You therefore need to check whether **tabindex** is already in it, because if dev specifies a value for this attribute, you can't have it twice in your component. If `ReadOnly` is activated and there is no tabindex in this list, then you need to add `tabindex="-1"`, otherwise you can use the keyboard to access the component. The pointer-events style is for mouse click only. Go to `FluentAvatar.razor.cs` where this has already been done.

src/Core/Components/Switch/FluentSwitch.razor.css
```diff
@@ -0,0 +1,3 @@
+fluent-field[readonly]:has(fluent-switch) {
+  pointer-events: none;
```
dvoituron, line 2: This style must be applied to µ fluent-switch` and not `fluent-field` => `fluent-switch[readonly] { pointer-events: none; }`

dvoituron (changes requested):

tests/Core/Components/Switch/FluentSwitchTests.FluentCheckbox_ReadOnly.verified.razor.html
```diff
@@ -1,5 +1,5 @@
 
-<fluent-field id="xxx" label-position="after" style="pointer-events: none;">
+<fluent-field id="xxx" label-position="after" readonly="">
```
dvoituron, line 2: You already have a `readonly` attribute on the `fluent-switch` :-(

src/Core/Components/Switch/FluentSwitch.razor
agriffard: Style removed from field.

src/Core/Components/Switch/FluentSwitch.razor
agriffard: `GetTabIndexValue` as in `FluentAvatar.razor.cs`

src/Core/Components/Switch/FluentSwitch.razor.css
agriffard: ok, done but it doesn't have the expected behaviour in the example.

tests/Core/Components/Switch/FluentSwitchTests.FluentCheckbox_ReadOnly.verified.razor.html
agriffard: Yes, but it doesn't seem to be taken into account.

agriffard:
> Indeed. You need to find why. 😀

Same behaviour with 3 Checkboxes bound to @value1 (Only two components are enough to show this de-synchronization effect).

The html of all components is corresponding to the state of the `@value1` (`checked` added or removed) but the components are not rendered accordingly.

agriffard:
> 1. I think these attributes could be removed from the new component.
> 2. But to keep a compatibility, you could create another `partial class` in the Migration folder with these 3 parameters flagged as `Obsolete` and include it in the component: updat the `Label` property with `CheckedMessage` or `UncheckedMessage` if the `Label` is initially `null` and if the `LabelPosition` is only `After`. You must document all of this on the Migration page.

[Obsolete] properties added in partial class and FluentSwitch added to Migration.

I could use this in the field for the label:
`Label="@(!string.IsNullOrEmpty(CheckedMessage) && CurrentValue ? CheckedMessage : (!string.IsNullOrEmpty(UncheckedMessage) && !CurrentValue ? UncheckedMessage : Label))"`
but as the Properties are Obsolete, it doesn't compile when I use them.

src/Core/Migration/FluentSwitch.cs
```diff
@@ -0,0 +1,31 @@
+// ------------------------------------------------------------------------
+// MIT License - Copyright (c) Microsoft Corporation. All rights reserved.
+// ------------------------------------------------------------------------
+
+using Microsoft.AspNetCore.Components;
+
+namespace Microsoft.FluentUI.AspNetCore.Components;
+
+public partial class FluentSwitch
```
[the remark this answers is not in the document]
agriffard: I added a `SwitchMessage` example but I had to comment the [Obsolete] attributes on `CheckedMessage` and `UncheckedMessage` unless it won't compile.
When set, the `Label` is overridden with their values based on the checked state.

About `ChildContent`, indeed, I can't handle it.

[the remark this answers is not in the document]
dvoituron: Nice. But can you create a private method `GetLabel` to have a maximum of C# code in the .cs file ?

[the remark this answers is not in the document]
agriffard: ok, GetLabel added.

tests/Core/Components/Switch/FluentSwitchTests.razor
```diff
@@ -0,0 +1,88 @@
...
+    public void FluentCheckbox_LabelPosition()
+    {
+        // Arrange && Act
+        var cut = Render(
+            @<FluentSwitch LabelPosition="LabelPosition.Above" />
+        );
+
+        // Assert
+        cut.Verify();
+    }
```
vnbaaij, line 35: When an enum parameters is involved, we want to have tests for all possible values (including `null` and a random numeric value). You need to use [Theory] and [InlineData()] and `cut.Verify(suffix:...)`

See for an example the `FluentCounterBadge_Positioning` test.

vnbaaij (changes requested):

tests/Core/Components/Switch/FluentSwitchTests.razor
agriffard: ok, Tests added with InlineData on LabelPosition.

src/Core/Components/Switch/FluentSwitch.razor.css
```diff
@@ -0,0 +1,3 @@
+fluent-switch[readonly] {
```
dvoituron, line 1: Have you checked that the ReadOnly parameter blocks the use of the mouse and keyboard?

I think you need to apply this CSS style (on the “field” element and not on the “switch” element):
```css
fluent-field:has(fluent-switch[readonly]) {
  pointer-events: none;
}
```

And to block the keyboard, you need to set the `tabindex` on the `fluent-switch` (like done) but also on the `FluentField` component. If not, using `Tab` the user can navigate to the component and press `Space`

examples/Demo/FluentUI.Demo.Client/Documentation/GetStarted/Migration/MigrationFluentSwitch.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Migration FluentSwitch
+route: /Migration/Switch
+hidden: true
+---
+
+### Removed values💥
+The `ChildContent` property has been removed. Use `FluentField` instead.
```
dvoituron, line 8: Can you replace with
> The `ChildContent` property has been removed. Use `Label` parameter instead.

src/Core/Migration/FluentSwitch.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// </summary>
+    [Parameter]
+    [Obsolete("This property is not supported anymore and will be removed in a future release.")]
+    public RenderFragment? ChildContent { get; set; }
+
+    /// <summary>
+    /// Gets or sets the checked message
+    /// </summary>
+    [Parameter]
+    //[Obsolete("This property is not supported anymore and will be removed in a future release.")]
```
dvoituron, line 22: Can you uncomment these `Obsolete` attribute?

tests/Core/Components/Switch/FluentSwitchTests.razor
```diff
@@ -0,0 +1,103 @@
+﻿@using Microsoft.FluentUI.AspNetCore.Components.Extensions
+@using Microsoft.FluentUI.AspNetCore.Components.Tests.Extensions
+@using Microsoft.FluentUI.AspNetCore.Components.Utilities
+@using System.ComponentModel.DataAnnotations
+@using Xunit;
+@inherits TestContext
+
+@code
+{
+    public FluentSwitchTests()
```
dvoituron, line 10: The Obsolete attributes (ChildContent, CheckedMessage and UncheckedMessage) must be unit tested.

You can run the file `_StartCodeCoverage.cmd` to validate if all code lines a covered.

tests/Core/Components/Switch/FluentSwitchTests.razor
```diff
@@ -0,0 +1,103 @@
...
+    public void FluentSwitch_Default()
+    {
+        // Arrange && Act
+        var cut = Render(@<FluentSwitch />);
+        // Assert
+        cut.Verify();
+    }
+
+    [Fact]
+    public void FluentCheckbox_LabelPosition()
```
dvoituron, line 26: This test can be omitted: you'll already be testing this position on the next next.

dvoituron (changes requested):

src/Core/Migration/FluentSwitch.cs
agriffard: If there is an `[Obsolete]`, I doesn't compile anymore, with this message:

`'FluentSwitch.UncheckedMessage' is obsolete: 'This property is not supported anymore and will be removed in a future release.'`

dvoituron: You need to add a `pragma` condition. We already used this in the code.

tests/Core/Components/Switch/FluentSwitchTests.razor
agriffard: Ok, `FluentCheckbox_LabelPosition` single test removed.

examples/Demo/FluentUI.Demo.Client/Documentation/GetStarted/Migration/MigrationFluentSwitch.md
agriffard: Ok, documentation for migration updated.

tests/Core/Components/Switch/FluentSwitchTests.razor
agriffard: Ok, Tests added for ChildContent, CheckedMessage and UncheckedMessage.

src/Core/Components/Switch/FluentSwitch.razor.css
agriffard: Ok, css added for readonly and tabIndex on FluentField.

tests/Core/Components/Switch/FluentSwitchTests.razor
agriffard: More tests added and `_StartCodeCoverage.cmd` run. 100% on branch coverage and a bit like `FluentCheckBox` on Line coverage.

dvoituron (approved):

vnbaaij (approved):

agriffard:
Thank you. Please tell me if you need help on migrating other components.

PR #3453, opened 2025-02-26
[DataGrid] Only add Selectable items when using Select All

### LINKED ISSUE #3447: fix: SelectAll selects all items even if unselectable due to Selectable in DataGrid's SelectColumn
# 🐛 Bug Report

When Selectable is set and multiple rows are not selectable, select all still selects every item in the Datagrid.

## 💻 Repro or Code Sample

```
<FluentDataGrid
    TGridItem="Test"
    Items="Data"
    OnRowClick="PrintSelected">
    <ChildContent>
        <SelectColumn
            TGridItem="Test"
            Selectable="@(test => test.Count > 2)" 
            SelectMode="DataGridSelectMode.Multiple" 
            SelectFromEntireRow="false" 
            @bind-SelectedItems="_selected"/>
        <PropertyColumn Property="@(t => t.Count)" Title="Count" Sortable="true"/>
    </ChildContent>
</FluentDataGrid>

@code
{
    private class Test
    {
        public int Count { get; init; }
    }

    private IQueryable<Test> Data { get; set; } = new List<Test>()
    {
        new Test { Count = 1 }, 
        new Test { Count = 2 },
        new Test { Count = 3 },
        new Test { Count = 4 },
        new Test { Count = 5 }
    }.AsQueryable();

    private IEnumerable<Test> _selected = new List<Test>();

    private void PrintSelected()
    { 
        Console.WriteLine(_selected.Count()); 
    }
}
```

## 🤔 Expected Behavior

When select is clicked, only the items that are deemed selectable by the Selectable function should be added to SelectedItems.

## 😯 Current Behavior

Currently all values are added to SelectedItems even if they are not selectable.

## 💁 Possible Solution

From looking through the code, the solution seems to be to add another check for if Selectable is set and if true do something like this `_selectedItems.AddRange(InternalGridContext.Grid.Items?.Where(Selectable).ToArray() ?? InternalGridContext.Items);` on the OnClickAllAsync method;

```
internal async Task OnClickAllAsync(MouseEventArgs e)
    {
        if (Grid == null || SelectMode != DataGridSelectMode.Multiple || SelectAllDisabled)
        {
            return;
        }

        // SelectAllChanged
        SelectAll = !(SelectAll == true);
        if (SelectAllChanged.HasDelegate)
        {
            await SelectAllChanged.InvokeAsync(SelectAll);
        }

        // SelectedItems
        _selectedItems.Clear();
        if (SelectAll == true)
        {
            _selectedItems.AddRange(InternalGridContext.Grid.Items?.ToArray() ?? InternalGridContext.Items);
        }

        if (SelectedItemsChanged.HasDelegate)
        {
            await SelectedItemsChanged.InvokeAsync(SelectedItems);
        }

        RefreshHeaderContent();
    }
```

## 🔦 Context

It simply adds items that are not applicable for selection, which means that backend calls can fail due to invalid data.

## 🌍 Your Environment

* OS & Device: macOS Sequoia 15.3.1 on an M2 MacBook Air
* Browser: Microsoft Edge
* .NET and Fluent UI Blazor library Version:  8.0 and 4.11.5

vnbaaij:
Fix #3447 

Add extra check when selecting all items so only items that are actually selectable get selected.

### DIFF
--- a/src/Core/Components/DataGrid/Columns/SelectColumn.cs
+++ b/src/Core/Components/DataGrid/Columns/SelectColumn.cs
@@ -534,9 +534,8 @@ public class SelectColumn<TGridItem> : ColumnBase<TGridItem>
         {
             // Only add selectable items
             _selectedItems.AddRange((InternalGridContext.Grid.Items?.ToList() ?? InternalGridContext.Items)
-                .Where(item => Selectable?.Invoke(item) ?? false)
-                .ToArray()
-);
+                .Where(item => Selectable?.Invoke(item) ?? true)
+            );
         }
 
         if (SelectedItemsChanged.HasDelegate)
--- a/src/Core/Components/DataGrid/Columns/SelectColumn.cs
+++ b/src/Core/Components/DataGrid/Columns/SelectColumn.cs
@@ -1,3 +1,7 @@
+// ------------------------------------------------------------------------
+// MIT License - Copyright (c) Microsoft Corporation. All rights reserved.
+// ------------------------------------------------------------------------
+
 using Microsoft.AspNetCore.Components;
 using Microsoft.AspNetCore.Components.Rendering;
 using Microsoft.AspNetCore.Components.Web;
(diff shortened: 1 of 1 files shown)


src/Core/Components/DataGrid/Columns/SelectColumn.cs
```diff
@@ -528,7 +532,11 @@ internal async Task OnClickAllAsync(MouseEventArgs e)
         _selectedItems.Clear();
         if (SelectAll == true)
         {
-            _selectedItems.AddRange(InternalGridContext.Grid.Items?.ToArray() ?? InternalGridContext.Items);
+            // Only add selectable items
+            _selectedItems.AddRange((InternalGridContext.Grid.Items?.ToList() ?? InternalGridContext.Items)
+                .Where(item => Selectable?.Invoke(item) ?? false)
+                .ToArray()
+);
```
dvoituron, line 539: Can you align this end if line with the previous command?

dvoituron (approved):

PR #3460, opened 2025-03-01
[DataGrid] Add IGridSort interface, fixes sorting when using ItemsProvider

### LINKED ISSUE #3290: feat: GridSort somewhat incompatible with ItemsProvider
# 🙋 Feature Request

GridSort assumes the sorting is only possible on columns where the column expression is a `MemberExpression` to be able to apply it to an IQueryable.

This means that any model used with an `ItemsProvider` must provider a Member for any sortable property, and that it must somehow map those member names to a sorting method.

## 💁 Possible Solution

I see 2 possible solutions for this:

a. Add `GridSort` constructor that allows a arbitrary key to be used as `PropertyName`, so that `SortedProperty.PropertyName` can be mapped to the custom sorting logic.

b. Add an optional `ColumnKey` to each column and allow GridSort to use null for `PropertyName`, but add the `ColumnKey` in `SortedProperty`.

## 🔦 Context

Have a complex code base where the underlying db model does not support IQueryable and columns are very dynamic.

## 💻 Examples

```razor
<FluentDataGrid ItemsProvider="@ItemsProvider" >
@foreach (var colDef in UserColumns) {
 <PropertyColumn Title="@colDef.CustomUserPropertyName" Property="@(x => colDef.GetValueFor(x))" Sortable="true" SortBy="NotPossible" />
}
</FluentDataGrid>
```

StevenRasmussen:
Fixes #3290 

Before writing unit tests etc... I wanted to get feedback on whether this would be an acceptable approach.

# Pull Request

### General

### Component-specific


vnbaaij (commented):
Apart from the namespace name and whitespace remarks, think this looks quite good already. 👍

I don't understand the StaticGridSort purpose. Is it the default IGridSort that gets applied. I don't see it called anywhere (yet?). If it is indeed the default, than I'd suggest to name it like that as well `DefaultGridSort`

StevenRasmussen:
@vnbaaij - Thanks for the feedback!  It was late last night when I created this and so I was struggling to come up with a good name for the new class.  I'm still not certain that `ColumnKeyGridSort` is the best... but I do think it more closely represents its purpose now.  I've added a demo page and so you can see it in action and what problem it is solving.  Unless you need anything else I think the PR should be ready for merging.  Thanks!

davhdavh:
Very nice way of solving the issue

vnbaaij (approved):

PR #3469, opened 2025-03-04
[dev-v5] Initial work for Option and Select implementation

vnbaaij:
- Update WC v3 version
- Initial work for FluentOption
- Initial work for FluentSelect
- Add Slot to FluentAvatar
- Fix and update tests


src/Core/Components/List/FluentOption.razor.cs
```diff
@@ -45,6 +45,12 @@ public partial class FluentOption : FluentComponentBase
     [Parameter]
     public EventCallback<bool> SelectedChanged { get; set; }
 
+    /// <summary>
+    /// Gets or sets the content to display in the description slot.
+    /// </summary>
+    [Parameter]
+    public string? Description { get; set; }
```
dvoituron, line 52: What is the difference between ChildContent and Description ? Maybe we can add this difference in the doc?

tests/Core/Components/Checkbox/FluentCheckboxTests.FluentCheckbox_Default.verified.razor.html
```diff
@@ -1,4 +1,5 @@
-<fluent-field id="xxx" label-position="after">
-    <fluent-checkbox aria-label="My aria label" autofocus="" checked="true" indeterminate="false" id="xxx" name="xxx" shape="square" size="medium" slot="input"></fluent-checkbox>
-    <fluent-text as="span" size="200" slot="message"></fluent-text>
-</fluent-field>
+
+<fluent-field id="xxx" label-position="after" size="medium">
```
dvoituron, line 2: Why the FluentCheckbox tests changed?

tests/Core/Components/List/FluentSelectTests.razor
```diff
@@ -45,9 +45,9 @@
         // Arrange and Act
         var cut = Render(
     @<FluentSelect TOption="string">
-        <FluentOption Value="One">One</FluentOption>
-        <FluentOption Value="Two">Two</FluentOption>
-        <FluentOption Value="Three">Three</FluentOption>
+        <FluentOption Value="@("One")">One</FluentOption>
```
dvoituron, line 48: Why did you update `"One"` to `"@("One")"`?

dvoituron (changes requested):

src/Core/Components/List/FluentOption.razor.cs
vnbaaij: See https://web-components.fluentui.dev/?path=/docs/components-dropdown-option--docs#start-content for the difference. Description is shown as an extra line under the ChildContent

tests/Core/Components/Checkbox/FluentCheckboxTests.FluentCheckbox_Default.verified.razor.html
vnbaaij: The Field/label did not take the Size into account yet.

tests/Core/Components/List/FluentSelectTests.razor
vnbaaij: FluentOption is now generic. When using string as the type, you need to specify the values like this. But made some more changes so will check if it is still necessary

src/Core/Components/List/FluentOption.razor.cs
dvoituron: Probably better to have a 
`RenderFragment` property? And a name like `SubChildContent` or `DescriptionTemplate`.

vnbaaij: Then we would enable people to do things like this:

image

Think it is better to keep it a simple string. Think of it like the `Message` we have on the `FluentField`

tests/Core/Components/Checkbox/FluentCheckboxTests.FluentCheckbox_Default.verified.razor.html
vnbaaij: I did not make a change in `FluentField`.  I added `slot` (==AdditionalAttribute) in `FluentCheckBox.razor`. So the rendered html is different.

src/Core/Components/List/FluentOption.razor.cs
dvoituron: Make sense 😀. And the dev can use the `ChildContent` to format the content. 

Can you add an explanation in this property doc: to explain that is a message below the `ChildContent` content.

tests/Core/Components/Checkbox/FluentCheckboxTests.FluentCheckbox_Default.verified.razor.html
vnbaaij: I actually made the change outside of this PR already but didn't update the verified files...doing that in `dev-v5` now

PR #3507, opened 2025-03-13
[Docs] Add cookie consent 

vnbaaij:
This PR adds a cookie consent banner and functionality to the Demo & Documentation site, both for the (live) Wasm site and for the Server version of the site. The content and options are according to the Microsoft guidelines. 

image

There are three categories that can be consented to: analytic cookies, social media cookies and advertising cookies (as per the Microsoft standard) The settings following the choices made will be put in local storage. When consent is provided for analytics, the Google Analytics and Microsoft Clarity scripts will be added to the page and cookies will be written. Currently, we do not do anything with the social media and advertising consent.
`Accept` gives consent for all three categories. `Reject` does not provide consent  to any of the categories. `Manage cookies` allows for providing consent to each category individually:

image

When later, a user wants to revise their consent settings, the Settings panel now offers a `Manage cookies` option as well:

image

### DIFF
--- a/examples/Demo/Shared/Components/Cookies/ManageCookies.razor
+++ b/examples/Demo/Shared/Components/Cookies/ManageCookies.razor
@@ -52,12 +52,12 @@
     public CookieState Content { get; set; } = default!;
 
     [CascadingParameter]
-    public FluentDialog? Dialog { get; set; }
+    public FluentDialog? Dialog { get; set; } = default!;
 
 
     private async Task HandleSaveAsync()
     {
-        await Dialog!.CloseAsync(Content);
+        await Dialog.CloseAsync(Content);
     }
 
     private async Task HandleCancelAsync()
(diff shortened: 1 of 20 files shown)


examples/Demo/Shared/Components/Cookies/ManageCookies.razor
```diff
@@ -0,0 +1,70 @@
...
+    [Parameter]
+    public CookieState Content { get; set; } = default!;
+
+    [CascadingParameter]
+    public FluentDialog? Dialog { get; set; }
+
+
+    private async Task HandleSaveAsync()
+    {
+        await Dialog!.CloseAsync(Content);
```
dvoituron, line 60: I presume that the Dialog will always defined?
So, you can set `public FluentDialog? Dialog { get; set; } = default!;` and remove the `!` for `Dialog!...`

dvoituron (changes requested):

dvoituron (approved):

PR #3519, opened 2025-03-14
[FluentSlider] Refactoring

dvoituron:
# [FluentSlider] Refactoring

- Update the documentation
- Remove the code to "manage" the `ReadOnly` attribute.
   And add this explaination:
   > The FluentSlider component does not support the ReadOnly attribute (using the keyboard), like the default 
   > [input type="range"](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range) HTML element. 
   > The ReadOnly attribute is not applicable to the slider because it is designed for text-based inputs, and sliders are 
   > primarily used for selecting a value within a range
   > The Disabled attribute is used to prevent user interaction with the slider, while ReadOnly would imply that the slider 
   > can still be interacted with but not modified, which doesn't align with the intended use of a slider.
- Remove unused `.verified` files


examples/Demo/FluentUI.Demo.Client/Documentation/Components/Slider/FluentSlider.md
```diff
@@ -26,20 +26,42 @@ The **FluentSlider** component is an input where the user selects a value from w
...
-{{ SliderLabel }}
+{{ SliderReadOnlyDisabled }}
+
+> The **FluentSlider** component does not support the `ReadOnly` attribute (using the keyboard),
+> like the default [input type="range"](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range) HTML element.
+> The `ReadOnly` attribute is not applicable to the slider because it is designed for text-based inputs,
+> and sliders are primarily used for selecting a value within a range
+>
+> The `Disabled` attribute is used to prevent user interaction with the slider, while `ReadOnly` would imply
+> that the slider can still be interacted with but not modified, which doesn't align with the intended use of a slider.
```
vnbaaij, line 53: > The **FluentSlider** component does not support the `ReadOnly` attribute. This is in line with the HTML standard that prescribes only inputs that accept text entry may be readonly. See  the [HTML spec](https://html.spec.whatwg.org/multipage/input.html#the-readonly-attribute) and the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range) for more information.
>
> The `Disabled` attribute is used to prevent user interaction with the slider.

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Slider/FluentSlider.md
```diff
@@ -52,17 +74,10 @@ If you add content that is **not part** of `Slot="thumb"`, it will be ignored.
...
-
-## EventCallback
-
-{{ SliderEventCallbacks }}
-
 ## Know restrictions
 
-At this point, the `FluentSliderLabel` elements cannot be applied.
+At this moment, the `FluentSliderLabel` elements, included in the v4 library, are not yet supported.
+The `FluentSliderLabel` elements are used to display the value of the slider steps, below the slider.
```
vnbaaij, line 80: The `FluentSliderLabel` component (used to display the value of the slider steps) from version 4 is no longer supported.

src/Core/Components/Slider/FluentSlider.razor.cs
```diff
@@ -41,60 +37,43 @@ public FluentSlider()
...
     /// <summary>
-    /// Gets or sets the slider's step value.
+    /// Gets or sets the slider's step value. Default is 1.
     /// </summary>
     [Parameter]
     public TValue? Step { get; set; }
 
     /// <summary>
-    /// Gets or sets the orientation of the stacked components.
+    /// Gets or sets the orientation of the stacked components. Default is <see cref="Orientation.Horizontal"/>.
```
vnbaaij, line 65: /// Gets or sets the orientation of the component. Default is <see cref="Orientation.Horizontal"/>.

vnbaaij (changes requested):

vnbaaij (approved):

PR #3547, opened 2025-03-20
[dev-v5] Add FluentCompoundButton

vnbaaij:
Ads the `FluentCompoundButton` component

A compound button renders a label and a description line in a button. 

The actual text for the button can be provided in 2 ways:
1) With the `Label` (string?) parameter.
2) With the `ChildContent` parameter.
We added this so you do not always have to supply a label in the form of a `<ChildContent>xyz</ChildContent>` form. When supplying both, both will be rendered.

The `Description` parameter is marked as `EditorRequired` because we assume you'd use a regular button if you do not want to use a description.

image

Tests have been added:

image

### DIFF
--- a/examples/Demo/FluentUI.Demo.Client/Documentation/Components/CompoundButton/Examples/CompoundButtonDefault.razor
+++ b/examples/Demo/FluentUI.Demo.Client/Documentation/Components/CompoundButton/Examples/CompoundButtonDefault.razor
@@ -1,0 +1,34 @@
+﻿<p>
+    You can use <code>Content</code> and/or <code>ChildContent</code> to specify the main text on the button.
+</p>
+<p>
+    When you supply both, both will be rendered.
+</p>
+
+<FluentStack HorizontalGap="10">
+    <div>
+        <strong>ChildContent</strong>
+        <br />
+        <FluentCompoundButton>
+            <ChildContent>Child content</ChildContent>
+            <Description>Description content</Description>
+        </FluentCompoundButton>
+    </div>
+
+    <div>
+        <strong>Label</strong>
+        <br />
+        <FluentCompoundButton Label="Label">
+            <Description>Description content</Description>
+        </FluentCompoundButton>
+    </div>
+
+    <div>
+        <strong>Both</strong>
+        <br />
+        <FluentCompoundButton Label="Both">
+            <ChildContent>Both</ChildContent>
+            <Description>Description content</Description>
+        </FluentCompoundButton>
+    </div>
+</FluentStack>
--- a/examples/Demo/FluentUI.Demo.Client/Documentation/Components/CompoundButton/FluentCompoundButton.md
+++ b/examples/Demo/FluentUI.Demo.Client/Documentation/Components/CompoundButton/FluentCompoundButton.md
@@ -1,0 +1,29 @@
+---
+title: CompoundButton
+route: /CompoundButton 
+---
+
+# CompoundButton
+
+## Default
+
+{{ CompoundButtonDefault }}
+
+## Appearances
+{{ CompoundButtonAppearances }}
+
+## Shapes
+{{ CompoundButtonShapes }}
+
+## Sizes
+{{ CompoundButtonSizes }}
+
+## Disabled
+{{ CompoundButtonDisabled }}
+
+## Long text
+{{ CompoundButtonLongText }}
+
+## API FluentCompoundButton
+
+{{ API Type=FluentCompoundButton }}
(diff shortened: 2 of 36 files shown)


examples/Demo/FluentUI.Demo.Client/Documentation/Components/CompoundButton/Examples/CompoundButtonDefault.razor
```diff
@@ -0,0 +1,34 @@
+﻿<p>
```
dvoituron, line 1: The razor files must contain only the source code, not the documentation.
The documentation must be included in the `FluentCompoundButton.md` file.

examples/Demo/FluentUI.Demo.Client/Documentation/Components/CompoundButton/FluentCompoundButton.md
```diff
@@ -0,0 +1,29 @@
+---
+title: CompoundButton
+route: /CompoundButton 
+---
+
+# CompoundButton
+
+## Default
+
+{{ CompoundButtonDefault }}
```
dvoituron, line 10: I don't think it's necessary to explain: a Blazor developer already knows how to use a `Label` parameter or a `ChildContent` parameter. The description of `Label` is sufficient to explain the difference between `ChildContent` and `Label`.

I would remove the Default.razor file, which seems to provide nothing.

src/Core/Components/Button/FluentCompoundButton.razor.cs
```diff
@@ -0,0 +1,162 @@
...
+    ///  This can be used as an alternative to specifying the content as a child component of the button.
+    ///  If both are specified, both will be rendered.
+    ///  </summary>
+    [Parameter]
+    public string? Label { get; set; }
+
+    /// <summary>
+    /// Gets or sets the content to be rendered inside description area of the button.
+    /// </summary>
+    [Parameter, EditorRequired]
```
dvoituron: This property doesn't have to be required: the component displays just fine without `Description`, and I think it can be useful in certain cases.

src/Core/Components/Button/FluentCompoundButton.razor.cs
```diff
@@ -0,0 +1,162 @@
...
+        }
+
+        await Task.CompletedTask;
+    }
+
+    /// <summary>
+    /// Disables the button.
+    /// </summary>
+    /// <param name="disabled">True to disable the button</param>
+    public void SetDisabled(bool disabled)
```
dvoituron, line 157: I think we need to look to the future and offer only `Async` methods (even if they're not currently useful).

tests/Core/Components/CompoundButton/FluentCompoundButtonTests.razor
```diff
@@ -0,0 +1,366 @@
...
+    {
+        JSInterop.Mode = JSRuntimeMode.Loose;
+        Services.AddFluentUIComponents();
+    }
+
+    [Fact]
+    public void FluentCompoundButton_Default()
+    {
+        // Arrange
+        using var id = Identifier.SequentialContext();
```
dvoituron, line 17: The ID is not used. You can remove this line.

tests/Core/Components/CompoundButton/FluentCompoundButtonTests.razor
```diff
@@ -0,0 +1,366 @@
...
+
+        // Act
+        cut.Find("fluent-compound-button").Click();
+
+        // Assert
+        Assert.True(clicked);
+    }
+
+    [Fact]
+    public void FluentCompoundButton_ClassAttribute()
```
dvoituron, line 328: Not a problem to have the tests on Class, Style, AdditionalAttribute... But these parameters are automatically validated by the Global test :-)

dvoituron (changes requested):

src/Core/Components/Button/FluentCompoundButton.razor.cs
vnbaaij: Ok, but then this needs to be looked at for Button as well (where I copied it from :))

tests/Core/Components/CompoundButton/FluentCompoundButtonTests.razor
vnbaaij: Copilot generated...😁

src/Core/Components/Button/FluentCompoundButton.razor.cs
dvoituron: It needs a new PR :-)

i can do that if you prefer.

examples/Demo/FluentUI.Demo.Client/Documentation/Components/CompoundButton/Examples/CompoundButtonDefault.razor
vnbaaij: Including the sentence about label/childcontent in the doc

src/Core/Components/Button/FluentCompoundButton.razor.cs
vnbaaij: Yes, lets do all of them in a new PR then

PR #3557, opened 2025-03-21
[dev-v5] Add FluentToggleButton

vnbaaij:
Ads the `FluentToggleButton` component

A toggle button is a two-state behavior that switches a state between on and off. A button’s rest state indicates “off” (`Pressed="false"`) while the selected state indicates “on” (`Pressed="true"`). The toggle state is in addition to the normal click behavior of a button.

image

### DIFF
--- a/src/Core/Components/Button/FluentToggleButton.razor.cs
+++ b/src/Core/Components/Button/FluentToggleButton.razor.cs
@@ -1,0 +1,34 @@
+// ------------------------------------------------------------------------
+// MIT License - Copyright (c) Microsoft Corporation. All rights reserved.
+// ------------------------------------------------------------------------
+
+using Microsoft.AspNetCore.Components;
+using Microsoft.AspNetCore.Components.Web;
+
+namespace Microsoft.FluentUI.AspNetCore.Components;
+
+/// <summary>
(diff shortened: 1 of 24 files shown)


src/Core/Components/Button/FluentToggleButton.razor.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// <summary>
+    /// Gets or sets the pressed state of the button.
+    /// </summary>
+    [Parameter]
+    public bool Pressed { get; set; }
+
+    /// <summary />
+    protected new async Task OnClickHandlerAsync(MouseEventArgs e)
+    {
+        Pressed = !Pressed;
```
dvoituron, line 31: You have a `pressedChanged` event for this web component. Maybe you could use it to be sure the `Pressed` value is always correct?

dvoituron (changes requested):

src/Core/Components/Button/FluentToggleButton.razor.cs
vnbaaij: As discussed. It uses ElementInternals which is something different than custom events. No changes necessary atm

dvoituron (approved):

dvoituron (approved):

PR #3571, opened 2025-03-24
[AutoComplete] Enable "Multiple = false" when selecting a single item is desired

StevenRasmussen:
# Pull Request

## 📖 Description

When the `MaximumSelectedOptions == 1`, the control is rendered less than ideal due to the inherent nature that the control is expecting more than one item to be selected.  This PR seeks to improve both the UI and UX of the FluentAutoComplete when only 1 item is desired.

### 🎫 Issues

Here is how the control renders like when the `MaximumSelectedOptions == 1`

image

Here is how the control renders with this PR:

image

### General

### Component-specific

### DIFF
--- a/src/Core/Components/List/FluentAutocomplete.razor
+++ b/src/Core/Components/List/FluentAutocomplete.razor
@@ -115,11 +115,11 @@
 
             <FluentOverlay OnClose="@(e => IsMultiSelectOpened = false)" Visible="true" Transparent="true" FullScreen="true" />
             <FluentAnchoredRegion Anchor="@Id"
-                                    HorizontalDefaultPosition="HorizontalPosition.Right"
-                                    HorizontalInset="true"
-                                    VerticalDefaultPosition="@VerticalPosition.Unset"
-                                    Style="margin-top: 10px; border-radius: calc(var(--control-corner-radius) * 2px); background-color: var(--neutral-layer-floating);"
-                                    Shadow="ElevationShadow.Flyout">
+                                  HorizontalDefaultPosition="HorizontalPosition.Right"
+                                  HorizontalInset="true"
+                                  VerticalDefaultPosition="GetVerticalPosition()"
+                                  Style="margin-top: 10px; border-radius: calc(var(--control-corner-radius) * 2px); background-color: var(--neutral-layer-floating);"
+                                  Shadow="ElevationShadow.Flyout">
                 @if (HeaderContent != null)
                 {
                         @HeaderContent(Items ?? Array.Empty<TOption>())
@@ -158,11 +158,11 @@
         {
             <FluentOverlay OnClose="@(e => IsReachedMaxItems = false)" Visible="true" Transparent="true" FullScreen="true" />
             <FluentAnchoredRegion Anchor="@Id"
-                                    HorizontalDefaultPosition="HorizontalPosition.Right"
-                                    HorizontalInset="true"
-                                    VerticalDefaultPosition="@VerticalPosition.Unset"
-                                    Style="margin-top: 10px; border-radius: calc(var(--control-corner-radius) * 2px); background-color: var(--neutral-layer-floating); padding: 10px;"
-                                    Shadow="ElevationShadow.Flyout">
+                                  HorizontalDefaultPosition="HorizontalPosition.Right"
+                                  HorizontalInset="true"
+                                  VerticalDefaultPosition="GetVerticalPosition()"
+                                  Style="margin-top: 10px; border-radius: calc(var(--control-corner-radius) * 2px); background-color: var(--neutral-layer-floating); padding: 10px;"
+                                  Shadow="ElevationShadow.Flyout">
                 @MaximumSelectedOptionsMessage
             </FluentAnchoredRegion>
         }
--- a/src/Core/Components/List/FluentAutocomplete.razor.cs
+++ b/src/Core/Components/List/FluentAutocomplete.razor.cs
@@ -57,6 +57,12 @@ public partial class FluentAutocomplete<TOption> : ListComponentBase<TOption> wh
     public EventCallback<string> ValueTextChanged { get; set; }
 
     /// <summary>
+    /// Gets or sets the position of the options popup.
+    /// </summary>
+    [Parameter]
+    public SelectPosition? Position { get; set; }
+
+    /// <summary>
     /// Gets or sets the value of the input. This should be used with two-way binding.
     /// For the FluentAutocomplete component, use the <see cref="ValueText"/> property instead.
     /// </summary>
@@ -631,6 +637,18 @@ public partial class FluentAutocomplete<TOption> : ListComponentBase<TOption> wh
         }
 
     }
+
+    /// <summary>
+    /// Gets the position of the popup.
+    /// </summary>
+    /// <returns></returns>
+    private VerticalPosition? GetVerticalPosition()
+            => Position switch
+            {
+                SelectPosition.Above => VerticalPosition.Top,
+                SelectPosition.Below => VerticalPosition.Bottom,
+                _ => VerticalPosition.Unset,
+            };
 
     /// <summary />
     private bool MustBeClosed()
--- a/examples/Demo/Shared/Microsoft.FluentUI.AspNetCore.Components.xml
+++ b/examples/Demo/Shared/Microsoft.FluentUI.AspNetCore.Components.xml
@@ -5907,6 +5907,11 @@
             Gets or sets the callback that is invoked when the text field value changes.
             </summary>
         </member>
+        <member name="P:Microsoft.FluentUI.AspNetCore.Components.FluentAutocomplete`1.Position">
+            <summary>
+            Gets or sets the position of the options popup.
+            </summary>
+        </member>
         <member name="P:Microsoft.FluentUI.AspNetCore.Components.FluentAutocomplete`1.Value">
             <summary>
             Gets or sets the value of the input. This should be used with two-way binding.
@@ -6092,6 +6097,12 @@
         </member>
         <member name="M:Microsoft.FluentUI.AspNetCore.Components.FluentAutocomplete`1.RaiseValueTextChangedAsync(System.String)">
             <summary />
+        </member>
+        <member name="M:Microsoft.FluentUI.AspNetCore.Components.FluentAutocomplete`1.GetVerticalPosition">
+            <summary>
+            Gets the position of the popup.
+            </summary>
+            <returns></returns>
         </member>
         <member name="M:Microsoft.FluentUI.AspNetCore.Components.FluentAutocomplete`1.MustBeClosed">
             <summary />


MarvinKlein1508:
@StevenRasmussen wouldn't it make more sence to provide an itemprovider to the  `FluentCombobox` ?
https://www.fluentui-blazor.net/Combobox

But I agree the current behaviour for single items within `FluentAutocomplete` could need some love :)

StevenRasmussen:
@MarvinKlein1508 - That might work if you only have a small list of items.  However, when working with large datasets and you want to be able to search/filter items then the AutoComplete allows you to do this with the `OnOptionsSearch` callback.  I don't think the ComboBox provides this type of functionality OOB.

dvoituron:
Two important points:

1. You cannot modify the code and adapt unit tests on existing components. Unit tests check that the result has not changed. For example, you may have modified the CSS class `fluent-autocomplete-multiselect`, which will impact hundreds of thousands of developers who may have customized the style. You can add features, but you can't remove or modify them without creating a breaking change.

2. This visual change for `MaximumSelectedOptions=“1”` makes the code incompatible with **Accessibility**.
  Before  
  image
  After  
  image

tests/Core/List/FluentAutocompleteTests.FluentAutocomplete_Keyboard-ArrowDown.verified.razor.html
```diff
@@ -1,5 +1,5 @@
 
-<div class=" fluent-autocomplete-multiselect" style="width: 100%;" b-hg72r5b4ox="">
+<div class=" fluent-autocomplete-multiselect item-selected" style="width: 100%;" b-hg72r5b4ox="">
```
StevenRasmussen, line 2: @dvoituron - Is this an acceptable modification to the logic and thus the unit test?  A new class is being added if the count of the `SelectedOptions > 0`.  This shouldn't break anyone since the main class name is un-changed.

src/Core/Components/List/FluentAutocomplete.razor
```diff
@@ -204,7 +204,11 @@
                 {
                     var text = @GetOptionText(item);
 
-                    if (ReadOnly || Disabled)
+                    if (MaximumSelectedOptions == 1)
+                    {
+                        <FluentLabel aria-label="@GetOptionText(item)" role="checkbox" aria-checked="true">@text</FluentLabel>
```
StevenRasmussen, line 209: @dvoituron - I'm not sure if you think it appropriate to include a tab stop here or not. I left it out but could add it back in if you like.

src/Core/Components/List/FluentAutocomplete.razor
```diff
@@ -79,7 +79,7 @@
                                     Width="12px"
                                     Style="cursor: pointer;"
                                     Slot="end"
-                                    Title="@AccessibilityIconDismiss"
+                                    Title="@(MaximumSelectedOptions == 1 ? string.Format(AccessibilityRemoveItem, GetOptionText(SelectedOption)) : AccessibilityIconDismiss)"
```
StevenRasmussen, line 82: @dvoituron - When this is a single select (`MaximumSelectedOptions == 1`) then the accessibility text for the `Clear` button becomes the `AccessibilityRemoveItem` text which is something like `Remove {0}`.

tests/Core/List/FluentAutocompleteTests.FluentAutocomplete_Keyboard-ArrowDown.verified.razor.html
dvoituron: Nop. It's not acceptable to change all Unit Tests.
You can do the "inverse": you can add an attribute `single-selection` only for you feature. And adapt the CSS in this case= `.fluent-autocomplete-multiselect[single-selection] { ... }`

```xml
<div class="fluent-autocomplete-multiselect" single-selection> ... 
```

src/Core/Components/List/FluentAutocomplete.razor
dvoituron: Why did you change the current Title result: `DismissTitle="@(string.Format(AccessibilityRemoveItem, text))"` ?

dvoituron:
> 2. I'm not really all that familiar with how to make the component compatible with Accessibility.  Assuming I fix item 1 above, do you have any pointers and what needs to change to make it compatible?

Here are a few websites to explain Accessibility: 
- https://learn.microsoft.com/en-us/microsoft-edge/accessibility/design
- https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Accessibility/What_is_accessibility

The best solution is to use a tool such as [NVDA](https://nvaccess.org/download/) to validate the result.

src/Core/Components/List/FluentAutocomplete.razor
StevenRasmussen: This was an attempt to improve the "accessibility" of the component. Since you can no longer remove the individual item, I felt it might be good for the 'Clear' button to include the name of the selected option and so I just updated it to use the same text as if you were removing a single item.  I'm happy to change it back if you don't feel like this is a helpful update.

StevenRasmussen:
@dvoituron - Thanks for the pointers regarding accessibility.  I introduced changes that I believe have addressed your concern.  That being said, I feel a bit like I'm blindfolded and trying to hit a dart board... I'm not quite sure what an acceptable solution is for accessibility.  Please let me know if additional changes need to be made, otherwise I feel like the PR is ready for review again. Thanks!

StevenRasmussen:
@dvoituron - I wanted to check-in to see if you needed any additional changes to this PR. Thanks!

src/Core/Components/List/FluentAutocomplete.razor.cs
```diff
@@ -513,6 +513,34 @@ async Task KeyDown_EnterAsync()
         }
     }
 
+    /// <summary />
+    protected override void OnParametersSet()
+    {
+        base.OnParametersSet();
+
+        if (MaximumSelectedOptions == 1
+            && SelectedOptions?.Count() > 0
```
dvoituron, line 522: For better performances, you can use `SelectedOptions?.Any()` (and not `Count()`

src/Core/Components/List/FluentAutocomplete.razor
dvoituron: It's an option to add additional information in this case. But my biggest accessibility problem is that the component must “read” (with NVDA) all the elements present on the screen. This is still not the case when an item is selected, the user must be able to position himself on the component (focusable) and NVDA must announce something like “<SelectedItem> selected”.

To move forward with this PR (and to be compatible with what already exists), it would probably be simpler to allow the `Multiple=false` parameter, which would activate your functionality. If `Multiple=true`, everything would work as it does now, and Accessibility would be compatbile.

src/Core/Components/List/FluentAutocomplete.razor.cs
```diff
@@ -513,6 +513,34 @@ async Task KeyDown_EnterAsync()
...
+            && !(AdditionalAttributes?.TryGetValue("single-select", out var isSingleSelect) ?? false))
+        {
+            var additionalAttributes = (AdditionalAttributes is not null)
```
dvoituron, line 525: You don't need to update the `AdditionalAttributes`. You can add an attribute in the razor file like this:
```xml
<FluentTextField role="combobox" single-select="@GetSingleSelect()" ... >
```

```c#
private bool GetSingleSelect() => MaximumSelectedOptions == 1 && SelectedOptions?.Any() && Multiple == false;
```

dvoituron (changes requested):

src/Core/Components/List/FluentAutocomplete.razor.cs
StevenRasmussen: Done

src/Core/Components/List/FluentAutocomplete.razor.cs
StevenRasmussen: Done

src/Core/Components/List/FluentAutocomplete.razor
StevenRasmussen: @dvoituron - I believe adding the tab stop is what you are looking for as far as accessibility goes.  I have added it so that you can see what the result is. Let me know if you still want other changes or if you still think we should pursue allowing `Multiple` = false;

dvoituron: I'll check this out, but having a Multiple=false is still useful for people who want to keep the previous behavior.

Trapulo:
With this improvement (thanks! really important) Is there also a way to have a two-way simple binding to one single object it MaximumSelectedOptions  = 1 to perform a simple two-way binding to an object's property referring to the selected object?

dvoituron:
> With this improvement (thanks! really important) Is there also a way to have a two-way simple binding to one single object it MaximumSelectedOptions  = 1 to perform a simple two-way binding to an object's property referring to the selected object? 

No. Bu default this component is dedicated to bind multiple items. Including when you set `MaximumSelectedOptions=1`

src/Core/Components/List/FluentAutocomplete.razor
dvoituron: What do you think about this idea?

StevenRasmussen: I tried this by making the following changes:

image

However, the control did not seem to work.  It wouldn't allow me to select anything.  I'm not opposed to this idea, I'm just not sure what else would be required to get it to work.  At the end of the day, I'm simply looking for a way to only allow a single item to be selected and also have the UI styled in such a way that is optimal for a single item.  As it stands, I believe this PR accomplishes that and does it in a way that does not break anything existing and is hopefully acceptable from an accessibility perspective with the latest change I pushed.

StevenRasmussen:
@dvoituron @vnbaaij - Any other thoughts on these changes?  There's clearly interest from others in this type of an improvement just by the fact that 2 other people have commented on this PR.  As far as I know there are not any outstanding items for me to address. Please let me know if I missed something or there are other things you'd like me to look at. Thanks!

dvoituron:
> @dvoituron @vnbaaij - Any other thoughts on these changes? There's clearly interest from others in this type of an improvement just by the fact that 2 other people have commented on this PR. As far as I know there are not any outstanding items for me to address. Please let me know if I missed something or there are other things you'd like me to look at. Thanks!

Unless I'm mistaken, Accessibility has not been resolved. This can be resolved with this change. Is it already applied (comment is still Unresolved)
https://github.com/microsoft/fluentui-blazor/pull/3571#discussion_r2024269282

StevenRasmussen:
@dvoituron - I wasn’t certain if I should be the one resolving conversations or not. Anyway, regarding that particular point, I did mention in that thread that I attempted that solution but that the control did not even work after that. I didn’t investigate further after that. However, I believe that your concern over accessibility was addressed in this comment: [https://github.com/microsoft/fluentui-blazor/pull/3571#discussion_r2026200496](https://github.com/microsoft/fluentui-blazor/pull/3571#discussion_r2026200496)

dvoituron:
The Accessibility issue can be bypassed by setting the Multiple=false parameter. Autocomplete is currently used by many developers who need accessibility. And they don't want to change the way it works or the current design. 

By using Multiple=false, you don't have to change anything (including Accessibility). And by modifying this Multiple parameter, you can have your design too.

StevenRasmussen:
Sorry… I’m not quite understanding I think. As I mentioned I tried that and the control just stopped working. The dropdown wouldn’t even open. However, that was on an isolated branch from dev. Are you saying that I should make that change along with everything that I’ve already done here and that would work/suffice? I can try that to see what happens but wanted to make sure that I’m understanding correctly.

StevenRasmussen:
@dvoituron - Sorry, I misspoke.  The dropdown opens but the value does not persist.  Here is a video of how it behaves when you change the code like this:

image

IsMultiple

dvoituron:
My point is that we can't merge this PR at this moment. Because it has an impact on current applications (design and Accessibility). So we need a parameter to keep, by default, the current way of working (and I suggest using parameter Multiple)

StevenRasmussen:
@dvoituron - Alright!  I think I finally understand 😄.  I have made the appropriate updates to make `Multiple = false` work for the `FluentAutoComplete` and this is what triggers the changes in this PR to take effect.  This should not break any existing behavior for consumers.  A side benefit of this is that the underlying code in the `ListComponentBase` populates the `SelectedOption` instead of the `SelectedOptions` property which now allows binding to the `SelectedOption` when 'Multiple = false`.  In-fact, that is the only way now as the `SelectedOptions` property does NOT get populated when `Multiple = false`.  I think this is actually an improvement and what @Trapulo was asking for above.  I removed the unit test which was checking to make sure that an exception was thrown when `Multiple = false`... which for obvious reasons is no longer applicable.  Let me know if you need anything else here. Thanks!

src/Core/Components/List/FluentAutocomplete.razor
```diff
@@ -196,15 +197,25 @@
 
     private RenderFragment RenderSelectedOptions => __builder =>
     {
-        if (SelectedOptions != null)
+        var selectedOptions = new List<TOption>();
+        if(Multiple && SelectedOptions.Any())
```
dvoituron, line 201: `SelectedOptions` is nullable. You need to test it.
We prefer to use the `if` syntax using "curly braces" (one on each line).

dvoituron (commented):
This seems good. Small syntax change and I'll check the working component

src/Core/Components/List/FluentAutocomplete.razor
StevenRasmussen: Fixed

dvoituron (approved):

vnbaaij (approved):

dvoituron (approved):

StevenRasmussen:
@dvoituron, @vnbaaij - Thanks for your consideration on this.  I think the community is going to appreciate this enhancement 😄

Trapulo:
just implemented the new version: it solves a lot of problems, thanks

MarvinKlein1508:
@StevenRasmussen is there any easy way to allow the user to set the selection back to null/empty when he hits the delete/backspace key?

xebamart:
In version 4.11.9, when the FluentAutocomplete component sets the Multiple=false parameter, the following error occurs:

Unable to set property 'Multiple' on object of type 'Microsoft.FluentUI.AspNetCore.Components.FluentAutocomplete`1[[Reprehendo.Dtos.ScadaGroupDto, Reprehendo.Dtos, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]]'. The error was: For FluentAutocomplete, this property must be True. Set the MaximumSelectedOptions property to 1 to select just one item.

Code example:

              <FluentAutocomplete TOption="ScadaGroupDto"
                                    AutoComplete="off"
                                    Autofocus="true"
                                    Label="Select a country"
                                    Width="250px"
                                    Placeholder="Select a country"
                                    OnOptionsSearch="@OnSearchScadaGroup"
                                    Multiple="false"
                                    OptionText="@(item => item.Title)"
                                    @bind-SelectedOption=_selectedScadaGroup />

Is this already resolved?

vnbaaij:
> Well it doesn't work when I replicate it in my solution.

If you are 100% sure you are using the correct version, then please create a new issue (**with ready to run reproduction code**) for this.

MarvinKlein1508:
@xebamart can you please create a new issue and provide a ready to tun repository?

StevenRasmussen:
> @StevenRasmussen is there any easy way to allow the user to set the selection back to null/empty when he hits the delete/backspace key?

@MarvinKlein1508 - You can unselect an option by tabbing to the `X` and hitting the `Enter` key:

Autocomplete_Unselect_Option

It could be a good enhancement to add support for the `DELETE` or `BACKSPACE` keyboard keys too though.

xebamart:
In version 4.12.0 it works correctly.

PR #3590, opened 2025-03-28
[dev-v5] Extend Option with Description parameter

vnbaaij:
In the web component the Option component exposes a description attribute. This PR adds it to the Blazor component. Also,
it is possible to use the 'start' slot in an option. This PR shows examples on how to do that by adding a separate Option page. Lastly there were some tests added to ensure the new functionality works as expected.

image

- Extend Avatar with Slot parameter (+test)
- Extend Option with Description parameter (+test)
- Add Option with start slot and description example
- Add an Option page
- Add Option tests


examples/Demo/FluentUI.Demo.Client/Documentation/Components/List/Option/FluentOption.md
```diff
@@ -0,0 +1,19 @@
+---
+title: List / Option
+route: /List/Option
+---
+
+# Option
```
dvoituron, line 6: I don't think we need to make a page for `FluentOption`.
It's completely unnecessary, since it's necessarily usable in a List.
And it overloads the documentation.

We could add an example in the `FluentSelect.md` if you want.

tests/Core/Components/List/FluentOptionTests.razor
```diff
@@ -0,0 +1,63 @@
+﻿@using Microsoft.FluentUI.AspNetCore.Components.Utilities
+@using Xunit;
+@using Microsoft.FluentUI.AspNetCore.Components.Tests.Samples;
+@inherits TestContext
+@code
+{
+    private readonly IEnumerable<string?> Digits = new[] { null, "One", "Two", "Three" };
+    private enum MyDigitsEnum
```
dvoituron, line 8: Can you add a line space after all method? To have a clearer code to read.

dvoituron (changes requested):

examples/Demo/FluentUI.Demo.Client/Documentation/Components/List/Option/FluentOption.md
vnbaaij: 'Problem' is that it is not only used on select but also on combobox and others that will be introduced later. Maybe we need like a 'List' overview page?

tests/Core/Components/List/FluentOptionTests.razor
vnbaaij: Oops. Copilot generated text...

examples/Demo/FluentUI.Demo.Client/Documentation/Components/List/Option/FluentOption.md
vnbaaij: Added options content to Manuel part of select page

dvoituron: I'm still convinced that it would be better to put this doc in the FluentSelect.md page, but we'll see about that later :-)

dvoituron (approved):

PR #3605, opened 2025-04-01
Swallow the JSDisconnectedException when trying to call the 'dispose' Javascript method.

StevenRasmussen:
# Pull Request

## 📖 Description

The following error occurs when refreshing a page that has a `FluentInputFile` control:

image

We can't really do much about it other than swallow the error. See here: [https://stackoverflow.com/questions/72488563/blazor-server-side-application-throwing-system-invalidoperationexception-javas](https://stackoverflow.com/questions/72488563/blazor-server-side-application-throwing-system-invalidoperationexception-javas)

### General

### Component-specific

### DIFF
--- a/src/Core/Components/InputFile/FluentInputFile.razor.cs
+++ b/src/Core/Components/InputFile/FluentInputFile.razor.cs
@@ -429,7 +429,15 @@ public partial class FluentInputFile : FluentComponentBase, IAsyncDisposable
     {
         if (_containerInstance != null)
         {
-            await _containerInstance.InvokeVoidAsync("dispose");
+            try
+            {
+                await _containerInstance.InvokeVoidAsync("dispose");
+            }
+            catch (JSDisconnectedException)
+            {
+                // Swallow. See: https://stackoverflow.com/questions/72488563/blazor-server-side-application-throwing-system-invalidoperationexception-javas
+            }
+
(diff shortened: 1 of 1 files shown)


src/Core/Components/InputFile/FluentInputFile.razor.cs
vnbaaij, line 1: Good catch! Please make it follow what we have in other components already, like the `FluentDataGrid`:

```
 try
        {
            if (_jsEventDisposable is not null)
            {
                await _jsEventDisposable.InvokeVoidAsync("stop");
                await _jsEventDisposable.DisposeAsync().ConfigureAwait(false);
            }

            if (Module is not null)
            {
                await Module.DisposeAsync().ConfigureAwait(false);
            }
        }
        catch (Exception ex) when (ex is JSDisconnectedException ||
                                   ex is OperationCanceledException)
        {
            // The JSRuntime side may routinely be gone already if the reason we're disposing is that
            // the client disconnected. This is not an error.
        }
```
so, no linking to SO 😉

vnbaaij: I think this is precisely what I meant. Let me check on a desktop screen tonight before I approve.

vnbaaij (approved):

dvoituron (approved):

PR #3610, opened 2025-04-01
[dev-v5] Add SplitButton 

vnbaaij:
Adds the `FluentSplitButton` component. A SplitButton is a combination of a button and a menu. The left part of the button triggers the primary action. The right part (with the (customizable) chevron down icon) triggers a menu.

image


src/Core/Components/Button/FluentSplitButton.razor
```diff
@@ -0,0 +1,28 @@
...
+
+<FluentMenu Class="@ClassValue" Style="@StyleValue" OnClick="@OnMenuClick" OnCheckedChanged="@OnMenuCheckedChanged" split="true">
+    <FluentButton IconStart="@IconStart"
+                  IconOnly="@IconOnly"
+                  Appearance="@(Appearance == ButtonAppearance.Default ? null : Appearance)"
+                  Shape="@Shape"
+                  Size="@Size"
+                  BackgroundColor="@BackgroundColor"
+                  Color="@Color"
+                  OnClick="@OnClick"
```
dvoituron, line 14: You probably want to have `OnClick="@OnClickHandlerAsync"` or you can remove the `OnClickHandlerAsync` method?

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Buttons/SplitButton/Examples/SplitButtonDefault.razor
```diff
@@ -0,0 +1,60 @@
+﻿<FluentSplitButton OnClick="@ButtonClick"
+                   OnMenuClick="@MenuClick"
+                   Appearance="ButtonAppearance.Default"
+                   Title="Open menu"
+                   Label="Default">
+    @RenderMenuItems()
+</FluentSplitButton>
+
+<FluentSplitButton Appearance="ButtonAppearance.Primary"
```
dvoituron, line 9: Can you add `OnClick` and `OnMenuClick` on all other buttons?

src/Core/Components/Button/FluentSplitButton.razor.cs
```diff
@@ -0,0 +1,150 @@
...
+    /// Gets or sets a way to prevent further propagation of the current event in the capturing and bubbling phases.
+    /// </summary>
+    [Parameter]
+    public bool StopPropagation { get; set; } = false;
+
+    /// <summary>
+    /// Gets or sets the max height of the menu, e.g. 300px
+    /// </summary>
+    [Parameter]
+    public string? Height { get; set; }
```
dvoituron, line 114: This parameter is not used

src/Core/Components/Button/FluentSplitButton.razor.cs
```diff
@@ -0,0 +1,150 @@
...
+    ///  If both are specified, both will be rendered.
+    ///  </summary>
+    [Parameter]
+    public string? Label { get; set; }
+
+    /// <summary>
+    /// Gets or sets a way to prevent further propagation of the current event in the capturing and bubbling phases.
+    /// </summary>
+    [Parameter]
+    public bool StopPropagation { get; set; } = false;
```
dvoituron, line 108: This parameter is not used

src/Core/Components/Button/FluentSplitButton.razor.cs
```diff
@@ -0,0 +1,150 @@
...
+    [Parameter]
+    public string? Title { get; set; }
+
+    /// <summary>
+    ///  Gets or sets the content to be rendered inside the button.
+    ///  This can be used as an alternative to specifying the content as a child component of the button.
+    ///  If both are specified, both will be rendered.
+    ///  </summary>
+    [Parameter]
+    public string? Label { get; set; }
```
dvoituron, line 102: Should we have a `LabelTemplate` to help customization?

dvoituron (changes requested):
You need to write the Unit Tests
Current Unit Tests fail because some "default" properties are not defined.

src/Core/Components/Button/FluentSplitButton.razor
vnbaaij: OnClickHandlerAsync no needed anymore and removed

src/Core/Components/Button/FluentSplitButton.razor.cs
vnbaaij: No. In this case the Label is the actual text displayed on the button. We can't use ChildContent for this in this case as that holds the menu items

src/Core/Components/Button/FluentSplitButton.razor.cs
vnbaaij: Fixed

src/Core/Components/Button/FluentSplitButton.razor.cs
vnbaaij: It is only for the text displayed on the button. We could add it later but for now it is not needed I think. We do not have any example where the label on a button is using formattted (HTML) text

dvoituron (approved):

Issue #3618, opened 2025-04-02
[dev-v5] FluentSpacer


PascalVorwerk:
Migrating the v4 version to a v5 version :)

PR #3619, opened 2025-04-02
[dev-v5] FluentSpacer

PascalVorwerk:
# Pull Request

## 📖 Description

Migration, or pretty much a copy of, the fluent spacer component from v4 to v5. Result in example page:

image

### 🎫 Issues
#3618 

## 👩‍💻 Reviewer Notes

If we wan't some more flexible options for the span, let me know, i wasn't sure if i should take the freedom to add things or not, i felt like simply migrating was the best option. 

I've added some text to each example, but if you wan't me to change up the wording or anything let me know!

## 📑 Test Plan

Unit tests are added very every flow

### General


tests/Core/Components/Spacer/FluentSpacerTests.razor
vnbaaij, line 1: We don't always need to have verified files. In the last test for example, doing the Assert will be enough. The `cut.Verify()` can be removed. 
Something similar could be done for default test as well, I think.

src/Core/Components/Spacer/FluentSpacer.razor.cs
```diff
@@ -0,0 +1,20 @@
...
+/// <summary>
+/// Spacer component, used to create space between elements.
+/// </summary>
+public partial class FluentSpacer : ComponentBase
+{
+    /// <summary>
+    /// Gets or sets the width of the spacer (in pixels).
+    /// </summary>
+    [Parameter]
+    public int? Width { get; set; }
```
vnbaaij, line 18: This was the 'easy option' in v4. For v5, do we want to extend this to `string?` so other options like 'rem' or '%' become possible too?

vnbaaij (changes requested):

src/Core/Components/Spacer/FluentSpacer.razor.cs
dvoituron: Yes we need to change this to a `string`.

PascalVorwerk: Cool, i will change it. 

I was also wondering if we would want to support vertical spacing with the element? Maybe with a optional 'Height' parameter?

dvoituron: Personally, I find this useful, as long as you have an `Orientation` parameter (nullable and by default `Horizontal`). You should add in the documentation that this depends on the container style, which must be `display: flex` and which depends on `flex-direction`.

src/Core/Components/Spacer/FluentSpacer.razor.cs
```diff
@@ -0,0 +1,20 @@
...
+// ------------------------------------------------------------------------
+
+using Microsoft.AspNetCore.Components;
+
+namespace Microsoft.FluentUI.AspNetCore.Components;
+
+/// <summary>
+/// Spacer component, used to create space between elements.
+/// </summary>
+public partial class FluentSpacer : ComponentBase
```
dvoituron, line 12: All our FluentUI components must inherits from `FluentComponentBase` to include a list of default parameters (Data, Class, Style, ...)

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Spacer/Examples/SpacerWidth.razor
```diff
@@ -0,0 +1,14 @@
+﻿<div style="display: flex; width=100%">
+    <FluentIcon Value="@(new Icons.Filled.Size48.Alert())" Color="@Color.Neutral" />
+    <FluentSpacer Width="@Width" />
```
dvoituron, line 3: This is an example, so that will be easier for the dev to see `<FluentSpacer Width="50px" />` (and not adding a variable)

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Spacer/Examples/SpacerWidth.razor
```diff
@@ -0,0 +1,14 @@
+﻿<div style="display: flex; width=100%">
```
dvoituron, line 1: Can you use the FluentUI components :-) => `FluentStack`

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Spacer/Examples/SpacerDefault.razor
```diff
@@ -0,0 +1,9 @@
+﻿<div style="display: flex; width=100%">
```
dvoituron, line 1: Can you use the FluentUI components :-) => `FluentStack`

dvoituron: And the style `width=100%` is incorrect: (1) not an `=` symbol and missing `;` and the end.

dvoituron (changes requested):

src/Core/Components/Spacer/FluentSpacer.razor.cs
PascalVorwerk: I could finish this PR without the 'enhancement' , set up a new issue for it? That way we already have the migration ready, but we can still see if the enhancement can be included later?

src/Core/Components/Spacer/FluentSpacer.razor.cs
PascalVorwerk: Hmm okay, i can add this but this will be a lot more migration compared to the v4 version, as we will now expose many more attributes! Is that okay?

dvoituron: This is not a breaking change. The “migration” page will therefore contain only the list of new parameters. And I'm not even sure it's interesting to give this info :-)... but it's for completeness.

src/Core/Components/Spacer/FluentSpacer.razor.cs
dvoituron: It's not a big PR at the moment. If we decide to have this feature, it would be simpler to integrate it directly into this one. @vnbaaij Do you think it's unnecessary to have an `Orientation` and `Height` parameter?

src/Core/Components/Spacer/FluentSpacer.razor.cs
dvoituron: Yes, You can find an example here: https://github.com/microsoft/fluentui-blazor/blob/dev-v5/src/Core/Components/Divider/FluentDivider.razor

A global Unit Test will automatically validate if all of these parameters are used (`ComponentBaseTests.cs`).

```razor
id="@Id"
class="@ClassValue"
style="@StyleValue"
@attributes="AdditionalAttributes"
```

PascalVorwerk: Okay thanks!

src/Core/Components/Spacer/FluentSpacer.razor.cs
PascalVorwerk: I can't set the Orientation parameter as nullable and with a default of Horizontal.

dvoituron: ??? yes you can

```csharp
[Parameter]
public Orientation? Orientation { get; set; }
```
```csharp
var isHorizontal = Orientation is null || Orientation == Orientation.Horizontal;
```

PascalVorwerk: Ah i thought u meant as a default behind the parameter like so:

```csharp
[Parameter]
public Orientation? Orientation { get; set; } = Orientation.Horizontal
```

PascalVorwerk: I have a 'way' implemented now, but i can imagine some points of view that might be different. Let me know if you would prefer a different approach.

vnbaaij: Late to the party but adding vertical option now is fine by me too :)

dvoituron: I prefer to keep a `Width` parameter and to add a `Height` parameter (and not a common `Size`).
I think `Width`/`Height` is clearer and with the advantage to stay compatible with v4

PascalVorwerk: Okay my reasoning was:

- A size is clear, because you only want 1 'size' in 1 direction, if you have a width and a height, a person could think you could have spacer with both parameters set? Or is that what we want?

- If a `Height` is set, but the orientation is horizontal, do we just do nothing with the height?.

vnbaaij: Yes, Lets do Width and Height. Orientation dictates which one will be applied. Horizontal is default

PascalVorwerk: In the docs, i've added a example what happens if you set a parameter that doesn't work with the orientation. Do we want this?

image

image

PascalVorwerk: > No, think we can leave that out. We do want to have a test for it though!

Tests are made :), will remove the extra examples for it.

src/Core/Components/Spacer/FluentSpacer.razor
```diff
@@ -0,0 +1,11 @@
+﻿@namespace Microsoft.FluentUI.AspNetCore.Components
+@inherits FluentComponentBase
+
+<div style="@StyleValue"
+     class="@ClassValue"
+     id="@Id"
+     @attributes="AdditionalAttributes"/>
+
```
dvoituron, line 8: Can you remove these extra empty lines (not used)?

src/Core/Components/Spacer/FluentSpacer.razor.cs
```diff
@@ -0,0 +1,44 @@
...
+    public string? Width { get; set; }
+
+    /// <summary>
+    /// Specify the orientation of the parent container, which can be either horizontal or vertical.
+    /// </summary>
+    [Parameter]
+    public Orientation Orientation { get; set; } = Orientation.Horizontal;
+
+    /// <summary/>
+    public string? ClassValue => DefaultClassBuilder
```
dvoituron, line 33: Like in other components, can you move these 2 properties before the [Parameter] properties?

tests/Core/Components/Spacer/FluentSpacerTests.razor
```diff
@@ -0,0 +1,67 @@
...
+
+    [Fact]
+    public void FluentSpacer_Horizontal_Width_Pixels()
+    {
+        // Arrange && Act
+        var cut = Render(@<FluentSpacer Width="50px" />);
+
+        // Assert
+        var width = cut.Find("div").GetStyle()["width"];
+        Assert.Equal("50px", width);
```
dvoituron, line 32: You should also check that the `flex-grow` parameter is not set.

tests/Core/Components/Spacer/FluentSpacerTests.razor
```diff
@@ -0,0 +1,67 @@
...
+
+    [Fact]
+    public void FluentSpacer_Vertical_Height_Pixels()
+    {
+        // Arrange && Act
+        var cut = Render(@<FluentSpacer Height="50px" Orientation="Orientation.Vertical"/>);
+
+        // Assert
+        var width = cut.Find("div").GetStyle()["height"];
+        Assert.Equal("50px", width);
```
dvoituron, line 43: You should also check that the `flex-grow` parameter is not set.

tests/Core/Components/Spacer/FluentSpacerTests.razor
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    [Fact]
+    public void FluentSpacer_Vertical_Height_Pixels()
+    {
+        // Arrange && Act
+        var cut = Render(@<FluentSpacer Height="50px" Orientation="Orientation.Vertical"/>);
+
+        // Assert
+        var width = cut.Find("div").GetStyle()["height"];
```
dvoituron, line 42: The variable must be named `height`

tests/Core/Components/Spacer/FluentSpacerTests.razor
```diff
@@ -0,0 +1,67 @@
...
+
+    [Fact]
+    public void FluentSpacer_Vertical_WidthDoesNotSetStyle()
+    {
+        // Arrange && Act
+        var cut = Render(@<FluentSpacer Width="50px" Orientation="Orientation.Vertical"/>);
+
+        // Assert
+        var width = cut.Find("div").GetStyle()["width"];
+        Assert.Empty(width);
```
dvoituron, line 54: You should also check that the `flex-grow` parameter is not set.

tests/Core/Components/Spacer/FluentSpacerTests.razor
```diff
@@ -0,0 +1,67 @@
...
+
+    [Fact]
+    public void FluentSpacer_Horizontal_HeightDoesNotSetStyle()
+    {
+        // Arrange && Act
+        var cut = Render(@<FluentSpacer Height="50px" />);
+
+        // Assert
+        var width = cut.Find("div").GetStyle()["height"];
+        Assert.Empty(width);
```
dvoituron, line 65: You should also check that the `flex-grow` parameter is not set.

dvoituron (changes requested):

dvoituron:
Great job. Thanks

vnbaaij (approved):

dvoituron:
@PascalVorwerk Since we merged your PR, we've been getting these warnings. Could you create a new PR to update your examples to use the correct colors?

```
SpacerVerticalPixels.razor(2,68,2,81): warning CS0618: 'Color.Neutral' is obsolete: 'Use Default instead.'
SpacerVerticalPixels.razor(4,68,4,80): warning CS0618: 'Color.Accent' is obsolete: 'Use Primary instead.'
SpacerVertical.razor(2,68,2,81): warning CS0618: 'Color.Neutral' is obsolete: 'Use Default instead.'
SpacerVertical.razor(4,68,4,80): warning CS0618: 'Color.Accent' is obsolete: 'Use Primary instead.'
SpacerDefault.razor(2,68,2,81): warning CS0618: 'Color.Neutral' is obsolete: 'Use Default instead.'
SpacerDefault.razor(4,68,4,80): warning CS0618: 'Color.Accent' is obsolete: 'Use Primary instead.'
SpacerHorizontalPixels.razor(2,68,2,81): warning CS0618: 'Color.Neutral' is obsolete: 'Use Default instead.'
```

PascalVorwerk:
> @PascalVorwerk Since we merged your PR, we've been getting these warnings. Could you create a new PR to update your examples to use the correct colors?
> 
> ```
> SpacerVerticalPixels.razor(2,68,2,81): warning CS0618: 'Color.Neutral' is obsolete: 'Use Default instead.'
> SpacerVerticalPixels.razor(4,68,4,80): warning CS0618: 'Color.Accent' is obsolete: 'Use Primary instead.'
> SpacerVertical.razor(2,68,2,81): warning CS0618: 'Color.Neutral' is obsolete: 'Use Default instead.'
> SpacerVertical.razor(4,68,4,80): warning CS0618: 'Color.Accent' is obsolete: 'Use Primary instead.'
> SpacerDefault.razor(2,68,2,81): warning CS0618: 'Color.Neutral' is obsolete: 'Use Default instead.'
> SpacerDefault.razor(4,68,4,80): warning CS0618: 'Color.Accent' is obsolete: 'Use Primary instead.'
> SpacerHorizontalPixels.razor(2,68,2,81): warning CS0618: 'Color.Neutral' is obsolete: 'Use Default instead.'
> ```

I completely forgot about that, ofcourse, i will send a PR later today!

Issue #3620, opened 2025-04-03
fix: vertical scrolling in FluentDataGrid with lots of columns


vpetit-reimagine:
# 🐛 Bug Report
When using FluentDataGrid with a lot of columns (that take more horizontal space than the current screen resolution), it is impossible to scroll down vertically from the right of the screen.

## 💻 Repro or Code Sample
Please use the attached project as a showcase example.

1. Open the page and scroll horizontally to the right of the screen (first left column should not be visible anymore)
2. Vertically scroll (mouse/touchpad/scrollbar)

## 🤔 Expected Behavior
The table should display new rows as usual

## 😯 Current Behavior
The table does scroll but no new rows are displayed until you scroll back to the left of the page

## 🔦 Context
- We are using `Virtualize` and a lot of `@if (_columns.Contains("ColumnName")` to determine which columns to display (user selected)
- We have a table with ~4k rows and more than 20 columns
- They also contain other fluent elements (not needed to reproduce the issue)

## 🌍 Your Environment
* OS & Device: Windows and Linux
* Browser Microsoft Edge, Chrome, Firefox
* .NET and Fluent UI Blazor library Version: 8 and latest FluentUI version

[FluentUiDataGridDemo.zip](https://github.com/user-attachments/files/19581432/FluentUiDataGridDemo.zip)

vnbaaij:
Ok, I looked and see two issues with your code...

1) You **cannot** use both `Virtualize` **and** `MultiLine` at the same time. It is a requirement of the standard Blazor `Virtualize` component that each item (in this case a row) needs to have the same height. You can set that height with the `ItemSize` parameter (int for number of pixels, defaults to 32).
2) https://www.fluentui-blazor.net/datagrid#displaymode When using `Virtualize` (assuming point 1 above is resolved), we **highly** recommend using the Table display mode. To get the layout you have now, make sure to set the `Width` parameter on every column! 

After changing the DataGrid to:
```
  <FluentDataGrid TGridItem="MyItem" Items="@(_myItems.AsQueryable())"
                    Virtualize="true"
                    GenerateHeader="GenerateHeaderOption.Sticky"
                    DisplayMode="DataGridDisplayMode.Table">
:
```

scrolling works perfectly well. Only thing is it also scrols back horizontally but that is not something we can change at the moment. It is caused by Virtualize component.

![Image](https://github.com/user-attachments/assets/fc42c61e-ba2c-4b6b-8c36-07162343af4d)

[...]

Issue #3611, opened 2025-04-02
fix: Scroll on Virtualized data in FluentDataGrid falls in infinite rendering loop


svetovid:
# 🐛 Bug Report

Scroll on Virtualized data in FluentDataGrid falls in infinite rendering loop.

When I as a user scroll up and down aggressively in the FluentDataGrid control, rendering breaks down and looks like on the image (often flickering).
![Image](https://github.com/user-attachments/assets/41d1cbcc-a9bd-4f99-89c7-21b3ca721a2d)

Chrome inspect tool on "Elements" tab goes crazy.
![Image](https://github.com/user-attachments/assets/13ec63f0-0610-4289-83e2-3aaeaf73af64)

## 💻 Repro or Code Sample

```
<FluentDataGrid Items=@FilteredItems
                            Virtualize="true"
...
```
`FilteredItems` is a list property containing POCO items with 2 string properties
The size of the list is about 600 items

```
public record CultureDataItem(string Name, string DisplayName);

private string m_NameFilter = string.Empty;
    
public IQueryable<CultureDataItem> Content { get; set; } = CultureInfo.GetCultures(CultureTypes.SpecificCultures).Select(x => new CultureDataItem(x.Name, x.DisplayName)).AsQueryable();

private IQueryable<CultureDataItem>? FilteredItems
    => Content.Where(x => x.DisplayName.Contains(m_NameFilter, StringComparison.OrdinalIgnoreCase));
```

## 🤔 Expected Behavior

Data loads quicker and renders smoother

## 😯 Current Behavior

It gets stuck on rendering FluentDataGrid items while scrolling, when Virtualization is enabled 

## 🌍 Your Environment

An app running in WebView2 control
WebView2 127.0.2651.86
.NET 8.0.11
Microsoft.FluentUI.AspNetCore.Components 4.11.7

vnbaaij:
We can't (and won't) look at this without ready to run reproduction code. Standard reply follows...

Please supply us with ready-to-run reproduction code in the form of something we can copy/paste, a (zipped) project structure or a GitHub repository.

We do **not** have capacity to craft or compose a reproduction for every issue that gets raised. 

If no code or repository is provided, this issue will be closed in **3** days

Help us to help you. Thanks.

svetovid:
Hi @vnbaaij ,

Please, check this [project](https://github.com/svetovid/fluent-data-grid-issue)

100% reproducible while using mouse wheel for scrolling

Thanks in advance

[...]

PR #3621, opened 2025-04-03
[DataGrid] Add icons to column menus

vnbaaij:
[DataGrid] Add icon to column menu items, update docs to how how to use/not use.. Also add the label records to the doc page (from #3611)

[DataGrid] Runtime error when using Virtualize together with Multiple (from #3620)


src/Core/Components/DataGrid/Columns/ColumnBase.razor
```diff
@@ -52,15 +52,37 @@
                     {
                         <FluentMenuItem OnClick="@(async () => await Grid.SortByColumnAsync(this))" @onkeydown="HandleSortMenuKeyDownAsync">
                             @GetSortOptionText()
+                            @if (Grid.ColumnSortLabels.Icon is not null)
+                            {
+                                <span style="display:flex; align-items: center;" slot="@Grid.ColumnSortLabels.Slot">
+                                    <FluentIcon Value="@(Grid.ColumnSortLabels.Icon)" Slot="@Grid.ColumnSortLabels.Slot" />
```
dvoituron, line 58: Do you need to set the `slot` for the `span` and also for the `FluentIcon` ?
Same for other ResizeMenu and OptionsMenu below

src/Core/Components/DataGrid/Columns/ColumnOptionsLabels.cs
```diff
@@ -10,6 +10,17 @@ public record ColumnOptionsLabels
...
+    /// <summary>
+    /// Gets or sets the icon to show in the column menu
+    /// </summary>
+    public Icon? Icon { get; set; } = new CoreIcons.Regular.Size16.Filter();
+
+    /// <summary>
+    /// Gets or sets the position the icon can be showed in.
+    /// Possible options are 'start' (default) or 'end'
+    /// </summary>
+    public string Slot { get; set; } = "start";
```
dvoituron, line 22: Why do you allow other string values? Why not a bool `bool IconPositionStart`?

dvoituron (changes requested):

src/Core/Components/DataGrid/Columns/ColumnBase.razor
vnbaaij: Yes it is a web component shortcoming. See https://fluentui-blazor.net/Menu#withicons

src/Core/Components/DataGrid/Columns/ColumnOptionsLabels.cs
vnbaaij: Yeah, can change that.

Issue #3624, opened 2025-04-03
[dev-v5] FluentImage


PascalVorwerk:
Web component docs: https://web-components.fluentui.dev/?path=/docs/components-image--docs

PR #3630, opened 2025-04-04
[dev-v5] FluentImage

PascalVorwerk:
# Pull Request

## 📖 Description

I used the descriptions as best as possible from : https://web-components.fluentui.dev/?path=/docs/components-image--docs

### 🎫 Issues

#3624

## 👩‍💻 Reviewer Notes

I can image some different ways of implementing this, like for example exposing a source and alt text property and not accepting a render fragment, but then we don't allow multiple images? I wasn't exactly sure, so i wen't with the most straight forward approach. Obviously, if someone supplies something different then a image component, weird things happen.

Opening it in draft mode, as I can imagine some points of view on this.

Also, i was not sure what image to use for the examples and tests, so let me know if should use something different 😄 

### General

### Component-specific


dvoituron:
I think we should make this component more suitable for Blazor developers. 
For example, to offer (in addition to what's already been done) a `Source` (src), `Alternate` (alt), `Width` and `Height` property. This would avoid the need to use one or more `img` tags (although this would still be possible).

PascalVorwerk:
> I think we should make this component more suitable for Blazor developers. For example, to offer (in addition to what's already been done) a `Source` (src), `Alternate` (alt), `Width` and `Height` property. This would avoid the need to use one or more `img` tags (although this would still be possible).

I see your point but, how would we still support multiple 'different' images for example? Or do you wan't 1 FluentImage to be limited to 1 'src', alternate, width and height property?

dvoituron:
> I see your point but, how would we still support multiple 'different' images for example? Or do you wan't 1 FluentImage to be limited to 1 'src', alternate, width and height property?

I think 99% of cases will have 1 single `img` per `FluentImage`, but as said before, we must not remove the current possibility of putting “anything” in the `ChildContent`. If the dev wants to put a list of images there, no problem.

PascalVorwerk:
> > I see your point but, how would we still support multiple 'different' images for example? Or do you wan't 1 FluentImage to be limited to 1 'src', alternate, width and height property?
> 
> I think 99% of cases will have 1 single `img` per `FluentImage`, but as said before, we must not remove the current possibility of putting “anything” in the `ChildContent`. If the dev wants to put a list of images there, no problem.

Ok, so if the developer uses the source, Alternate, width and height properties and doesn't supply a child content, we fill the img tag ourselves, but if the user defines childcontent, we simply render their component? Or am i misunderstanding?

dvoituron:
> Ok, so if the developer uses the source, Alternate, width and height properties and doesn't supply a child content, we fill the img tag ourselves, but if the user defines childcontent, we simply render their component? Or am i misunderstanding?

Exact, but you can combine the both: if the user set a value for `Source` **and also** for `ChildContent`, you component will be:

```razor
<fluent-image id="@Id" ...>
    <img src="@Source" ... />
    @ChildContent
</fluent-image>
```

PascalVorwerk:
> > Ok, so if the developer uses the source, Alternate, width and height properties and doesn't supply a child content, we fill the img tag ourselves, but if the user defines childcontent, we simply render their component? Or am i misunderstanding?
> 
> Exact, but you can combine the both: if the user set a value for `Source` **and also** for `ChildContent`, you component will be:
> 
> ```
> <fluent-image id="@Id" ...>
>     
>     @ChildContent
> </fluent-image>
> ```

Okay cool! I understand! Will work on it further later today!

dvoituron:
> Okay cool! I understand! Will work on it further later today! However, this would mean that we always render that img component, even if the end user does not provide those properties, is that what we want? Maybe it is more clear with : Either use the properties or define your content?

Obviously, if the user doesn't provide a Source, the first `img` is not rendered. 

```razor
<fluent-image id="@Id" ...>
    @if (!string.IsNullOrEmpty(Source))
    {
        <img src="@Source" ... />
    }
    @ChildContent
</fluent-image>
```
In Blazor, you can't prevent a dev from filling in parameters under certain conditions.
And if he gives a value to `Source` and also to `ChildContent`, it's logical that both should be rendered.

PascalVorwerk:
@dvoituron I have added the paramaters you suggested. I still need to add some more tests for these new parameters, which i will pick up now.

I have also renamed the 'size' parameter, to me it felt confusing.. The documentation of the webcomponents is also not consisten, it calls it the 'Size' parameter, but the web component actually sets it as 'Fit': https://web-components.fluentui.dev/?path=/docs/components-image--docs

I felt like renaming it to 'Fit' was the best option, but keeping it Size felt conflicting especially with the Width and Height parameter.

PascalVorwerk:
Ok, i have added some more tests but i came to the conclusion that having the `Style` and `Class` attributes apply to the internal `img` actually breaks the base component tests, because it checks if the fluent component implements it, not the internal stuff.

I have added 2 extra properties that allows the developer to adjust the styling on the image. 

Its actually kinda weird, because the fluent-image component class and styling doesnt actually apply anything to the image styling.

For example, if setting a `Width` and `Height` styling on the `fluent-image` component itself will not actually change the width and height of the image:

image
image

The width and height parameter now actually set those properties for the internal `img`:
image
image

I'm not completely satisfied with this approach, what do you guys think? It's pretty confusing to have this outer styling and class component that does not really affect the styling on the part you want to style?

dvoituron:
I won't be adding the `ImageStyle` and `ImageClass` parameters. The best way is probably to set a `fluentui-image-item` class on the `img` tag. This makes it easier for the dev to customize this element if necessary.

PascalVorwerk:
> I won't be adding the `ImageStyle` and `ImageClass` parameters. The best way is probably to set a `fluentui-image-item` class on the `img` tag. This makes it easier for the dev to customize this element if necessary.

Ok, the `Width` and `Height` parameters still need to apply to the img tag, so we do keep a internal StyleBuilder for the image?

dvoituron:
Yes you can keep the ImageStyle since this is not a public property.

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Image/Examples/FluentImageBorder.razor
```diff
@@ -0,0 +1 @@
+﻿<FluentImage Bordered="true" Source="https://picsum.photos/200"/>
```
dvoituron, line 1: You can't see any difference with “normal” images. Can you use this to render “empty” and display the border?

```html
<FluentImage Bordered="true" Source="data:image/gif;base64,R0lGODlhAQABAAAAACH5BAEKAAEALAAAAAABAAEAAAICTAEAOw==" Width="100px" Height="100px" />
```

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Image/Examples/FluentImageCustomStyling.razor.css
```diff
@@ -0,0 +1,4 @@
+﻿::deep .fluent-image-item{
```
dvoituron, line 1: Can you use a style like this to make a border size more visible? `border: 3px dashed dodgerblue;`

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Image/Examples/FluentImageMultiple.razor
```diff
@@ -0,0 +1,4 @@
+﻿<FluentImage Shape="ImageShape.Circular">
+    <img style="border: 1px dashed dodgerblue" src="https://picsum.photos/200" alt="Placeholder Image"/>
```
dvoituron, line 2: Increase the border to 3px

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Image/FluentImage.md
```diff
@@ -0,0 +1,47 @@
...
+## Customize styling on the image
+FluentImage is a wrapper around an image tag if the source parameter is set. To customize the styling of the internal image,
+you can use the `fluent-image-item` class.
+{{FluentImageCustomStyling Files=Razor:FluentImageCustomStyling.razor;CSS:FluentImageCustomStyling.razor.css}}
+
+## Custom content
+You can provide any content inside the Fluent image as long as it is a image or a list of images. Any image inside the FluentImage will have the parameters of the component,
+except for the `Style`, `Class`, `Width` and `Height` parameters. This can, for example, be used with multiple images.
+
+{{FluentImageMultiple}}
```
dvoituron, line 47: You need to add a section API

src/Core/Components/Image/FluentImage.razor.cs
```diff
@@ -0,0 +1,93 @@
...
+        .Build();
+
+    /// <summary/>
+    private string? ImageStyleValue => new StyleBuilder()
+        .AddStyle("width", Width, () => !string.IsNullOrEmpty(Width))
+        .AddStyle("height", Height, () => !string.IsNullOrEmpty(Height))
+        .Build();
+
+    /// <summary>
+    /// The height of the image.
```
dvoituron, line 34: The height of the image, **applies only to the <see cref="Source"/> parameter.**

src/Core/Components/Image/FluentImage.razor.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// </summary>
+    [Parameter]
+    public string? Height { get; set; }
+
+    /// <summary>
+    /// The width of the image.
```
dvoituron, line 40: The width of the image, **applies only to the <see cref="Source"/> parameter.**

src/Core/Components/Image/FluentImage.razor.cs
```diff
@@ -0,0 +1,93 @@
...
+    public string? Width { get; set; }
+
+    /// <summary>
+    /// The source link for the image.
+    /// </summary>
+    [Parameter]
+    public string? Source { get; set; }
+
+    /// <summary>
+    /// The alternate text for the image.
```
dvoituron, line 52: The alternate text for the image, **applies only to the <see cref="Source"/> parameter.**

src/Core/Components/Image/FluentImage.razor.cs
```diff
@@ -0,0 +1,93 @@
...
+    public string? Height { get; set; }
+
+    /// <summary>
+    /// The width of the image.
+    /// </summary>
+    [Parameter]
+    public string? Width { get; set; }
+
+    /// <summary>
+    /// The source link for the image.
```
dvoituron, line 46: Start all property comments with `Gets or sets ...`

src/Core/Enums/ImageFit.cs
```diff
@@ -0,0 +1,31 @@
...
+
+/// <summary>
+/// Determines how the image will be scaled and positioned within its parent container.
+/// </summary>
+public enum ImageFit
+{
+    /// <summary>
+    /// The image will not be resized or repositioned.
+    /// </summary>
+    None,
```
dvoituron, line 15: You need to add an attribute `Description` for ALL enum values. And to use `ToAttributeValue` to get this value in your razor file.

src/Core/Enums/ImageShape.cs
```diff
@@ -0,0 +1,26 @@
...
+
+/// <summary>
+/// The image shape.
+/// </summary>
+public enum ImageShape
+{
+    /// <summary>
+    /// The image will be displayed as a circle.
+    /// </summary>
+    Circular,
```
dvoituron, line 15: Same: Add `Description`

dvoituron (changes requested):

PascalVorwerk:
@dvoituron Thanks for the review! I believe i have addressed all the points, if i missed some, sorry in advance.

PS. The page got alot more 'lovely' with that license free image ❤️

dvoituron (approved):

PR #3640, opened 2025-04-07
[dev-v5] FluentLayout - Refactoring

dvoituron:
# [dev-v5] FluentLayout - Refactoring

1. Update the `DemoMainLayout.razor` file to use the latest components

3. To make it easier to customize your sub-components, you can use these CSS variables, whose default values are as follows.

You can adapt them using the Height parameter of the FluentLayout, FluentLayoutItem with Area="LayoutArea.Header", Area="LayoutArea.Content" and Area="LayoutArea.Footer".

```css
--layout-height: 100vh;
--layout-header-height: 44px;
--layout-footer-height: 36px;
--layout-body-height: calc(var(--layout-height) - var(--layout-header-height) - var(--layout-footer-height));
```

## Unit Tests

Updated


MarvinKlein1508:
@dvoituron would it be a good idea to use `100dvh` instead of `100vh` ? This would also consider the height based on the controls of the webbrowser on mobile devices.

dvoituron:
> @dvoituron would it be a good idea to use `100dvh` instead of `100vh` ? This would also consider the height based on the controls of the webbrowser on mobile devices.

That's a good point, and it's why I chose the `100vh`: on mobile, the space is smaller and I don't think the user would like to have a footer when the screen is displayed. In most cases, the footer contains additional information that the mobile user can see by slightly scrolling the screen. If this is not the developer's wish, he can easily change this value.

{B03F30E9-0E68-461F-AB84-C35E6E1EB117}

This value of '100vh` is already used in the site https://fluentui-blazor-v5.azurewebsites.net/

vnbaaij (approved):

PR #3647, opened 2025-04-08
[dev-v5] Add Radio and RadioGroup 

vnbaaij:
This PR adds `FluentRadio` and `FluentRadioGroup` components. These are modeled after the standard Blazor `InputRadio` and `InputRadioGroup`

image

Fully tested:
image

In earlier versions, `FluentRadio` inherited from `FluentComponentBase`. To make it work with the `FluentField` we use now, this has been changed so the component now inherits from `FluentinputBase`. Also, the `ChildComponent` parameter is no longer supported. You must use `Label` or `LabelTemplate` to set the label.


dvoituron (changes requested):

dvoituron (changes requested):
Can you add a PR description and update the status to "Ready for review"?

src/Core/Components/Radio/FluentRadio.razor.cs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    [CascadingParameter]
+    private FluentRadioContext? CascadedContext { get; set; }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the element is checked.
+    /// </summary>
+    [Parameter]
+    public bool? Checked { get; set; }
```
dvoituron, line 35: I don't understand in which conditions this parameter will be used (outside of a Group). And in this case, we should create a bindable property. But for me, we could remove it and "force" the dev to use an Radio only inside a group.

src/Core/Components/Radio/FluentRadio.razor.cs
```diff
@@ -0,0 +1,81 @@
...
+public partial class FluentRadio<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] TValue> : FluentInputBase<TValue>, IFluentField
+{
+    bool _trueValueToggle;
+
+    internal FluentRadioContext? Context { get; private set; }
+
+    /// <summary />
+    public FluentRadio()
+    {
+        Id = Identifier.NewId();
```
dvoituron, line 25: Why do we force to have an ID? Not working without that?

dvoituron (changes requested):

src/Core/Components/Radio/FluentRadio.razor.cs
vnbaaij: I assumed it was needed for the FluentField. But it isn't, so I've removed the constructor code.

src/Core/Components/Radio/FluentRadio.razor.cs
vnbaaij: It is used to set the initial checked state of a radio item in a group. We do not want to make this bindable as the binding takes place on the group level (with the `Value` parameter)

dvoituron: Weird. And what if the first option `FluentOption.Checked=true` and the `FluentOptionGroup.Value="Second"` ?
Why not to use only the bindable FluentOptionGroupd value?

dvoituron (approved):

src/Core/Components/Radio/FluentRadio.razor.cs
```diff
@@ -0,0 +1,75 @@
...
+using Microsoft.FluentUI.AspNetCore.Components.Extensions;
+//using Microsoft.FluentUI.AspNetCore.Components.Utilities;
+
+namespace Microsoft.FluentUI.AspNetCore.Components;
+
+/// <summary>
+/// A Fluent Radio button component.
+/// </summary>
+/// <typeparam name="TValue">The type for the value of the radio button</typeparam>
+public partial class FluentRadio<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] TValue> : FluentInputBase<TValue>, IFluentField
```
dvoituron, line 16: I don't think you need to implement `IFluentField`. Because the dev could use all properties like `LabelPosition`, `Message`, `MessageIcon` ... and without effect :-(

dvoituron (changes requested):

src/Core/Components/Radio/FluentRadio.razor
```diff
@@ -0,0 +1,28 @@
...
+             Class="@ClassValue"
+             Style="@StyleValue"
+             Label="@Label"
+             LabelPosition="@Components.LabelPosition.After"
+             LabelWidth="@LabelWidth"
+             IncludeInputSlot="false">
+    <LabelTemplate>@LabelTemplate</LabelTemplate>
+    <ChildContent>
+        <fluent-radio id="@Id"
+                      checked="@(Context.CurrentValue?.Equals(Value) == true ? GetToggledTrueValue() : null)"
```
dvoituron, line 19: As I understand the web component, the `RadioGroup.Value` is used to check the right Radio. This property would only be used if a `FluentRadio` was used outside a `Group`, which is not possible with us.

src/Core/Components/Radio/FluentRadio.razor
```diff
@@ -0,0 +1,28 @@
...
+                      disabled="@Disabled"
+                      slot="@FluentSlot.FieldInput"
+                      value="@(Value?.ToString() ?? Label ?? null)"
+                      name="@Context.GroupName"
+                      @onchange="@Context.ChangeEventCallback"
+                      @attributes="AdditionalAttributes">
```
dvoituron, line 25: Can you add the `@` to be clearer? `@AdditionalAttributes`

src/Core/Components/Radio/FluentRadio.razor
```diff
@@ -0,0 +1,28 @@
...
+             IncludeInputSlot="false">
+    <LabelTemplate>@LabelTemplate</LabelTemplate>
+    <ChildContent>
+        <fluent-radio id="@Id"
+                      checked="@(Context.CurrentValue?.Equals(Value) == true ? GetToggledTrueValue() : null)"
+                      disabled="@Disabled"
+                      slot="@FluentSlot.FieldInput"
+                      value="@(Value?.ToString() ?? Label ?? null)"
+                      name="@Context.GroupName"
+                      @onchange="@Context.ChangeEventCallback"
```
dvoituron, line 24: You can catch the value changed from the `FluentRadioGroup`

src/Core/Components/Base/IFluentRadioValueProvider.cs
```diff
@@ -0,0 +1,10 @@
+// ------------------------------------------------------------------------
+// MIT License - Copyright (c) Microsoft Corporation. All rights reserved.
+// ------------------------------------------------------------------------
+
+namespace Microsoft.FluentUI.AspNetCore.Components;
+
+internal interface IFluentRadioValueProvider
```
dvoituron, line 7: This interface is only used with the FluentRadio elements. Probably better to move this file in the `Radio` folder

src/Core/Extensions/FluentInputExtensions.cs
```diff
@@ -2,9 +2,9 @@
 // MIT License - Copyright (c) Microsoft Corporation. All rights reserved.
 // ------------------------------------------------------------------------
 
-using Microsoft.AspNetCore.Components;
```
dvoituron, line 5: Why did you changed that?

src/Core/Components/Radio/FluentRadio.razor
```diff
@@ -0,0 +1,28 @@
...
+             LabelPosition="@Components.LabelPosition.After"
+             LabelWidth="@LabelWidth"
+             IncludeInputSlot="false">
+    <LabelTemplate>@LabelTemplate</LabelTemplate>
+    <ChildContent>
+        <fluent-radio id="@Id"
+                      checked="@(Context.CurrentValue?.Equals(Value) == true ? GetToggledTrueValue() : null)"
+                      disabled="@Disabled"
+                      slot="@FluentSlot.FieldInput"
+                      value="@(Value?.ToString() ?? Label ?? null)"
```
dvoituron, line 22: If the value is a "complex" object, this result will be a very long string of this object :-(

dvoituron (changes requested):

src/Core/Components/Radio/FluentRadio.razor
vnbaaij: As discussed, will be addressed in a future version

dvoituron (approved):

PR #3685, opened 2025-04-15
[dev-v5] Add the FluentMultiSplitter

dvoituron:
# [dev-v5] Add the FluentMultiSplitter

The **MultiSplitter** splits the page into multiple panels and allows the user to control the page layout.

You can include as many **Panes** as you like in a container.
By default, the panes are resizable, but the `Resizable` parameter can be used to block this functionality.
The initial `Size` is in percent or pixels and has a `Min` and `Max` value.  
A Panel can be fully collapsed by setting the `Collapsible` parameter to `true`.

The `OnResize` event returns the new size as a percentage.

peek_2
Note: the cursor was not correctly recorded in this video.

## Restrictions

This component must be able to communicate between the DOM (browser) and the .NET runtime.
Consequently, this component **requires an Interactive mode** (interactive SSR, interactive WASM).
Consequently, it does not function correctly in static rendering mode.

By design, the pane can only be resized and collapsed/expanded when `Resizable="true"`.

## Accessibility

This component is not accessible yet.

## Unit Tests

{CCF89D43-A452-499D-AB1E-E67B9F725DF5}


examples/Demo/FluentUI.Demo.Client/Documentation/Components/Splitter/MultiSplitter.md
```diff
@@ -0,0 +1,57 @@
...
+## Example
+
+In this example, each time a pane is resized, a trace is written to the console.
+
+{{ MultiSplitterDefault }}
+
+## Restrictions
+
+This component must be able to communicate between the DOM (browser) and the .NET runtime.
+Consequently, this component **requires an Interactive mode** (interactive SSR, interactive WASM).
```
vnbaaij, line 26: Please rephrase and use the same terms Blazor uses: Interactive Server, Interactive WebAssembly or Auto. Calling it SSR might cause confusion.

src/Core.Scripts/src/ExportedMethods.ts
```diff
@@ -2,6 +2,7 @@ import { Microsoft as LoggerFile } from './Utilities/Logger';
 import { Microsoft as AttributesFile } from './Utilities/Attributes';
 import { Microsoft as FluentDialogFile } from './Components/Dialog/FluentDialog';
 import { Microsoft as FluentTabsFile } from './Components/Tabs/FluentTabs';
+import { Microsoft as FluentMultiSplitterFile } from './Components/Splitter/FluentMultiSplitter';
```
vnbaaij, line 5: Why do we use the `File`-suffix here? Seems off. Maybe use `Script`-suffix instead?

src/Core/Components/Splitter/FluentMultiSplitter.razor.cs
```diff
@@ -0,0 +1,334 @@
...
+    public string? Width { get; set; }
+
+    /// <summary>
+    /// Gets or sets the height of the container.
+    /// </summary>
+    [Parameter]
+    public string? Height { get; set; }
+
+    /// <summary />
+    protected string? ClassValue =>DefaultClassBuilder
```
vnbaaij, line 88: Space missing

vnbaaij (changes requested):

src/Core.Scripts/src/ExportedMethods.ts
dvoituron: This is only internal to this file. And I need to use a suffix to distinct the "file" and the "import". We can change this suffix in another PR.

vnbaaij (approved):

PR #3712, opened 2025-04-18
[dev-v5] MessageBar (WIP!!)

vnbaaij:
Just storing it here to not loose anything. NOT DONE YET!

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ When using [Visual Studio](https://visualstudio.microsoft.com/), you can also us
 The templates will be available under the **Blazor** category.
 
 ### - Manual Install
-To start using the **Fluent UI Blazor components** from scratch, you first need to install the main [Nuget package](https://www.nuget.org/packages/Microsoft.FluentUI.AspNetCore.Components/) in the project you want to use the library and its components.
+To start using the **Fluent UI Blazor components** from scratch, you first need to install the main [NuGet package](https://www.nuget.org/packages/Microsoft.FluentUI.AspNetCore.Components/) in the project you want to use the library and its components.
 You can use the NuGet package manager in your IDE or use the following command when using a CLI:
 
 ```shell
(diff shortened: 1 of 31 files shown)


MarvinKlein1508:
@vnbaaij since you are working on this component I would like to mention one behaviour from v4 which really annoys my. By default all MessageBoxes are using fade animations. But for the most part I use them as warning or information boxes which will be initiated on page load. I know I can change this behaviour by setting `IsFade` to `false` but this needs to be done for every component I place in my project. I would like to have some way to change the default behaviour for this if possible.

Maybe you can consider it for v5 :)

vnbaaij:
> @vnbaaij since you are working on this component I would like to mention one behaviour from v4 which really annoys my. By default all MessageBoxes are using fade animations. But for the most part I use them as warning or information boxes which will be initiated on page load. I know I can change this behaviour by setting `FadeIn` to `false` but this needs to be done for every component I place in my project. I would like to have some way to change the default behaviour for this if possible.
> 
> Maybe also for `AllowDismiss`
> 
> Maybe you can consider it for v5 :)

Yes, want have these available on a provider level, I think.

But don't expect any new code/progress for the next 7 days. Enjoying some time off without access to an IDE 😉

MarvinKlein1508:
@vnbaaij sure, no problem! :) I just thought I might take the chance to mention this. This kind of functionality would be nice for other components as well. Perhaps there is some room for future ideas. 

I'm thinking about `FluentStack` for example where I always change the direction.

But first enjoy your days off! :)

vnbaaij:
> @vnbaaij sure, no problem! :) I just thought I might take the chance to mention this. This kind of functionality would be nice for other components as well. Perhaps there is some room for future ideas.

Definitely! But in the short term probably only for our own 'native' Blazor components. For the web components our influence is more limited. 

> 
> I'm thinking about `FluentStack` for example where I always change the direction.

We must choose either orientation as a default. What is your idea to solve your issue? Create something like a `FluentVerticalStack` perhaps?

> 
> But first enjoy your days off! :) 

Already enjoying it. I'm answering this poolside from a sunbed, so... 😁

dvoituron:
> I'm thinking about `FluentStack` for example where I always change the direction.
What do you mean? By default, the `FluentStack` has an `Orientation=Horizontal` because the default HTML orientation is vertical.

In summary:
- Need a Horizontal orientation => `FluentStack`
- Need a Vertical orientation => `div`

So. Why do you want to be able to change the default `Orientation`?

MarvinKlein1508:
> So. Why do you want to be able to change the default `Orientation`?

Because I use the `FluentStack` to have equal spacing between each item. Since the component allows you to change orientation, why not use it? :) 

I kinda use it like the Bootstrap col system.

dvoituron:
> Because I use the `FluentStack` to have equal spacing between each item. Since the component allows you to change orientation, why not use it? :)

Yes you can use it, and I also use it. But in this case, I set `Orientation="Vertical"` and all is fine :-)

PR #3714, opened 2025-04-19
[dev-v5] Add DateTimeProvider and DateTimeExtensions

dvoituron:
# [dev-v5] Add DateTimeProvider and DateTimeExtensions

1. Add a **DateTimeProvider** and **DateTimeProviderContext**
   See the **ReadMe.md** documentation in the **Utilities/DateTime** folder.

   > To simplify the possibility to unit test a current Date Time,
   > we propose a solution that uses the **Ambient Context Model**.
   >
   > To do this, use a `DateTimeProvider` class that determines the current date:
   > the default `DateTime.Now` is replaced by `DateTimeProvider.Now` in the code.
   >
   > Example:
   > ```csharp
   > int GetQuarter()
   > {
   >     return (DateTimeProvider.Today.Month - 1) / 3 + 1;
   > }
   > ```
   >
   > This **provider** returns the system's current date. However, by using it in a unit test, 
   > we can adapt the context to specify a predefined (fake / fixed) date.
   > 
   > ```csharp
   > var fakeDate = new DateTime(2018, 5, 26);
   > using (var context = new DateTimeProviderContext(fakeDate))
   > {
   >     var quarter = GetQuarter();  // 2
   > }

2. Add some `DateTimeExtensions` methods
3. Add a method `ToTimeAgo(TimeSpan)` to return an equivalent localizable string.

## Unit Tests

image
image

### DIFF
--- a/src/Core/Localization/ReadMe.md
+++ b/src/Core/Localization/ReadMe.md
@@ -1,0 +1,38 @@
+# Localization
+
+Localization allows the text in some components to be translated.
+
+## Explanation
+
+**Fluent UI Blazor** itself provides English language strings for texts found in.
+To customize translations in **Fluent UI Blazor**, the developer can register a
+custom `IFluentLocalizer` implementation as a service, and register this custom localizer
+in the `Program.cs` file, during the service registration.
+
+```csharp
+builder.Services.AddFluentUIComponents(config => config.Localizer = new CustomFluentLocalizer());
+```
+
+## How to add a new resource string?
+
+In this library, to add a new resource string, follow these steps:
+
+1. Open the file **LanguageResource.resx** and add a new entry with the name `<prefix>_<item>`.
+   Where `<prefix>` is the name of the component, and `<item>` is the name of the string.
+   Example:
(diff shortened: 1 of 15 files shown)


src/Core/Localization/ReadMe.md
```diff
@@ -0,0 +1,38 @@
+# Localization
+
+Localization allows the text in some components to be translated.
+
+## Explanation
+
+**Fluent UI Blazor** itself provides English language strings for texts found in.
```
vnbaaij, line 7: found in its components.

vnbaaij (changes requested):
Should we not also have these extensions and helpers for DateTimeOffset?

dvoituron:
Good question. This is the v4 extension file. And at this moment the v4 don't use DateTimeOffset extensions. 

So I propose to add these extensions later, only if we need it.

src/Core/Localization/ReadMe.md
dvoituron: Updated

vnbaaij (approved):

Issue #3734, opened 2025-04-29
fix: [FluentCombobox uses Text instead of Value] in [FluentCombobox]


Programmierhans:
# 🐛 Bug Report

Having a List with value and text, the FluentCombox uses the text-column as value and text.

## 💻 Repro or Code Sample

Having a table like:
Value Text
"1" "One"
"2" "Two"

The FluentCombobox displays the Text (One) but internally also uses "One" as value instead of using the value "1".

`@page "/BugComboBox"
@rendermode InteractiveServer
<FluentCombobox Items=@stringOptions
                TOption="Option<string>"
                OptionText="@(i => i.Text)"
                OptionValue="@(i => i.Value)"
                @bind-Value="@stringValue" />
<p>
    Selected Value: @stringValue<br />
</p>

@code {
    string? stringValue;

    List<Option<string>> stringOptions = new()
        {
            { new Option<string> { Value = "1", Text = "One" } },
            { new Option<string> { Value = "2", Text = "Two"} },
            { new Option<string> { Value = "3", Text = "Three" } }

        };
}
`

## 🤔 Expected Behavior

The FluentCombobox should use the value-member and not the text-member to store in the bound property

..> the Selected Value should show 1/2/3 and not One/Two/Three when the user selects an entry from the ComboBox

## 😯 Current Behavior

The FluentCombobox uses a wrong sequence to read the value.
It takes the Text and if not available it falls down to the Value and if not available it executes the ToString.

`    protected override string? GetOptionValue(TOption? item)
    {
        if (item != null)
        {
            return OptionText.Invoke(item) ?? OptionValue.Invoke(item) ?? item.ToString();
        }
        else
        {
            return null;
        }
    }`

## 💁 Possible Solution

by removing the GetOptionValue overload in the FluentCombobox, the code from the ListComponentBase would be invoked. And in the ListComponentBase the sequence is what i would expect.
It takes the Value  and if not available it falls back to Text and then to the ToString

`    protected virtual string? GetOptionValue(TOption? item)
    {
        if (item != null)
        {
            return OptionValue?.Invoke(item) ?? OptionText?.Invoke(item) ?? item?.ToString();
        }
        else
        {
            return null;
        }
    }`

## 🔦 Context

I detected this because the FluentCombobox always writes the visible value back to the datastore instead of the value behind the text.

## 🌍 Your Environment

I used the latest Fluent 4.11.8

vnbaaij:
[...]

PR #3739, opened 2025-04-30
Fix combobox GetOptionValue

dvoituron:
# Fix combobox GetOptionValue

The correct `GetOptionValue` is already included in the `ListComponentBase` class.

```csharp
// From ListComponentBase

protected virtual string? GetOptionValue(TOption? item)
{
    if (item != null)
    {
        return OptionValue?.Invoke(item) ?? OptionText?.Invoke(item) ?? item?.ToString();
    }
    else
    {
        return null;
    }
}
```

#3734

### DIFF
--- a/src/Core/Components/List/FluentCombobox.razor
+++ b/src/Core/Components/List/FluentCombobox.razor
@@ -14,7 +14,7 @@
                      style="@StyleValue"
                      autocomplete="@Autocomplete.ToAttributeValue()"
                      open=@Open
-                     current-value=@Value
+                     current-value=@GetComboboxContent()
                      placeholder=@Placeholder
                      position=@Position.ToAttributeValue()
                      disabled=@Disabled
--- a/src/Core/Components/List/FluentCombobox.razor.cs
+++ b/src/Core/Components/List/FluentCombobox.razor.cs
@@ -163,6 +163,18 @@ public partial class FluentCombobox<TOption> : ListComponentBase<TOption>, IAsyn
         }
     }
 
+    private string? GetComboboxContent()
+    {
+        if (SelectedOption != null)
+        {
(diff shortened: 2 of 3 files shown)


vnbaaij:
Are we sure this returns the correct value for a Combobox in the case of editing in the text box part (ie when not selecting an item from the list of options). I recall we had issues there before.

dvoituron:
> Are we sure this returns the correct value for a Combobox in the case of editing in the text box part (ie when not selecting an item from the list of options). I recall we had issues there before.

Good question. I will check this case.

dvoituron:
> Are we sure this returns the correct value for a Combobox in the case of editing in the text box part (ie when not selecting an item from the list of options). I recall we had issues there before.

@vnbaaij  You're right. I updated the code to display the `Text` while assigning the correct `Value`.

## before
Before

## after
After

I used this code to validate:
```razor
<FluentCombobox Items=@stringOptions
                OptionText="@(i => i.Text)"
                OptionValue="@(i => i.Value.ToString())"
                @bind-Value="@stringValue"
                @bind-SelectedOption="selectedOption" />

<div>
    Selected Value: @stringValue
</div>
<div>
    Selected Option Text: @selectedOption?.Text
</div>
<div>
    Selected Option Value: @selectedOption?.Value
</div>

@code {
    string? stringValue;
    Option<int>? selectedOption;

    List<Option<int>> stringOptions = new()
    {
        { new Option<int> { Value = 1, Text = "One" } },
        { new Option<int> { Value = 2, Text = "Two"} },
        { new Option<int> { Value = 3, Text = "Three" } }
    };
}
```

Can you double check before to merge?

vnbaaij (approved):
Confirmed it works as supposed

PR #3767, opened 2025-05-09
[DataGrid] Resize enhancements

vnbaaij:
Previously, when a DataGrid has `ResizeColumns="true"`, you could only resize a column by grabbing the right edge of the column header:

DataGrid-resize-old
>[!NOTE] the screen capture does not show the resize handle, but it is there.

With this PR we are extending the resizing to **all** rows and on top of that, we are showing an indicator across the whole column to make it clearer how the resizing will affect the column. 

DataGrid-resize-new

> [!NOTE] it is very well possible that when moving the mouse too far or when overlapping another column, the resize handle will disappear. That is just a limitation that we do not have a solution for currently.


adamint:
This looks great to me. Does the color of the bar meets contrast requirements?

MarvinKlein1508:
Just because I'd like to know. Is there any way to save the sizing somewhere e.g. in the localStorage?

vnbaaij:
> Just because I'd like to know. Is there any way to save the sizing somewhere e.g. in the localStorage? 

The original column values are saved in the script on initialization of the page. You can reset to those by using the reset button in the resizing option dialogs (discrete/exact resizing) or by using the shift+R key combination.

vnbaaij:
> This looks great to me. Does the color of the bar meet contrast requirements? 

I'll check. It is using a token value, so it should be ok. I'll choose another one if not.

PR #3770, opened 2025-05-09
[dev-v5] Add markdown shortcuts [!NOTE], [!TIP] and [!WARNING]

dvoituron:
# [dev-v5] Add markdown shortcuts [!NOTE], [!TIP] and [!WARNING]

Now we can use these shortcuts to be coherent with the Microsoft learn pages (see an example [here](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/program-structure/main-command-line).

```
> [!NOTE]
> My note

> [!TIP]
> My tip

>[!WARNING]
> My warning
```

image


vnbaaij (approved):

MarvinKlein1508:
I've added two new warnings to `InputFile` those needs to be updated to make use of this :)

PR #3870, opened 2025-06-04
[MultiSplitter] Remove position from panel CSS

### LINKED ISSUE #3868: FluentCombobox inside short FluentMultiSplitterPane clipped out
FluentMultiSplitter layout, I have a FluentCombobox in the top Toolbar component that is clipped out when open (see top right area in this screenshot):

Image

The pop-up combo list pane should be on top of everything...

This is not a grid (I read related posts), so it is legitimate to use this component here.

vnbaaij:
Just a small and simple CSS change to fix #3868. 

As discussed in the issue, a change is needed on the code using MultiSplitter as well.

No Unit Tests affected


src/Core/Microsoft.FluentUI.AspNetCore.Components.csproj
```diff
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk.Razor">
   <PropertyGroup>
-    <TargetFrameworks>net8.0;net9.0;net10.0</TargetFrameworks>
```
dvoituron: Why did you removed the NET10?

vnbaaij:
Oops

src/Core/Microsoft.FluentUI.AspNetCore.Components.csproj
vnbaaij: Added it back

PR #3882, opened 2025-06-05
[AutoComplete] Add OptionComparer to prevent duplicate items from different object instances (#3881)

### LINKED ISSUE #3881: fix: "duplicate" items in FluentAutoComplete
# 🐛 Bug Report

`FluentAutoComplete` adds "the same" item twice when selected. Although the dropdown menu shows that the item has already been added. This happens when you create a new instance of the search results with the same data.

In my real app this is happening because I fetch new rows from the database everytime with Dapper.

## 💻 Repro or Code Sample

For IssueTester:

```razor
<FluentAutocomplete TOption="MyUser"
                    Label="Users"
                    Class="w-100"
                    Placeholder="Name"
                    OnOptionsSearch="@OnSearchUserAsync"
                    OptionText="@(item => $"{item.Username}" )"
                    @bind-SelectedOptions="Input.Users" />

@code {
    public MyInput Input { get; set; } = new();   
    public class MyInput
    {
        public IEnumerable<MyUser> Users { get; set; } = [];
    }

    public class MyUser
    {
        public int UserId { get; set; }
        public string Username { get; set; }
    }

    private async Task OnSearchUserAsync(OptionsSearchEventArgs<MyUser> e)
    {
        var users = new List<MyUser>
        {
            new MyUser { UserId = 1, Username = "Alice Wonder" },
            new MyUser { UserId = 2, Username = "Marvin Klein" },
            new MyUser { UserId = 3, Username = "Vincent Baaji" },
        };

        e.Items = users;
    }
}
```

## 🤔 Expected Behavior

I expect the control to remove the first entry because it is already been added. Somehow the component already knows that this item is part of the component because the drop down menu shows it as already added.

## 😯 Current Behavior

You can add the same item unlimited times.

## 💁 Possible Solution

Maybe use the `SelectedOption` EventCallback to determine if an item is already part of the list and prevent the add then.

## 🌍 Your Environment

* Browser Mozilla Firefox latest version
* .NET and Fluent UI Blazor library Version .NET 9.0.5 and 4.11.9

MarvinKlein1508:
# Pull Request

## 📖 Description

This PR introduces a new parameter `OptionComparer` to `ListComponentBase` which can be used to check an item before it's being added to a `FluentAutocomplete` component. This makes it much easier to use this component when you are fetching new object instances from external APIs or from a database without Entity Framework.

Demo and Unit-test have also been added.

Example:
autocomplete_comparer_example

### 🎫 Issues

Fix #3881

### General

### Component-specific


vnbaaij:
I'll try to check and get this in ASAP. Was about to release 4.12 but think we can squize this in.

MarvinKlein1508:
You won't believe how happy this would make me!

src/Core/Components/List/ListComponentBase.razor.cs
```diff
@@ -128,6 +128,12 @@ protected string? InternalValue
     /// </summary>
     [Parameter]
     public virtual Func<TOption, bool>? OptionSelected { get; set; }
+    /// <summary>
+    /// Gets or sets the function used to determine if an option is already added to the internal list.
+    /// The first parameter is the item from the internal list to check and the second parameter is the actual selected item from the menu.
+    /// </summary>
+    [Parameter]
+    public virtual Func<TOption, TOption, bool>? OptionComparer { get; set; }
```
dvoituron: Have you tried using [IComparable](https://learn.microsoft.com/en-us/dotnet/api/system.icomparable)?
This interface is used by .NET to compare two objects, and you can customize the implementation.

I prefer to use the standard .NET feature rather than add a `Func`, which would be called thousands of times. And **Autocomplete** is primarily designed to handle a very large number of items. Adding an external call each time can be critical.

dvoituron (changes requested):

src/Core/Components/List/ListComponentBase.razor.cs
MarvinKlein1508: Do you think it will make that much of a difference? You still need to go through all items in the internal list. And how many items to you expect the user to select? 

I mean sure we can use `IComparer` but then you are limited to one compare definition. You still need to loop through the list and you'll need to implement `IComparer` for your objects and this will impact other components as well since you cannot skip looping the list if the object implements IComparer. With this method you actually can

dvoituron: Sorry, I meant [IEqualityComparer<T>](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.iequalitycomparer-1)

dvoituron: First `IEqualityComparer` is a default class used in all .NET methods: `Contains`, `Distinct`, `Except`, ...

I think it can make a difference. Because I don't think this is the only place where this comparison needs to be made (to verify). Once this property is available for **all** lists, whether **multiple** or not, this comparison tool should be applied.

```csharp
[Parameter]
public IEqualityComparer<TOption>? OptionComparer { get; set; }
```

```csharp
if (OptionComparer is null && _selectedOptions.Contains(item))
{
    RemoveSelectedItem(item);
    await RaiseChangedEventsAsync();
}
else if (OptionComparer is not null && _selectedOptions.Find(x => OptionComparer.Equals(x, item)) is TOption addedItem)
{
    RemoveSelectedItem(addedItem);
    await RaiseChangedEventsAsync();
}
```

The usage is similar (longer): you create a **Comparison** class and you set the instance: `OptionComparer="@MyCompare.Instance"`

```csharp
class MyCompare : IEqualityComparer<SimplePerson>
{
    public static readonly MyCompare Instance = new MyCompare();

    public bool Equals(SimplePerson? x, SimplePerson? y)
    {
        return x?.Firstname == y?.Firstname && x?.Lastname == y?.Lastname;
    }

    public int GetHashCode(SimplePerson obj) => HashCode.Combine(obj.Firstname, obj.Lastname);
}
```

dvoituron: Sorry. I think I'm nitpicking. A `Func` will probably do the stuff. 

On the other hand, you'll need to check how its use in all the other lists.

MarvinKlein1508: Guess you never stop learning. I thought I need to implement this on class level. But writing custom comparers is pretty smart! :) I've updated the PR to use `IEqualityComparer<TOption>?`. Thanks @dvoituron ! :)

MarvinKlein1508:
btw. can you tell me how you can merge the master this fast into the branch? In my own projects I always need to pull it in VS and merge it with the git menu.

src/Core/Components/List/ListComponentBase.razor.cs
MarvinKlein1508: @dvoituron the other lists aren't affected since they provide no way to get items from a different `IEnumerable`. Only the `FluentAutocomplete` does so with it's search function.

MarvinKlein1508:
@dvoituron would it be a good idea to overwrite `OptionComparer` in the other lists and mark it there as `Obsolete`? So no people will try to use it there.

dvoituron:
No. If we add a feature, we have to develop it from A-Z. Why should it only be for Autocomplete?

MarvinKlein1508:
> No. If we add a feature, we have to develop it from A-Z. Why should it only be for Autocomplete?

Because all other lists doesn't allow you to fetch different instances. The all have one parameter to provide items. Only `FluentAutocomplete` allows you to get items with new instances using it's `OnSearch` parameter.

MarvinKlein1508:
@dvoituron I have added tests for the other components too. `FluentCombobox` doesn't support `multiple` so it doesn't affect this component at all.

The other two components can only be affected when your binded list contains different instances then the provided list via item parameter. This could only happen if you fetch this the list which is bound to from the database and a second list of items which contain all items.

dvoituron:
If this parameter is only used with the **Autocomplete**, why to add this parameter in the **ListBase**?

You can add this parameter in the **Autocomplete** and override `OnSelectedItemsChangeHandlerAsync`. And calling the Base method to avoid to rewrite all. 

This is logical to have this parameter only for the Autocomplete.

MarvinKlein1508:
@dvoituron forget what I said about only for `FluentAutoComplete`. Every component which support `Multiple` is benefitting from this change.

`FluentListbox` and `FluentSelect` behaved the same when you did the following:

1. Load your `IEnumerable` with assigned items from the database for your specific object 
2. Load all items for these components from the database and provide them via `Items` parameter

This will also fix this behaviour as well :)

src/Core/Components/List/ListComponentBase.razor.cs
```diff
@@ -538,6 +543,11 @@ protected virtual async Task OnSelectedItemChangedHandlerAsync(TOption? item)
                 RemoveSelectedItem(item);
                 await RaiseChangedEventsAsync();
             }
+            else if (OptionComparer is not null && _selectedOptions.FirstOrDefault(x => OptionComparer.Equals(x, item)) is TOption addedItem)
```
dvoituron: Can you add a test `OptionComparer is null` from the previous `if` and use `Find` (and not `FirstOrDefault` (for better performances)

```csharp
if (OptionComparer is null && _selectedOptions.Contains(item))
{
    RemoveSelectedItem(item);
    await RaiseChangedEventsAsync();
}
else if (OptionComparer is not null && _selectedOptions.Find(x => OptionComparer.Equals(x, item)) is TOption addedItem)
{
    RemoveSelectedItem(addedItem);
    await RaiseChangedEventsAsync();
}
```

src/Core/Components/List/ListComponentBase.razor.cs
```diff
@@ -128,6 +128,11 @@ protected string? InternalValue
     /// </summary>
     [Parameter]
     public virtual Func<TOption, bool>? OptionSelected { get; set; }
+    /// <summary>
```
dvoituron, line 132: Can you add a space between the two properties?

src/Core/Components/List/ListComponentBase.razor.cs
```diff
@@ -128,6 +128,11 @@ protected string? InternalValue
...
+    /// Gets or sets the <see cref="IEqualityComparer{T}"/> used to determine if an option is already added to the internal list.
+    /// </summary>
```
dvoituron, line 135: Can you add `Available only when Multiple = true.`?

MarvinKlein1508: done.

src/Core/Components/List/ListComponentBase.razor.cs
MarvinKlein1508: done.

src/Core/Components/List/ListComponentBase.razor.cs
MarvinKlein1508: done

PR #3910, opened 2025-06-12
[dev-v5] Add DataGrid - part 1, the groundworks

vnbaaij:
- Add Enums
- Add doc start
- Add Paginator / PaginationState

There are still errors because of references in the XML comments. These will be resolved with a later PR

Paginator:
image


src/Core/Components/Icons/CoreIcons.cs
```diff
@@ -11,13 +11,32 @@ internal static partial class CoreIcons
 {
     /*
      * *************************** NOTES ***************************
-     * 
+     *
      * Try to use only the Regular.Size20 and Filled.Size20 icons,
      * to avoid duplicating the same icon in different sizes.
-     * 
+
```
dvoituron: Can you rollback this change?

src/Core/Components/Icons/CoreIcons.cs
```diff
@@ -11,13 +11,32 @@ internal static partial class CoreIcons
...
      * *************************************************************
      */
 
+    internal static partial class Regular
+    {
+        // TODO: We might need these for the FluentDataGrid header UI. Size 20 couldbe too big. Will delete if not needed
+        [global::System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage]
+        internal static partial class Size16
```
dvoituron: I don't think we need to distinct the CoreIcons sizes: this is vectorial images and the size difference is very small.
Examples: 
 - ArrowSort: From Size16 to Size20   = +50 chars
 - ChevronLeft: From Size24 to Size20 = -14 chars

I found better to simplify the code using only the common size: Size20 for all "internal" icons.

spelling.dic
```diff
@@ -84,3 +84,11 @@ menuclicked
 menuchecked
 rendertree
 testid
+columnheader
```
dvoituron: Can you sort the list alphabetically :-)
(see the comment in the top of the file)

src/Core/Infrastructure/EventCallbackSubscribable.cs
```diff
@@ -0,0 +1,37 @@
...
+
+/// <summary>
+/// Represents an event that you may subscribe to. This differs from normal C# events in that the handlers
+/// are EventCallback<typeparamref name="T"/>, and so may have async behaviors and cause component re-rendering
+/// while retaining error flow.
+/// </summary>
+/// <typeparam name="T">A type for the eventargs.</typeparam>
+internal sealed class EventCallbackSubscribable<T>
+{
+    private readonly Dictionary<EventCallbackSubscriber<T>, EventCallback<T>> _callbacks = [];
```
dvoituron: Can you replace the `Dictionary` with a `ConcurrentDictionary`?

src/Core/Enums/DataGridDisplayMode.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// With this mode fr units can be used to set the column widths.
+    /// </summary>
+    Grid,
+
+    /// <summary>
+    /// Uses HTML table elements only to render the DataGrid.
+    /// With this mode fr units cannot be used to set the column widths.
+    /// </summary>
+    Table,
+
```
dvoituron: Can you remove this extra line?

src/Core/Enums/DataGridRowType.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    /// <summary>
+    /// A normal row .
+    /// </summary>
+    Default,
+
+    /// <summary>
+    /// A header row.
+    /// </summary>
+    Header,
```
dvoituron, line 24: Wouldn't it be better to add a `Description` attribute as well?

src/Core/Components/Pagination/PaginationState.cs
```diff
@@ -0,0 +1,104 @@
...
+
+    /// <summary>
+    /// An event that is raised when the total item count has changed.
+    /// </summary>
+    public event EventHandler<TotalItemCountChangedEventArgs>? TotalItemCountChanged;
+
+    internal EventCallbackSubscribable<PaginationState> CurrentPageItemsChanged { get; } = new();
+    internal EventCallbackSubscribable<PaginationState> TotalItemCountChangedSubscribable { get; } = new();
+
+    /// <inheritdoc />
```
dvoituron, line 46: To verify but the Doc Generator need to add `cref`
```
<inheritdoc cref="..." />
```

src/Core/Components/Pagination/PaginationState.cs
```diff
@@ -0,0 +1,104 @@
...
+            return Task.CompletedTask;
+        }
+
+        TotalItemCount = totalItemCount;
+
+        if (CurrentPageIndex > 0 && CurrentPageIndex > LastPageIndex)
+        {
+            // If the number of items has reduced such that the current page index is no longer valid, move
+            // automatically to the final valid page index and trigger a further data load.
+            _ = SetCurrentPageIndexAsync(LastPageIndex.Value);
```
dvoituron: Are you sure you are not using the keyword `await`?
This method will be executed in a separate thread without waiting for the execution to finish.

Probably better to add `async` and `await` :-)

src/Core/Components/Pagination/FluentPaginator.razor.cs
```diff
@@ -0,0 +1,95 @@
...
+    /// <summary>
+    /// Disables the pagination buttons
+    /// </summary>
+    [Parameter]
+    public bool Disabled { get; set; }
+
+    /// <summary>
+    /// Gets or sets the associated <see cref="PaginationState"/>. This parameter is required.
+    /// </summary>
+    [Parameter, EditorRequired]
```
dvoituron, line 54: You can remove the `EditorRequired` and add `required`. So, you can remove the `default!`

```csharp
public required PaginationState State { get; set; }
```

src/Core/Components/Pagination/FluentPaginator.razor.cs
```diff
@@ -0,0 +1,95 @@
...
+    /// {your State parameter name}.CurrentPageIndex (zero-based, so +1 for the current page number)
+    /// {your State parameter name}.LastPageIndex (zero-based, so +1 for the total number of pages)
+    /// </summary>
+    [Parameter]
+    public RenderFragment? PaginationTextTemplate { get; set; }
+
+    /// <summary>
+    /// Constructs an instance of <see cref="FluentPaginator" />.
+    /// </summary>
+    public FluentPaginator()
```
dvoituron: Can you move the ctor before the properties?
We can respect the order:
- Private variables
- Constructors
- properties
- Methods
- private methods

src/Core/Components/Pagination/FluentPaginator.razor.cs
```diff
@@ -0,0 +1,95 @@
...
+    {
+        // The "total item count" handler doesn't need to do anything except cause this component to re-render
+        _totalItemCountChanged = new(new EventCallback<PaginationState>(this, null));
+        _currentPageItemsChanged = new(new EventCallback<PaginationState>(this, null));
+    }
+
+    private Task GoFirstAsync() => GoToPageAsync(0);
+    private Task GoPreviousAsync() => GoToPageAsync(State.CurrentPageIndex - 1);
```
dvoituron, line 94: Can you add a empty line bewteen all methods?

src/Core/Components/Pagination/FluentPaginator.razor.cs
```diff
@@ -0,0 +1,95 @@
...
+    private Task GoNextAsync() => GoToPageAsync(State.CurrentPageIndex + 1);
+    private Task GoLastAsync() => GoToPageAsync(State.LastPageIndex.GetValueOrDefault(0));
+
+    private bool CanGoBack => State.CurrentPageIndex > 0;
```
dvoituron: Can you add the properties before the methods?

src/Core/Components/Pagination/FluentPaginator.razor.cs
```diff
@@ -0,0 +1,95 @@
...
+        if (CurrentPageIndexChanged.HasDelegate)
+        {
+            await CurrentPageIndexChanged.InvokeAsync(State.CurrentPageIndex);
+        }
+    }
+
+    /// <inheritdoc />
+    protected override void OnParametersSet()
+    {
+        _totalItemCountChanged.SubscribeOrMove(State.TotalItemCountChangedSubscribable);
```
dvoituron: Since the `State` is required, you should have this State value in the `OnInitialized` method?
The `OnParametersSet` method is called every time the component’s parameters are set or updated, not just during the initial rendering. So many many times :-)

src/Core/Components/Pagination/FluentPaginator.razor
```diff
@@ -0,0 +1,42 @@
+﻿@namespace Microsoft.FluentUI.AspNetCore.Components
+@inherits FluentComponentBase
+<div class="paginator" style="@Style">
+    @if (State.TotalItemCount.HasValue)
+    {
+        <div class="summary">
```
dvoituron, line 10: This class is not defined in the CSS

src/Core/Components/Pagination/FluentPaginator.razor
```diff
@@ -0,0 +1,42 @@
...
+            {
+                @SummaryTemplate
+            }
+            else
+            {
+                <text><strong>@State.TotalItemCount</strong> items</text>
+            }
+        </div>
+        <nav role="navigation" class="paginator-nav">
+            <FluentButton @onclick="GoFirstAsync" Disabled="@(!CanGoBack || Disabled)" title="Go to first page" aria-label="Go to first page">
```
dvoituron: You need to use the Localizer to "translate" the text "Go to first page" (same for other texts)
See Localization/Readme.md file

src/Core/Components/Pagination/FluentPaginator.razor
```diff
@@ -0,0 +1,42 @@
...
+                <FluentIcon Value="@(new CoreIcons.Regular.Size20.ChevronDoubleLeft())" Width="20px" />
+            </FluentButton>
+            <FluentButton @onclick="GoPreviousAsync" Disabled="@(!CanGoBack || Disabled)" title="Go to previous page" aria-label="Go to previous page">
```
dvoituron: You need to use `OnClick` and not `@onclick` attribute.
Why not to use the `IconOnly` property to avoid to include a single `FluentIcon`?

src/Core/Components/Pagination/FluentPaginator.razor.css
```diff
@@ -0,0 +1,25 @@
+.paginator {
```
dvoituron: You need to use "fluentui" classes. A name `paginator` can be in conflict with other names used by the dev.
For example: `.fluent-paginator`

And the other classes must be "sub classes": `.fluent-paginator .pagination-text`, `.fluent-paginator .paginator-nav`

src/Core/Components/Pagination/FluentPaginator.razor.css
```diff
@@ -0,0 +1,25 @@
...
+    padding: 0; 
+    display: flex;
+    margin-inline-start: auto;
+    margin-inline-end: 0;
+    gap: 0.5rem;
+    align-items: center;
+}
+
+    
+[dir="rtl"] * ::deep fluent-button > svg {
```
dvoituron: `::deep` has no effect in v5. So, you need to add a class to apply this style only for your component.

`[dir="rtl"] .fluent-paginator fluent-button > svg {`

dvoituron (changes requested):

spelling.dic
vnbaaij: This list is maintained from VS. I just click ignore in the code action and then it gets added here. Don't think we need to keep it sorted.

src/Core/Components/Icons/CoreIcons.cs
vnbaaij: We'll go with 20 for now. But as said, they might turn out to be too big. We'll see later

spelling.dic
dvoituron: The longer the list gets, the harder it will be to maintain. While it's not mandatory, keeping things in order simplifies the process.

src/Core/Components/Pagination/FluentPaginator.razor
vnbaaij: In the spirit of doing smaller PRs, I planned to do that in a later PR

dvoituron: A small PR must be completed. Using `Localizer[Localization.LanguageResource.Paginator_GoFirstPage]` and not an hardcoded string is not too complex :-)

src/Core/Components/Pagination/FluentPaginator.razor.cs
vnbaaij: I want EditorRequired to indicate to the devolper in VS/VS Code the need to provide a value for the parameter

dvoituron: Make sense, but in this case the `State` value must be set for the Razor part (`EditorRequired`) but also for the C# part (`required`). So, probably better to keep the both.
What disturbs me most is the `Default!`, which is no longer necessary with `required`.

spelling.dic
vnbaaij: But that is the point. We are not maintaining it. It is the tool that adds things here

src/Core/Components/Pagination/FluentPaginator.razor
vnbaaij: But with it being here we allow customization by the dev, no?

dvoituron: Oh. OK... In this case, I propose to add an empty class in the CSS (to confirm that we are "using" this class)

src/Core/Components/Pagination/FluentPaginator.razor
vnbaaij: OnClick done and changed to use IconStart

vnbaaij: I'll do the translations after all comments processed

src/Core/Components/Pagination/FluentPaginator.razor.cs
vnbaaij: I added both

src/Core/Components/Pagination/PaginationState.cs
vnbaaij: Added

src/Core/Enums/DataGridRowType.cs
vnbaaij: The description value would then be exactly the same as the value itself. We can add it, but it doesn't 'add' anything

src/Core/Components/Pagination/FluentPaginator.razor.cs
vnbaaij: Order has been changed (using Copilot Agent

src/Core/Components/Icons/CoreIcons.cs
```diff
@@ -26,14 +25,47 @@ internal static partial class Regular
...
+            public class TableResizeColumn : Icon { public TableResizeColumn() : base("TableResizeColumn", IconVariant.Regular, IconSize.Size20, "<path d=\"M7.35 8.15c.2.2.2.5 0 .7l-.64.65h6.58l-.64-.65a.5.5 0 0 1 .7-.7l1.5 1.5c.2.2.2.5 0 .7l-1.5 1.5a.5.5 0 0 1-.7-.7l.64-.65H6.71l.64.65a.5.5 0 0 1-.7.7l-1.5-1.5a.5.5 0 0 1 0-.7l1.5-1.5c.2-.2.5-.2.7 0ZM17 6a3 3 0 0 0-3-3H6a3 3 0 0 0-3 3v8a3 3 0 0 0 3 3h8a3 3 0 0 0 3-3V6Zm-4-2v3c.36 0 .72.13 1 .38V4a2 2 0 0 1 2 2v8a2 2 0 0 1-2 2v-3.38a1.5 1.5 0 0 1-1 .38v3H7v-3a1.5 1.5 0 0 1-1-.38V16a2 2 0 0 1-2-2V6c0-1.1.9-2 2-2v3.38A1.5 1.5 0 0 1 7 7V4h6Z\"/>") { } }
         }
     }
 
+    internal static partial class Regular
+    {
+        [global::System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage]
+        internal static partial class Size24
+        {
+            public class ChevronLeft : Icon { public ChevronLeft() : base("ChevronLeft", IconVariant.Regular, IconSize.Size24, "<path d=\"M15.53 4.22c.3.3.3.77 0 1.06L8.81 12l6.72 6.72a.75.75 0 1 1-1.06 1.06l-7.25-7.25a.75.75 0 0 1 0-1.06l7.25-7.25c.3-.3.77-.3 1.06 0Z\"/>") { } }
```
dvoituron: I think these 2 icons can also be included in the Size20

src/Core/Microsoft.FluentUI.AspNetCore.Components.csproj
```diff
@@ -117,4 +117,7 @@
       <NoWarn>CS1591</NoWarn>
     </EmbeddedResource>
   </ItemGroup>
+  <ItemGroup>
+    <Folder Include="Components\DataGrid\" />
```
dvoituron: Can you remove this line?

tests/Core/Components.Tests.csproj
```diff
@@ -73,4 +73,8 @@
       <DependentUpon>FluentLocalizer.resx</DependentUpon>
     </Compile>
   </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="Components\Paginator\" />
```
dvoituron: Can you remobe this line?

tests/Core/Components/Paginator/FluentPaginatorTests.FluentPaginator_Renders_CustomSummaryTemplate.verified.razor.html
```diff
@@ -0,0 +1,29 @@
...
+      <svg style="width: 20px;" focusable="false" viewBox="0 0 20 20" aria-hidden="true" blazor:onkeydown="x" blazor:onclick="x">
+        <path d="M11.35 15.85a.5.5 0 0 1-.7 0L5.16 10.4a.55.55 0 0 1 0-.78l5.49-5.46a.5.5 0 1 1 .7.7L6.2 10l5.16 5.15c.2.2.2.5 0 .7Zm4 0a.5.5 0 0 1-.7 0L9.16 10.4a.55.55 0 0 1 0-.78l5.49-5.46a.5.5 0 1 1 .7.7L10.2 10l5.16 5.15c.2.2.2.5 0 .7Z"></path>
+      </svg>
+    </fluent-button>
+    <fluent-button icon-only="" disabled="" title="Go to previous page" blazor:onclick="x" aria-label="Go to previous page">
+      <svg style="width: 24px;" focusable="false" viewBox="0 0 24 24" aria-hidden="true" blazor:onkeydown="x" blazor:onclick="x">
+        <path d="M15.53 4.22c.3.3.3.77 0 1.06L8.81 12l6.72 6.72a.75.75 0 1 1-1.06 1.06l-7.25-7.25a.75.75 0 0 1 0-1.06l7.25-7.25c.3-.3.77-.3 1.06 0Z"></path>
+      </svg>
+    </fluent-button>
+    <div class="pagination-text">Page &lt;strong&gt;1&lt;/strong&gt; of &lt;strong&gt;10&lt;/strong&gt;</div>
```
dvoituron: I don't think this is correct to add the strong text and not the HTML tag.

src/Core/Components/Pagination/FluentPaginator.razor.css
```diff
@@ -0,0 +1,28 @@
...
+  /*border-top: 1px solid var(--neutral-stroke-divider-rest);*/
+  margin-top: 0.5rem;
+  padding: 0.25rem 0;
+  align-items: center;
+}
+
+  .fluent-paginator .summary {
+  }
+
+  .fluent-paginator .pagination-text {
```
dvoituron, line 12: Is it correct to have "pagination" and not "paginator"?

dvoituron (changes requested):

src/Core/Components/Pagination/FluentPaginator.razor.css
vnbaaij: The object that does stuff is the Paginator. The things it shows are pagination elements.

tests/Core/Components/Paginator/FluentPaginatorTests.FluentPaginator_Renders_CustomSummaryTemplate.verified.razor.html
vnbaaij: Agreed. Changed it to only do the text which is specified in the resx file (with replacements for current and max). If formatting is desired, they can use the `PaginationTextTemplate` parameter

src/Core/Components/Icons/CoreIcons.cs
vnbaaij: Done

vnbaaij: Done

PR #3927, opened 2025-06-17
[DataGrid] Add IsFixed parameter

### LINKED ISSUE #3911: 4.12.0 Bug on FluentDataGrid
Hello,

Today, I performed an upgrade of the package from version 4.11.9 to 4.12.0. I did not make any changes to my code — I only updated the package via NuGet.

However, it appears that something is behaving differently than usual.

Issue 1: Table does not reflect new items instantly
Previously, when I added a new item to the collection bound to a DataGrid table, the new item would appear instantly without needing to call InvokeAsync(StateHasChanged) or @ref.RefreshDataAsync.

But today, after the upgrade, the item does not appear immediately. Initially, I thought that perhaps the new version requires manual intervention to refresh the UI. However, after filtering/searching the table, I noticed that the new data is actually present — it's just not displayed initially.
So I suspect this might be a bug introduced in version 4.12.0.

Issue 2: Blue line appearing in the table
Secondly, I noticed a blue line appearing in the table, and I’m not sure what it represents.

I have several projects using FluentUI, and after checking them all — only those that were upgraded today are affected. This issue is consistent across upgraded projects.

Image

Image

### LINKED ISSUE #3920: feat: Add IsFixed parameter to DataGrid
Add a boolean parameter `IsFixed` to the FluentDataGrid component (like the IsFixed parameter that exists on the standard Blazor Cascading Value component). 

The default value of false means the grid continues to behave like it did in v4.11.9. When set to true, the developer can indicate that the grid's dataset is not expected to change during its lifetime.
This replaces the new refresh logic that was implemented in v4.12.0

vnbaaij:
This PR adds an `IsFixed` parameter to the `FluentDataGrid` component that allows developers to optimize performance when working with static datasets that are not expected to change during the grid's lifetime. Fixes #3911. Fixes #3920.

### New Parameter
- Added `IsFixed` boolean parameter with default value `false` to maintain backward compatibility
- When `IsFixed=true`, the grid optimizes refresh behavior for static datasets
- When `IsFixed=false` (default), maintains existing behavior from v4.11.9

### Performance Optimizations
When `IsFixed=true`:
- Skips hash computation for Items collection change detection
- Reduces automatic `RefreshItems` calls to only forced or initial requests
- Maintains full functionality for pagination, explicit refreshes, and data source changes

### Usage Example
```razor
<!-- Default behavior (existing functionality) -->
<FluentDataGrid Items="@dynamicData.AsQueryable()">
    <PropertyColumn Property="@(x => x.Name)" />
</FluentDataGrid>

<!-- Optimized for static datasets -->
<FluentDataGrid Items="@staticData.AsQueryable()" IsFixed="true">
    <PropertyColumn Property="@(x => x.Name)" />
</FluentDataGrid>
```

### Implementation Details
- Modified `OnParametersSetAsync()` to respect the `IsFixed` parameter in change detection logic
- Enhanced `RefreshDataCoreAsync()` to be more conservative about automatic refreshes when `IsFixed=true`
- Added comprehensive test coverage including edge cases and explicit refresh scenarios

### Backward Compatibility
- Default value of `false` ensures existing code continues to work unchanged
- All existing functionality (pagination, sorting, explicit refreshes) remains fully functional
- No breaking changes to existing APIs

This implementation replaces the automatic refresh logic that was added in v4.12.0 with more controlled behavior when developers know their dataset is static, providing significant performance benefits for such scenarios.

### DIFF
--- a/examples/Demo/Shared/Microsoft.FluentUI.AspNetCore.Components.xml
+++ b/examples/Demo/Shared/Microsoft.FluentUI.AspNetCore.Components.xml
@@ -1954,7 +1954,7 @@
         <member name="P:Microsoft.FluentUI.AspNetCore.Components.FluentDataGrid`1.ResizeColumnOnAllRows">
             <summary>
             Gets or sets a value indicating whether column resize handles should extend the full height of the grid.
-            When true, columns can be resized by dragging from any row. When false, columns can only be resized 
+            When true, columns can be resized by dragging from any row. When false, columns can only be resized
             by dragging from the column header. Default is true.
             </summary>
         </member>
@@ -2140,6 +2140,13 @@
             Gets or sets a value indicating whether the grids' first cell should be focused.
             </summary>
         </member>
+        <member name="P:Microsoft.FluentUI.AspNetCore.Components.FluentDataGrid`1.IsFixed">
+            <summary>
+            Gets or sets a value indicating whether the grid's dataset is not expected to change during its lifetime.
+            When set to true, reduces automatic refresh checks for better performance with static datasets.
+            Default is false to maintain backward compatibility.
+            </summary>
+        </member>
         <member name="M:Microsoft.FluentUI.AspNetCore.Components.FluentDataGrid`1.#ctor">
             <summary>
             Constructs an instance of <see cref="T:Microsoft.FluentUI.AspNetCore.Components.FluentDataGrid`1"/>.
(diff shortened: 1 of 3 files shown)


Paxol:
Hello, is there any ETA for when this PR will be merged into a public version of the NuGet package (even a prerelease)?
I’m currently experiencing an issue with version 4.12.0 and Blazor WASM when using the OData adapter. The `ComputeItemsHash` function is called with the `Items` parameter and attempts to enumerate the query synchronously. 
I believe this pull request would resolve that problem.

vnbaaij:
There are pre-release packages available already. See https://github.com/microsoft/fluentui-blazor/blob/dev/docs/using-latest-daily.md

v4.12.1 final will probably be available next week

PR #3930, opened 2025-06-17
[Autocomplete] Fix keyboard usage in Single mode

dvoituron:
# Fix Autocomplete Keyboard in Single mode

1. Allow an item to be deselected using the `Delete` or `Backspace` keys.
2. Update the **SelectableItem** item to get the first “not disabled” item.`

In this example, all items are selected using the keyboard only (arrows, Delete, Backspace keys)

peek_1

## Unit Tests

No change

### DIFF
--- a/src/Core/Components/List/FluentAutocomplete.razor
+++ b/src/Core/Components/List/FluentAutocomplete.razor
@@ -216,7 +216,14 @@
 
                     if (Multiple == false)
                     {
-                        <FluentLabel tabindex="0" aria-label="@GetOptionText(item)" role="checkbox" aria-checked="true">@text</FluentLabel>
+                        <FluentLabel Id="@($"{Id}-single")"
+                                     tabindex="0"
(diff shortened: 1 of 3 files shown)


MarvinKlein1508:
@dvoituron is there also an easy way to get keyboard working again when an item has already been selected? 

For example:
Albania is preselected and I click on the component and start typing Germany so it goes back to null to allow the user to select another item with search function without the need to clear the component first?

dvoituron:
> @dvoituron is there also an easy way to get keyboard working again when an item has already been selected?

This is only possible after removing the selected item. The "Text" field is not available when an item is selected.
However, you can use the values `Multiple=True` and `MaximumSelectedOptions=1` to do that.

Tyme-Bleyaert:
I had a bit different functionality in mind, in my local branch i could also partially remove text from the selected item.
But i guess this works too.

PR #3999, opened 2025-07-10
[NavLink] Track active state for menu items with an OnClick handler

### LINKED ISSUE #3992: fix: Active class css in FluentNavLink and FluentNavGroup
# 🐛 Bug Report

When using either `FluentNavLink` or `FluentNavGroup` when also using `OnClick` event callback, the active css class is not applied.

## 🤔 Expected Behavior

When clicking a link we should still see which page we are on.

## 😯 Current Behavior

Currently when you have an eventcallback for example closing an hamburger menu, the active class css is ignored.

vnbaaij:
Fix #3992. To fix this issue, we need to do the following:

1. The active class is put on the `<a>` tag that gets created by the standard NavLink component. We have no NavLink in the situation on a FluentNavLink with an OnClick handler, so  we must introduce a new element (`<span>`) that we can put the class on. This also means we need to re-create the unit test's verified files 

2. The menu css specifically looks for a.active and then `positioning-region` to draw the active bar. Because we add a span in 1) we need to change the CSS

3.  And now for the really hard part...how do we know when the menu item with event callback is active? Actually, we don't. 
Only way to find out is to do what normally gets done in the standard NavLink component. Sounds trivial but it is actually quite a bit of code (copied and adapted from standard Blazor NavLink component)

This PR implements these steps


src/Core/Components/NavMenu/FluentNavLink.razor.cs
```diff
@@ -33,4 +40,213 @@ public FluentNavLink()
...
+        {
+            _isActive = shouldBeActiveNow;
+            UpdateCssClass();
+            StateHasChanged();
+        }
+    }
+
+    /// <inheritdoc />
+    protected override void OnParametersSet()
+    {
```
dvoituron, line 67: Is it possible to move this code to the `OnInitialized` or `OnAfterRendered` methods? To avoid to execute this code multiple times, for each parameter changes.

dvoituron (changes requested):

src/Core/Components/NavMenu/FluentNavLink.razor.cs
vnbaaij: This is how it is done in the aspnetcore code as well. All parameter changes can lead to active state changes so must be evaluated every time

dvoituron: I'm not sur this is efficient when executing a "large" method. But OK.

dvoituron (approved):

Issue #4013, opened 2025-07-15
docs: Razor tab is blank for everything


IeuanWalker:
The example tab in the docs are working, but when you click razor to see the code, its blank - 

<img width="819" height="441" alt="Image" src="https://github.com/user-attachments/assets/ee1255ee-94cf-48f3-b151-b17868fc324c" />

MarvinKlein1508:
Can confirm this as well.

vnbaaij:
Yup, see it too...Not really sue yet how that happened. Looking at it now,

vnbaaij:
While trying to find out what is wrong, please use https://ambitious-island-005801e03.2.azurestaticapps.net/ for now. As far as I can see, the razor tab are working there.

Would be great if anyone can confirm and report back here.

I'm moving some stuff around, so there will probably be more disturbances in the coming hours/days

lsurma:
> While trying to find out what is wrong, please use https://ambitious-island-005801e03.2.azurestaticapps.net/ for now. As far as I can see, the razor tab are working there.
> 
> Would be great if anyone can confirm and report back here.
> 
> I'm moving some stuff around, so there will probably be more disturbances in the coming hours/days

It seems to be working for me.

vnbaaij:
I found what was causing the Razor files being not found and resolved it. Everything should be back to normal.

PR #4016, opened 2025-07-15
[General] Update GH Actions

vnbaaij:
Moving things to new subscription


examples/Demo/Client/wwwroot/staticwebapp.config.json
```diff
@@ -1,6 +1,6 @@
 {
   "navigationFallback": {
     "rewrite": "/index.html",
-    "exclude": [ "_content/FluentUI.Demo.Shared/sources/*" ]
+    "exclude": [ "_content/FluentUI.Demo.Shared/sources/*.txt" ]
```
StevenRasmussen: @vnbaaij, @dvoituron - Is this perhaps the cause of #4013?

vnbaaij: It was always there before so it should not be the cause.

dvoituron (approved):

dvoituron (approved):

PR #4046, opened 2025-07-30
[dev-v5] Add missing files and code to publish a standalone WASM demo

dvoituron:
# [dev-v5] Add missing files and code to publish a standalone WASM demo site

Add/move some wwwroot files to the Client project to enable the demo website to be published as a WASM project.

1. Use this command to generate all static files:
```
$\examples\Demo\FluentUI.Demo.Client  ❯  dotnet publish -c Release -o C:\MyDemo\
```

2. Use this command to host the static file locally:
```
dotnet serve -d C:\MyDemo\ --port 56770
```


examples/Demo/FluentUI.Demo.Client/Program.cs
```diff
@@ -3,10 +3,14 @@
 // ------------------------------------------------------------------------
 
 using FluentUI.Demo.Client;
+using Microsoft.AspNetCore.Components.Web;
 using Microsoft.AspNetCore.Components.WebAssembly.Hosting;
 using Microsoft.FluentUI.AspNetCore.Components;
 
 var builder = WebAssemblyHostBuilder.CreateDefault(args);
+builder.RootComponents.Add<Routes>("#app");
+builder.RootComponents.Add<HeadOutlet>("head::after");
```
vnbaaij, line 12: Isn't this a .NET 9 only construct? It is fine by me, but if 9 only we should mention it in the PR description (and other docs somewhere as well)

dvoituron: This is because the csproj is `<TargetFramework>net9.0</TargetFramework>`
And I think that is compatible with .NET8 and .NET10, using the same syntax.

PR #4093, opened 2025-08-29
[MenuButton] Accessibility fixes and allow custom button content

### LINKED ISSUE #4090: fix: tab/shift-tab does not work as expected in MenuButton
# 🐛 Bug Report

The MenuButton component is not accessible to keyboard users. When opening the MenuButton, entering tab should advance to the first element in the menu. Shift-tab should place focus back on the button.

## 💻 Repro or Code Sample

Can be reproduced in any of the MenuButton examples -> https://www.fluentui-blazor.net/MenuButton

## 💁 Possible Solution

Register an onkeydown listener on both the first menu item (if one exists) and the menu button. If focus is on the menu button, tab will place focus on the first item in the menu. If focus is on the first menu item, a shift-tab will place focus back on the button.

## 🔦 Context

Caught in manual testing. I will provide a fix

## 🌍 Your Environment

adamint:
# Pull Request

## 📖 Description

I really like this component, but it's currently unusable in Aspire because it's not keyboard-accessible. To make it accessible, the menu button needs to properly handle tab (tab on button menu is open focuses first menu item, tab when focus is on a menu item will focus the next focusable dom element outside the menu + close the menu), and shift+tab (shift+tab when focus is on the button will close the menu if open, or otherwise will focus the button if a menu item is focused).

Additionally, the button content was not customizable (no render fragment could be provided).

This PR fixes both problems. I added a new parameter `ButtonContent` that can be provided instead of `Text`. 

The accessibility changes were implemented client-side because of the focusing logic requiring DOM access. I also had to fix the FluentAnchoredRegion's `FocusableElement.findNextFocusableElement` function to support focusing fluent components that have focusable children (since they themselves are not focusable), which I reused in the FluentMenuButton. Please refer to this before and after:

Before (see keys pressed on screen for context)

https://github.com/user-attachments/assets/617c8470-b9a8-453f-afeb-8eeeaf5f77c5

After

https://github.com/user-attachments/assets/6d8e41b0-07e9-487f-abe2-4c17a78c71c0

### 🎫 Issues

Fixes #4090

## 👩‍💻 Reviewer Notes

N/A

## 📑 Test Plan

I added a couple tests for the `ButtonContent` parameter, but as most of the accessibility changes are in the component JS, you will need to test manually using the MenuButton sample page.

### General

### Component-specific

- [X] I have validated the [Unit Tests](https://github.com/Microsoft/fluentui-blazor/blob/master/unit-tests.md) for an existing component


src/Core/Components/MenuButton/FluentMenuButton.razor.cs
```diff
@@ -80,19 +85,57 @@ public partial class FluentMenuButton : FluentComponentBase
...
+    private IJSRuntime JSRuntime { get; set; } = default!;
+
+    /// <summary />
+    [Inject]
+    private LibraryConfiguration LibraryConfiguration { get; set; } = default!;
+
     protected override void OnInitialized()
     {
         _buttonId = Identifier.NewId();
+        _menuId = Identifier.NewId();
```
dvoituron: To be coherent with the button, can you set `_menuId = $"{_buttonId}-menu";`

src/Core/Components/MenuButton/FluentMenuButton.razor.cs
```diff
@@ -80,19 +85,57 @@ public partial class FluentMenuButton : FluentComponentBase
...
     }
 
     protected override void OnParametersSet()
     {
+        if (Text is not null && ButtonContent is not null)
```
dvoituron, line 104: I don't think you need to test this. But you must add to the Text property that this property will be ignored if ButtonContent is defined.

src/Core/Components/MenuButton/FluentMenuButton.razor.js
```diff
@@ -0,0 +1,96 @@
+let menuItemKeydownListener = null;
+let menuElement = null;
+
+let onButtonKeyDownListener = null;
+let buttonElement = null;
+
+export async function fluentMenuButtonOnRender(buttonId, menuId, anchoredRegionModule, dotNetHelper) {
```
dvoituron: Why are you using this JavaScript code and not C# code using FluentKeyCode for example?

src/Core/Components/MenuButton/FluentMenuButton.razor.js
```diff
@@ -0,0 +1,96 @@
...
+
+    if (!menuId) {
+        return;
+    }
+
+    const FocusableElement = anchoredRegionModule
+        ? anchoredRegionModule.FocusableElement
+        : null;
+    if (!FocusableElement) {
+        console.warn("FocusableElement not available from AnchoredRegion module");
```
dvoituron: We don't accept to display log in our Production applications. If there are no FocusableElement, the process can stop here.

src/Core/Components/MenuButton/FluentMenuButton.razor.js
```diff
@@ -0,0 +1,96 @@
...
+                }
+
+                dotNetHelper.invokeMethodAsync('ToggleMenuAsync');
+            } catch (ex) {
+                console.error("Failed to focus button:", ex);
+            }
+        }
+    };
+
+    menuElement.addEventListener("keydown", menuItemKeydownListener);
```
dvoituron: You need to test if menuElement and buttonElement are defined (exist): `if (menuElement) { ... }`

src/Core/Components/MenuButton/FluentMenuButton.razor.js
adamint: I didn’t want to split up the menu and button listeners, and the menu item listeners are in JS because they cannot be known before render.

I guess you could call back into blazor for the menu item events, but then you would have to invoke more JS to focus, so the JS feels a more appropriate place for the logic.

src/Core/Components/MenuButton/FluentMenuButton.razor.js
adamint: Done!

src/Core/Components/MenuButton/FluentMenuButton.razor.cs
adamint: Done!

src/Core/Components/MenuButton/FluentMenuButton.razor.cs
adamint: Done!

src/Core/Components/MenuButton/FluentMenuButton.razor.cs
```diff
@@ -80,19 +85,57 @@ public partial class FluentMenuButton : FluentComponentBase
...
+    protected override async Task OnAfterRenderAsync(bool firstRender)
+    {
+        if (firstRender)
+        {
+            _jsModule ??= await JSRuntime.InvokeAsync<IJSObjectReference>("import", JAVASCRIPT_FILE.FormatCollocatedUrl(LibraryConfiguration));
+            _anchoredRegionModule ??= await JSRuntime.InvokeAsync<IJSObjectReference>("import", ANCHORED_REGION_JAVASCRIPT_FILE.FormatCollocatedUrl(LibraryConfiguration));
+            _dotNetHelper = DotNetObjectReference.Create(this);
+        }
+
+        await _jsModule!.InvokeVoidAsync("fluentMenuButtonOnRender", _buttonId, _visible ? _menuId : null, _anchoredRegionModule, _dotNetHelper);
```
dvoituron: Please, don't use the null-forgiving operator. If the compiler says that there is a possibility that the `_jsModule` is null, you need to manage it.

```csharp
if (_jsModule is not null)
{
    wait _jsModule!.InvokeVoidAsync ...
}
```

src/Core/Components/MenuButton/FluentMenuButton.razor.js
```diff
@@ -0,0 +1,100 @@
+let menuItemKeydownListener = null;
```
dvoituron: Did you verify that?

These variables (menuElement, buttonElement, etc.) are defined at the module level, so they are shared across all instances of the component that use this JS file. This means if two FluentMenuButton components are used on the same page, their JS interop will conflict: the variables will be overwritten by whichever component was initialized last. The result is that event listeners and references will not be isolated per component instance, leading to bugs and unpredictable behavior.

You probably need to create a `const menuButtonState = new Map();` and
```js
// Store state for this instance
menuButtonState.set(buttonId, {
    menuElement,
    buttonElement,
    menuItemKeydownListener,
    onButtonKeyDownListener
});
```

vnbaaij:
@adamint Can you please finish this PR and fix the comments. Thanks.

src/Core/Components/MenuButton/FluentMenuButton.razor.js
adamint: fixed.

adamint:
@vnbaaij updated based on comments.

adamint:
Ok, I realized that this issue is not unique to just menu buttons and was happening to all menus. I have reverted the focus changes to the `FluentMenuButton`, and instead implemented them in `FluentMenu` so that we properly handle tab and escape in the `FluentDataGrid` and other uses of `FluentMenu` as well. I added an example of a fluent data grid that uses menus in its cell headers, which I used to help test the changes in this PR and which used to fail.

https://github.com/user-attachments/assets/ec584fc8-8002-4655-a78a-8bba19af8e97

src/Core/Components/Menu/FluentMenu.razor.cs
```diff
@@ -208,18 +210,22 @@ protected override void OnInitialized()
         base.OnInitialized();
     }
 
+    protected override void OnParametersSet()
+    {
+        Id ??= Identifier.NewId();
```
adamint: An ID is now always required to be able to attach listeners onto the menu.

src/Core/Components/Menu/FluentMenu.razor.cs
```diff
@@ -238,6 +244,11 @@ protected override async Task OnAfterRenderAsync(bool firstRender)
             }
         }
 
+        if (_jsModule is not null)
```
dvoituron: Why isn't this code in `firstRender`? In this case, the `initialize` method will be executed multiple times.

src/Core/Components/Menu/FluentMenu.razor.cs
```diff
@@ -309,6 +321,13 @@ public async Task OpenAsync(int screenWidth, int screenHeight, int x, int y)
 
     }
 
+    [JSInvokable]
+    public Task CloseMenuAsync()
```
dvoituron: Why did you created this method and not to use `CloseAsync` ?

src/Core/Components/Menu/FluentMenu.razor.cs
dvoituron: Can you move this `Id` initialization to the ctor? `OnParametersSet` will be called multiple times during the Blazor life cycle.

src/Core/Components/MenuButton/FluentMenuButton.razor.cs
dvoituron: You cannot force the developer to have a content. The dev can create a MenuButton without content.
So, just add in the Text property comment: `this property will be ignored if ButtonContent is defined`

src/Core/Components/Menu/FluentMenu.razor.cs
vnbaaij: Id is already getting a value assigned in the constructor (line 258)

src/Core/Components/Menu/FluentMenu.razor.js
```diff
@@ -31,3 +31,117 @@ export function isChecked(menuItemId) {
...
+    // Return if either the menu or anchor element have not rendered.
+    if (!menuElement || !anchorElement) {
+        return;
+    }
+
+    // We need to handle three cases to be fully accessible:
+    // 1. When Tab is pressed on the anchor, focus must be moved to the first focusable element in the menu
+    // 2. When Shift+Tab is pressed on any focusable element in the menu, focus must be moved back to the anchor. This will also close the menu.
+    // 3. When Tab is pressed on any focusable element in the menu, focus should continue to the next focusable element in the element's root. This will also close the menu.
+    // 4. When Escape is pressed on any focusable element in the menu, focus must be moved back to the anchor. This will also close the menu.
```
vnbaaij: You say three but have four cases :)

adamint: oops!! :smile:

updated the comment

vnbaaij:
Hi @adamint . Can we work to get this merged in. Please look at the latest open remarks. Thanks.

src/Core/Components/Menu/FluentMenu.razor.cs
adamint: It was leftover code I had not deleted. Removed, sorry

src/Core/Components/Menu/FluentMenu.razor.cs
adamint: I've added a comment and updated the behavior. There, `initialize` is run after every render, but `initialize` should actually be run any time `Opened` changes, as well as on first render. This is because when `Opened` is false, there is no rendered `FluentMenu` to attach event listeners to, so when `Opened` is set to false, we should dispose of the existing listeners, and when it is changed to true, add them.

src/Core/Components/MenuButton/FluentMenuButton.razor.cs
dvoituron: This is not correct with the new comment. You must remove this test: the dev can set a value in `Text` and `ButtonContent`. And the component will render only the `ButtonContent` (ignoring the `Text `prarameter).

vnbaaij:
@adamint, just one more small change requested...

PR #4095, opened 2025-08-29
Return focus when dialog is closed to element focused before open

### LINKED ISSUE #4094: fix: keyboard focus does not return to previous element after closing FluentDialog
# 🐛 Bug Report

As in title, keyboard focus should be placed back on the element focused before the dialog was opened after it has been closed.

## 💻 Repro or Code Sample

Can be reproduced in any of the FluentDialog examples -> https://www.fluentui-blazor.net/Dialog

## 💁 Possible Solution

Register an onkeydown listener on both the first menu item (if one exists) and the menu button. If focus is on the menu button, tab will place focus on the first item in the menu. If focus is on the first menu item, a shift-tab will place focus back on the button.

Save a reference to the element currently focused during dialog open, in the dialog service. Focus it again during close.

I will follow up with a fix tomorrow.

## 🔦 Context

https://github.com/dotnet/aspire/issues/10119

adamint:
# Pull Request

## 📖 Description

Exposes two functions to get the focused element and set focus in `dialog-utils.js`. Uses that in the dialog provider to get the focused element before the dialog is created (which messes with focus), and then to reset focus after close. Dialog doesn't always close immediately, so I added a small delay.

Before:

https://github.com/user-attachments/assets/d8edbe5c-a111-45bc-bb19-bf6c076a1234

After:

https://github.com/user-attachments/assets/ac758bea-aafc-4a28-b648-1dd7ddf3d7a8

### 🎫 Issues

Fixes #4094

## 👩‍💻 Reviewer Notes

You can test this using the Dialog samples.

### General

### Component-specific

### DIFF
--- a/src/Core/Components/Dialog/Services/DialogInstance.cs
+++ b/src/Core/Components/Dialog/Services/DialogInstance.cs
@@ -8,7 +8,7 @@ namespace Microsoft.FluentUI.AspNetCore.Components;
 
 public sealed class DialogInstance
 {
-    public DialogInstance(Type? type, DialogParameters parameters, object content, IJSObjectReference previouslyFocusedElement)
+    public DialogInstance(Type? type, DialogParameters parameters, object content, IJSObjectReference? previouslyFocusedElement)
     {
         ContentType = type;
         Parameters = parameters;
@@ -25,7 +25,7 @@ public sealed class DialogInstance
 
     public DialogParameters Parameters { get; internal set; }
 
-    internal IJSObjectReference PreviouslyFocusedElement { get; }
+    internal IJSObjectReference? PreviouslyFocusedElement { get; }
 
     internal Dictionary<string, object>? GetParameterDictionary()
     {


src/Core/wwwroot/js/dialog-utils.js
```diff
@@ -0,0 +1,7 @@
+export function getActiveElement() {
```
dvoituron: We cannot accept JavaScript code in wwwroot files. You must move your code to a `FluentDialogProvider.razor.js` file. You will find an example in the `FluentButton` button.

src/Core/Components/Dialog/FluentDialogProvider.razor.cs
```diff
@@ -40,24 +50,47 @@ protected override void OnInitialized()
...
+        return await Task.Run(async () =>
+        {
+            var previouslyFocusedElement = await _module.InvokeAsync<IJSObjectReference>("getActiveElement");
+
+            DialogInstance dialog = new(dialogComponent, parameters, content, previouslyFocusedElement);
             dialogReference.Instance = dialog;
 
             _internalDialogContext.References.Add(dialogReference);
-            InvokeAsync(StateHasChanged);
+            await InvokeAsync(StateHasChanged);
```
dvoituron: Why did you changed this line?

src/Core/Components/Dialog/Services/DialogInstance.cs
```diff
@@ -20,6 +23,8 @@ public DialogInstance(Type? type, DialogParameters parameters, object content)
 
     public object Content { get; internal set; } = default!;
 
+    public IJSObjectReference PreviouslyFocusedElement { get; }
```
dvoituron: Can you change this property to `internal`?

src/Core/Components/Dialog/FluentDialogProvider.razor.cs
```diff
@@ -4,20 +4,30 @@
...
     private IDialogService DialogService { get; set; } = default!;
 
     [Inject]
     private NavigationManager NavigationManager { get; set; } = default!;
 
+    [Inject]
+    private IJSRuntime JSRuntime { get; set; } = default!;
+
+    [Inject]
+    private LibraryConfiguration LibraryConfiguration { get; set; } = default!;
```
dvoituron, line 30: Why did you add this property?

src/Core/wwwroot/js/dialog-utils.js
```diff
@@ -0,0 +1,7 @@
...
+    return document.activeElement
+}
+
+export function focusElement(element) {
```
dvoituron: Yo avoid side effect in some situations, can you test if the element is defined?
```js
if (element) {
  element.focus();
}
```

dvoituron (changes requested):

src/Core/wwwroot/js/dialog-utils.js
adamint: I thought scoped JS file import wasn’t supported outside of a blazor component? Since unlike FluentButton, the provider isn’t a component, I didn’t see another place to put this file.

or do I misunderstand?

adamint: edit: I was misunderstood, apologies! It is a component 😄

MarvinKlein1508:
Just because I'm curious: what happens if I open the dialog via code somehow? Which element will be focused then?

For example in my case I open a modal directly in `OnParameterSetAsync` when a certain condition is invalid.

vnbaaij:
I would say nothing happens then. I think this is done when the code from @dvoituron 's last review comment gets added

src/Core/Components/Dialog/FluentDialogProvider.razor.cs
adamint: Reverted

src/Core/Components/Dialog/Services/DialogInstance.cs
adamint: Done!

src/Core/Components/Dialog/FluentDialogProvider.razor.cs
adamint: It's used in the module import

src/Core/wwwroot/js/dialog-utils.js
adamint: Done!

adamint:
> Just because I'm curious: what happens if I open the dialog via code somehow? Which element will be focused then?
> 
> For example in my case I open a modal directly in `OnParameterSetAsync` when a certain condition is invalid.

Whatever element that is currently focused at the time that `OnParameterSetAsync` is called. In my opinion, focus should always return back to the previous element so that users of screen readers can continue to navigate the page in order.

src/Core/Components/Dialog/FluentDialogProvider.razor.cs
```diff
@@ -40,20 +51,43 @@ protected override void OnInitialized()
...
-        return await Task.Run(() =>
+        if (_module is null)
         {
-            DialogInstance dialog = new(dialogComponent, parameters, content);
+            throw new InvalidOperationException("JS module is not loaded.");
+        }
+
+        return await Task.Run(async () =>
+        {
+            var previouslyFocusedElement = await _module.InvokeAsync<IJSObjectReference>("getActiveElement");
```
dvoituron, line 88: You will find in the documentation: "If the focused element is within a document tree that's not descended from the current document (for example, the focused element is in the main document, and the invoking document is an embedded iframe), then this will be null."
https://developer.mozilla.org/en-US/docs/Web/API/Document/activeElement

So, previouslyFocusedElement can be null and you need to manage this case in `DialogInstance` and all methods called.

src/Core/Components/Dialog/Services/DialogInstance.cs
```diff
@@ -22,6 +25,8 @@ public DialogInstance(Type? type, DialogParameters parameters, object content)
 
     public DialogParameters Parameters { get; internal set; }
 
+    internal IJSObjectReference PreviouslyFocusedElement { get; }
```
dvoituron: This property must be nullable. See my [other comment](https://github.com/microsoft/fluentui-blazor/pull/4095/files#r2311858864)

dvoituron:
This new feature should be added to the “dev-v5” branch for the new Dialog component.

src/Core/Components/Dialog/Services/DialogInstance.cs
vnbaaij: Changed by me

src/Core/Components/Dialog/FluentDialogProvider.razor.cs
vnbaaij: I've changed this in DialogInstance

dvoituron (approved):

vnbaaij:

PR #4100, opened 2025-08-31
[dev-v5] Add DatePicker

dvoituron:
# [dev-v5] Add DatePicker

## DatePicker
Picking a date can be tough without context. A date picker (DatePicker) offers a popup control that's optimized for picking a single date from a calendar view where contextual information like a disabled day or fullness of the calendar is important.

You can use the Culture parameter to display the month and day names in different languages; and to set the date format displayed in the input field.

## Views
The DatePicker can be displayed in different views: day, month, and year.
The user can switch between these views using the title item: the month name or the year number.

```xml
<FluentDatePicker Label="Days view" bind-Value="@SelectedValue" />
```

image

## RenderingMode
The DatePicker can be rendered in different modes: FluentUI, Browser.

The FluentUI rendering mode uses the Fluent UI styles and components. This mode provides a consistent look and feel with other Fluent UI components. This mode provides all the features of the component.

The Browser rendering mode uses the default browser styles. This mode is useful when you want to use a very simple DatePicker with a mobile device. In this case, the mobile picker will be used. This could be useful to use the native Android or iOS date picker.

> Warning
> 
> This mode is very limited in features and does not support the UI customization (DisabledDateFunc, DaysTemplate, ...). The 
> following parameters are ignored: Culture, DayFormat, DisabledDateFunc, DisabledCheckAllDaysOfMonthYear, 
> DisabledSelectable, DaysTemplate, PickerMonthChanged.

image

## Unit Tests
{B1B3000F-113B-4367-A15D-D791B7364D45}

### DIFF
--- a/src/Core/Enums/DatePickerRenderingMode.cs
+++ b/src/Core/Enums/DatePickerRenderingMode.cs
@@ -1,0 +1,15 @@
+// ------------------------------------------------------------------------
+// This file is licensed to you under the MIT License.
+// ------------------------------------------------------------------------
+
+namespace Microsoft.FluentUI.AspNetCore.Components;
+
+/// <summary />
+public enum DatePickerRenderingMode
+{
+    /// <summary />
+    FluentUI,
+
+    /// <summary />
+    Browser,
+}
--- a/Microsoft.FluentUI-v5.sln
+++ b/Microsoft.FluentUI-v5.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 17
-VisualStudioVersion = 17.11.35118.90
+# Visual Studio Version 18
+VisualStudioVersion = 18.0.10906.23 main
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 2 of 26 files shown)


src/Core/Enums/DatePickerRenderingMode.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace Microsoft.FluentUI.AspNetCore.Components;
+
+/// <summary />
+public enum DatePickerRenderingMode
+{
+    /// <summary />
+    FluentUI,
+
+    /// <summary />
+    Browser,
```
vnbaaij: I think we should use `Native` instead of `Browser` (better suited when, for example, using the components in Blazor Hybrid.

Also, a bit in doubt about `...RenderingMode` Could be confused with InteractiveServer/Wasm/Auto. Maybe change to `RenderStyle`

vnbaaij (changes requested):

src/Core/Enums/DatePickerRenderingMode.cs
dvoituron: I rename the parameter to `RenderStyle` and the enum value to `Native`

vnbaaij (approved):

PR #4123, opened 2025-09-09
[Dialog] Return focus to element when dialog is closed

vnbaaij:
This is an implementation of #4095 for V5. It works by capturing a reference to the focused element when `FluentDialog.ShowAsync()` is called and setting back the focus to that element at the end of the `DialogService.CloseAsync()` call.

Displaying the dialog is done in a method that has `[ExcludeFromCodeCoverage]` set so I don't think there are tests that need to be added

### DIFF
--- a/src/Core/Components/Dialog/FluentDialog.razor.cs
+++ b/src/Core/Components/Dialog/FluentDialog.razor.cs
@@ -12,7 +12,7 @@ namespace Microsoft.FluentUI.AspNetCore.Components;
 
 /// <summary>
 /// The dialog component is a window overlaid on either the primary window or another dialog window.
-/// Windows under a modal dialog are inert. 
+/// Windows under a modal dialog are inert.
 /// </summary>
 public partial class FluentDialog : FluentComponentBase
 {
@@ -76,11 +76,11 @@ public partial class FluentDialog : FluentComponentBase
     public EventCallback<DialogEventArgs> OnStateChange { get; set; }
 
     /// <summary>
-    /// 
+    ///
     /// </summary>
     /// <param name="firstRender"></param>
     /// <returns></returns>
-    protected override Task OnAfterRenderAsync(bool firstRender)
+    protected override async Task OnAfterRenderAsync(bool firstRender)
     {
         if (firstRender && LaunchedFromService)
         {
@@ -90,10 +90,15 @@ public partial class FluentDialog : FluentComponentBase
                 instance.FluentDialog = this;
             }
 
-            return ShowAsync();
+            var pfe = await ShowAsync();
+
+            if (instance is not null)
+            {
+                instance.PreviouslyFocusedElement = pfe;
+            }
         }
 
-        return Task.CompletedTask;
+        return;
     }
 
     /// <summary />
@@ -145,9 +150,9 @@ public partial class FluentDialog : FluentComponentBase
     /// Displays the dialog.
     /// </summary>
     [ExcludeFromCodeCoverage]
-    public async Task ShowAsync()
+    public async Task<IJSObjectReference> ShowAsync()
     {
-        await JSRuntime.InvokeVoidAsync("Microsoft.FluentUI.Blazor.Components.Dialog.Show", Id);
+        return await JSRuntime.InvokeAsync<IJSObjectReference>("Microsoft.FluentUI.Blazor.Components.Dialog.Show", Id);
     }
 
     /// <summary>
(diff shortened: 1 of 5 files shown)


vnbaaij:

src/Core/Components/Dialog/Services/IDialogInstance.cs
```diff
@@ -14,6 +16,11 @@ public interface IDialogInstance
     /// </summary>
     internal Type ComponentType { get; }
 
+    /// <summary>
+    /// Holds a reference to the previously focused element (before the dialog was opened).
+    /// </summary>
+    IJSObjectReference? PreviouslyFocusedElement { get; set; }
```
dvoituron: I think this property should be `internal`.
I don't think it's a good idea to return an `IJSObjectReference` to the developer.

src/Core/Components/Dialog/Services/DialogService.cs
```diff
@@ -3,6 +3,7 @@
 // ------------------------------------------------------------------------
 
 using System.Diagnostics.CodeAnalysis;
+
```
dvoituron: Can you remove this empty line?

src/Core/Components/Dialog/Services/DialogService.cs
```diff
@@ -47,6 +48,9 @@ public async Task CloseAsync(IDialogInstance dialog, DialogResult result)
 
         // Raise the DialogState.Closed event
         dialogInstance?.FluentDialog?.RaiseOnStateChangeAsync(dialog, DialogState.Closed);
+
+        //Focus the previously focused element (if any)
```
dvoituron: Small detail :-)
Can you add a space before "Focus"

src/Core/Components/Dialog/Services/DialogInstance.cs
```diff
@@ -40,6 +41,9 @@ internal DialogInstance(IDialogService dialogService, Type componentType, Dialog
     /// <inheritdoc cref="IDialogInstance.Result"/>
     public Task<DialogResult> Result => ResultCompletion.Task;
 
+    /// <inheritdoc cref="IDialogInstance.PreviouslyFocusedElement"/>"
+    public IJSObjectReference? PreviouslyFocusedElement { get; set; }
```
dvoituron: With my previous comment, this property will be `internal`

src/Core/Components/Dialog/FluentDialog.razor.cs
```diff
@@ -90,10 +90,15 @@ protected override Task OnAfterRenderAsync(bool firstRender)
                 instance.FluentDialog = this;
             }
 
-            return ShowAsync();
+            var pfe = await ShowAsync();
```
dvoituron: A explicit name like `focusedElement` or `previouslyFocusedElement` should be better.

src/Core/Components/Dialog/FluentDialog.razor.cs
```diff
@@ -90,10 +90,15 @@ protected override Task OnAfterRenderAsync(bool firstRender)
...
+
+            if (instance is not null)
```
dvoituron: The result of `ShowAsync` may be null. You must test it.

src/Core/Components/Dialog/FluentDialog.razor.cs
```diff
@@ -159,6 +164,20 @@ public async Task HideAsync()
         await JSRuntime.InvokeVoidAsync("Microsoft.FluentUI.Blazor.Components.Dialog.Hide", Id);
     }
 
+    /// <summary>
+    /// Set the focus back to the element that had focus before the dialog was opened.
+    /// </summary>
+    /// <returns></returns>
+    [ExcludeFromCodeCoverage]
+    public async Task FocusPreviousElementAsync()
```
dvoituron: This method should be “private” or “internal.”

src/Core/Components/Dialog/FluentDialog.razor.cs
```diff
@@ -145,9 +150,9 @@ private async Task<DialogEventArgs> RaiseOnStateChangeAsync(DialogEventArgs args
     /// Displays the dialog.
     /// </summary>
     [ExcludeFromCodeCoverage]
-    public async Task ShowAsync()
+    public async Task<IJSObjectReference> ShowAsync()
```
dvoituron: I don't think returning an IJSObjectReference is a good idea. This reference could be misused for other purposes.

Would it be possible to store this reference in a private variable in FluentDialog and then use it in `FocusPreviousElementAsync`?
```csharp
_previousElement = await JSRuntime.InvokeAsync<IJSObjectReference> ...
```

dvoituron (changes requested):

src/Core/Components/Dialog/FluentDialog.razor.cs
vnbaaij: The previously focused element needs to be retrieved by JS (document.activeElement). I can separate it out into its own javascript function, but then we need to do 2 interop calls to achieve the same. Seems that would introduce more overhead?

vnbaaij:

vnbaaij:

vnbaaij:

PR #4129, opened 2025-09-11
[InputFile] Remove .ToArray as it is not necessary

MarvinKlein1508:
# Pull Request

## 📖 Description

This PR removes the `ToArray` call for the `OnCompleted` event callback which is not necessary. This saves a few allocations for the copied array.

### General

### Component-specific

### DIFF
--- a/src/Core/Components/InputFile/FluentInputFile.razor.cs
+++ b/src/Core/Components/InputFile/FluentInputFile.razor.cs
@@ -361,7 +361,7 @@ public partial class FluentInputFile : FluentComponentBase, IAsyncDisposable
 
         if (OnCompleted.HasDelegate)
         {
-            await OnCompleted.InvokeAsync(uploadedFiles.ToArray());
+            await OnCompleted.InvokeAsync(uploadedFiles);
         }
     }
 


src/Core/Components/InputFile/FluentInputFile.razor.cs
Tyme-Bleyaert, line 1: Is it not because we want to send a decoupled snapshot of the list?

MarvinKlein1508: This could be a reason. But I don't believe that nobody is going to manually cast the `IEnumerable<T>` into a `List<T>` and appends or remove items to it. I cannot think of any good reason why someone even should try to do this.

But this is up for the maintainers to decide. I just noticed it while debugging on another issue! :)

dvoituron: Actually, I think it's better to keep the `ToArray` to ensure this decoupling. 
It has small impact on performance and avoids potential problems that a user might encounter when manipulating this list in OnCompleted`` . Although I can't immediately think of a practical example of a problem :-)

dvoituron: Why change something that works? 🤪

vnbaaij: Closing this one on account of "if it ain't broke, dont fix it'...

PR #4130, opened 2025-09-12
[General] Workflow maintenance

vnbaaij:
- Remove deploy step from build/test workflow. This should prevent the error we get after every merge (finally!)
- Sync demo/preview deploy files. Use same comments and steps


vnbaaij:

vnbaaij:

PR #4135, opened 2025-09-13
[ListBaseComponent] Fix for default selection options

### LINKED ISSUE #4128: fix: always first item selected in ` FluentAutocomplete`, when using `*Template`s
# 🐛 Bug Report

I'm using `FluentAutocomplete` with `OptionTemplate` and `SelectedOptionTemplate` both wrapping a `FluentPersona`. When clicking on an item from the popup, it always selects the first one, even if I scroll way down. When removing the templates, I can select the correct item.

## 💻 Repro or Code Sample

```csharp
<FluentAutocomplete Id="add-employee"
                    TOption="EmployeeDto"
                    AutoComplete="on"
                    Placeholder="-- Kennung auswählen --"
                    MaximumOptionsSearch="@_availableEmployees.Count()"
                    Multiple="false"
                    SelectValueOnTab="true"
                    ShowOverlayOnEmptyResults="false"
                    Virtualize="true"
                    @bind-SelectedOption="@_selectedEmployee"
                    OnOptionsSearch="@OnSearchTermChanged">
                    @* OptionText="@(item => $"{item.Nik}: {item.FullName}")"> *@
    <OptionTemplate Context="ctx">
        <FluentPersona ImageSize="25px"
                       Initials="@($"{ctx.Name?[..1]}{ctx.Lastname?[..1]}")"
                       Name="@($"{ctx.Nik}: {ctx.FullName}")" />
    </OptionTemplate>
    <SelectedOptionTemplate Context="ctx">
        <FluentPersona ImageSize="25px"
                       Initials="@($"{ctx.Name?[..1]}{ctx.Lastname?[..1]}")"
                       Name="@($"{ctx.Nik}: {ctx.FullName}")" />
    </SelectedOptionTemplate>
</FluentAutocomplete>
```

## 🤔 Expected Behavior

I excpet that the item I click on is selected.

## 😯 Current Behavior

When clicking on an item from the popup, it always selects the first one, even if I scroll way down. When removing the templates, I can select the correct item.

## 💁 Possible Solution

None.

## 🔦 Context

A workaround is to use use the `Multiple="true"` but limit it to `MaximumSelectedOptions="1"` and `OptionText="@(item => $"{item.Nik}: {item.FullName}")`.

## 🌍 Your Environment

* OS & Device: Windows 11
* Browser Microsoft Edge Version 140.0.3485.54 (Official build) (64-bit)
* .NET 8 and Fluent UI Blazor library Version 4.12.1

Tyme-Bleyaert:
# Pull Request

## 📖 Description

Replace `item?.Tostring()` to determine current selected value as this would result in `NamespaceName.MyClass` for most cases.
Now uses a hash key as the default and not the `GetHashCode` for security.

### 🎫 Issues

Fix #4128 

### General

### Component-specific


dvoituron:
Give me a few days to verify these changes. One initial note: You cannot change the unit tests. If that is required, it is a breaking change.

Tyme-Bleyaert:
> Give me a few days to verify these changes. One initial note: You cannot change the unit tests. If that is required, it is a breaking change.

No rush at all 😊
You’re right that changing tests typically implies a breaking change, but in this case, most of the tests I updated were either checking `ToString()` outputs or performing `*.verified.razor.html` comparisons with hardcoded option values. So while it could be a breaking change, I’d expect it to only affect very specific edge cases.

I’m also happy to discuss this further over a Teams call if that would help.

Tyme-Bleyaert:
Just realized something, when working with a list of strings the reference will be different while the value might be the same.
I need to address that issue first. So i marked the pr as draft again.

Tyme-Bleyaert:
Ok, that’s better. I was making it way harder than the issue actually was. 😐
Why write many lines when deleting one does the trick.

dvoituron (approved):

PR #4141, opened 2025-09-17
[General] Re-add net10.0 TFM and update to latest SDKs/NuGet packages

vnbaaij:
With earlier versions we had the issue that the esproj file contents would get served instead of the generated javascript (resulting in a non-functional site). This has now been fixed in the SDKs for both NET 9 and NET 10.
So, I've re-added the net10.0 TFM to the project files and configured the solution (and GH workflows/ADO pipelines) to use the latest available SDKs and NuGet packages.

Also updates the NET 10.0 Web App template source


vnbaaij:

vnbaaij:

vnbaaij:

vnbaaij:

dvoituron (approved):

Issue #4144, opened 2025-09-17
fix: KeyNotFoundException in FluentAppBar


JamesNK:
# 🐛 Bug Report

I sometimes see this error in Aspire when resizing the browser window:

<img width="1085" height="521" alt="Image" src="https://github.com/user-attachments/assets/22367483-6c7a-469e-b548-6ba9b8fe33d7" />

## 🤔 Expected Behavior

No error

## 😯 Current Behavior

Error

## 💁 Possible Solution

Use TryGetValue here:

https://github.com/microsoft/fluentui-blazor/blob/c9fe1ce04c28e0f3c1793efa69e98d864e44c51c/src/Core/Components/AppBar/FluentAppBar.razor.cs#L131

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* .NET and Fluent UI Blazor library Version [e.g. 8.0.2 and 4.4.1]

JamesNK:
(I would assign this to copilot to fix but I don't have permission in this repo)

MarvinKlein1508:
@JamesNK I assigned Copilot but I don't now how I can trigger it to start working.

vnbaaij:
@copilot implement the mentioned trygetvalue

dvoituron:
Hi. My name is not Copilot but I pushed this PR #4146 :-)

Tyme-Bleyaert:
> Hi. My name is not Copilot but I pushed this PR [#4146](https://github.com/microsoft/fluentui-blazor/pull/4146) :-)

Sound like something copilot would say 😄 .
Also would `ContainsKey` and `TryGetValue` not be a double key lookup?

dvoituron:
Yes, it's a double check. I started with the first one and tested the second one afterwards :-)

JamesNK:
TryGetValue does a contains internally. You don't need both.

This is fine:
```cs
            if (item.Id is not null && _internalAppBarContext.Apps.TryGetValue(item.Id, out var app))
            {
                app.Overflow = item.Overflow;
            }
```

PR #4146, opened 2025-09-18
[AppBar] Add TryGetValue and Apps.ContainsKey

dvoituron:
# [AppBar] Add TryGetValue and Apps.ContainsKey

Add a double test: Apps.ContainsKey and TryGetValue to verify than the key is in memory.

See #4144

# Unit Tests

no changes

### DIFF
--- a/src/Core/Components/AppBar/FluentAppBar.razor.cs
+++ b/src/Core/Components/AppBar/FluentAppBar.razor.cs
@@ -126,12 +126,9 @@ public partial class FluentAppBar : FluentComponentBase
 
         foreach (var item in items)
         {
-            if (item.Id is not null && _internalAppBarContext.Apps.ContainsKey(item.Id))
+            if (item.Id is not null && _internalAppBarContext.Apps.TryGetValue(item.Id, out var app))
             {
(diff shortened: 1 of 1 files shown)


src/Core/Components/AppBar/FluentAppBar.razor.cs
```diff
@@ -126,10 +126,12 @@ public async Task OverflowRaisedAsync(string value)
 
         foreach (var item in items)
         {
-            if (item.Id is not null)
+            if (item.Id is not null && _internalAppBarContext.Apps.ContainsKey(item.Id))
```
JamesNK: TryGetValue does a contains internally. You don't need both.

This is fine:
```cs
            if (item.Id is not null && _internalAppBarContext.Apps.TryGetValue(item.Id, out var app))
            {
                app.Overflow = item.Overflow;
            }
```

dvoituron: Yes, I know. I started with a test and forgot to delete the double check afterwards.
But you're not the first to point this out to me, and I can see that it's bothering everyone 😀
So I'll correct it to limit the debate.

PR #4179, opened 2025-09-30
Icons] Allow GetInstance to find icons regardless of casing

### LINKED ISSUE #4176: fix: Case sensitive icons in IconsExtensions.GetInstance
# 🐛 Bug Report

There are helper methods to get icon details using its name. It is case sensitive. Do different FluentUI icons ever differ only by case? Could the search be made case insensitive?

Scenario:
Someone using Aspire looked up a FluentUI icon name from the internet, `brainSparkle`, and tried to use it but it failed.

Image

FluentUI required `BrainSparkle` (capitalize B).

## 🤔 Expected Behavior

Can get icon instances without exact case.

## 😯 Current Behavior

Requires exact case.

## 💁 Possible Solution

Change this line to do a case insensitive compare:

https://github.com/microsoft/fluentui-blazor/blob/4d6024ebfe0d1ee7db5ada7c355985578f4f3487/src/Core/Components/Icons/IconsExtensions.cs#L41

## 🌍 Your Environment

vnbaaij:
As per the title: Use InvariantCultureIgnoreCase when finding icon based on full name.

All tests pass. Fix #4176

### DIFF
--- a/src/Core/Components/Icons/IconsExtensions.cs
+++ b/src/Core/Components/Icons/IconsExtensions.cs
@@ -38,7 +38,7 @@ public static partial class IconsExtensions
 
             // Ex. Microsoft.FluentUI.AspNetCore.Components.Icons.Filled.Size10+PresenceAvailable
             var iconFullName = $"{Namespace}.Icons.{icon.Variant}.Size{(int)icon.Size}+{icon.Name}";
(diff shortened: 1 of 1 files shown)


dvoituron (approved):

dvoituron:
@vnbaaij This change must be also applied to the dev-v5

Issue #4193, opened 2025-09-30
[dev-v5] fix: LayoutArea.Footer does not grow on smaller sizes


MarvinKlein1508:
# 🐛 Bug Report

The Footer within the demos does not grow on smaller screen sizes. This is also happening in v4 but less noticable.

<img width="468" height="73" alt="Image" src="https://github.com/user-attachments/assets/c70a4106-7caf-41e4-93d1-8ffd2e8329c3" />

It would be great if the v5 demos would also include a demo for a automatically growing footer area.

## 💻 Repro or Code Sample
https://fluentui-blazor-v5.azurewebsites.net/

## 🤔 Expected Behavior

I expect the footer to grow naturally.

## 😯 Current Behavior

Footer has a fixed size.

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* .NET and Fluent UI Blazor library Version [e.g. 8.0.2 and 4.4.1]

dvoituron:
Fixed #4195

PR #4195, opened 2025-09-30
[dev-v5] Layout with responsive foot

dvoituron:
# [dev-v5] Layout with responsive footer

Add `hidden-when="xs"` on extra content in the footer.

https://github.com/user-attachments/assets/466563d3-ce5c-4a79-984e-91f38dd93496

See #4193

### DIFF
--- a/examples/Demo/FluentUI.Demo.Client/Layout/DemoMainLayout.razor
+++ b/examples/Demo/FluentUI.Demo.Client/Layout/DemoMainLayout.razor
@@ -95,15 +95,19 @@ else
                     HorizontalGap="12px"
                     Style="font-size: smaller;">
 
-        Version: @AppVersion.Version
+        <div>
+			Version: @AppVersion.Version
+		</div>
 
         <FluentSpacer />
-        <a href="https://dotnet.microsoft.com/learn/aspnet/what-is-aspnet-core" target="_blank">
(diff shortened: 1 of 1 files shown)


MarvinKlein1508:
Simple fix. I still have a question regarding the component. Is there any easy way to get the footer to auto-size in this configuration:
grafik

This is a pretty common design setup for websites.

vnbaaij (approved):

dvoituron:
> I still have a question regarding the component. Is there any easy way to get the footer to auto-size in this configuration:

It is already “automatically resized”: but only horizontally. I assume you are referring to vertical auto-size? No, this is not possible because browsers have difficulty managing heights.
To be compatible with as many browsers as possible, you need to specify the height of the four elements. But there's nothing stopping the developer from using CSS functions such as “calc” to automatically calculate the height of the footer based on the other values in its design. In my opinion, headers and footers generally have fixed sizes on professional websites.

MarvinKlein1508:
Yes I mean vertical auto-size. Can't I just set `--layout-footer-height` to `auto` and remove it's value from the calc within `--layout-body-height`.

I tried to paint it together what I mean:
grafik

dvoituron:
Yes. This is an HTML problem. You need to specify a fixed value for the `--layout-footer-height`.

PR #4204, opened 2025-10-02
Add virtualization view to the icon and emoji explorer, so you can endless scroll it

paule96:
# Pull Request

## 📖 Description

The main change of this Pullrequest for the user is, that he can use on the documentation sites of icons and emoji to search icons with endless scrolling functionality.
In the code behind that was a bit bigger change. The first big noticeable change is that the classes `IconsExtensions` and `EmojiExtensions` now use for the list of `Info` classes caching. Also the `IconsExtension` got a new method to filter emojies that are available for authors, that change things inside of FluentUI.

The second big change is, that the result list of the Emoji and Icons Explorer is now a shared component. It just provides the functionality to place multiple items in a row and virtualize the rows.

## 👩‍💻 Reviewer Notes

There is no issue for this as far as I know. I just stumbled over it, while I was working on a draft for the unrelated discussion of #4098.

Another thing to notice is performance. In particular, on the emoji side it costs a lot of CPU resources to load the visible emojis on 4k screens or bigger. The reason is that it now shows as many emojis as can fit into the view area. That are a lot more as it was before. (before only 24)

### General

### Component-specific


vnbaaij:
Hi,

Thanks for your input and work, but I don't think this is something we want to implement at this time. The reasons for this are:

- We do not want to expose the core icons to the developers. They should only use what is available in the Icons libraries
- We do not want to impose the high resource utilization on the emoji explorer users. Delivering it in a paged sense keeps the load on the client system bearable.
- The code for the `DynamicRowElementVirtualisationList` seems a bit verbose and not optimal. For example, why is the JS interop needed and why is that not being called in `OnAfterRenderAsync`
- We have not seen many (it any at all) issues being raised on the current explorer implementations. WE rather leave it as-is and concentrate our efforts on the V5 version.

We absolutely welcome contributions and totally do not want to scare you off by not pulling this in. For the next time, maybe first start a discussion to see if there is interest in a specific change addition and gather input on how to implement that way.

paule96:
The thing with the core icons can be fast removed. It was just something that helped me a lot what Icons are allowed and how do they look like. Because as far as I'm aware of and  reading around, their is currently no where a visualized view for these icons.

The performance issue is something that I totally understand. But it's something that the user already has. (just less obvious) but I agree that this is an area where I could invest a bit more time to fix it. From an HTML perspective that emojis should render way faster. So their is a potential to make it a lot faster. 


To the JsInterop, it is needed to figure out how much width a row has. I was not able to find a better solution then that. Their is sadly no API in blazor available that I'm aware of that allows C# to measure the real width of an element. That's why it's needed. But you are right the loading should be moved in the 'on after render'.


Hm I can not speak for others only about my team. But what happens way to often, you go to the fluentui blazor documentation side. Open for example the icons page you go to the explore and you sit their and question yourself why you only see 20 elements. Until you realize that you didn't scrolled long enough down to see the page navigation. Also this page navigation is really bad in getting an overview about what is possible. Because you are always interrupted by the navigation to the next page. The solution for my team was to use the icons page of the react implementation of fluent. This page has a perfomant endless scroll.

Why I didn't openend an issue upfront because I initially thinked it's a minor problem. It's only a small change in docs and that's it. That I needed in the end 2 days for it I couldn't know 😅 but if you are worried about the changes in the components project I can revert them all without affecting to much the icon/emoji explorer.

I totally understand if you wanna focus your work on V5. Don't feel pressured by that PR.

 

> P.S.: I'm not near a PC for the next 10 days. (or I plan to)

vnbaaij:
> The thing with the core icons can be fast removed. It was just something that helped me a lot what Icons are allowed and how do they look like. Because as far as I'm aware of and reading around, their is currently no where a visualized view for these icons.

The core icons are just copies we make from the icon library into the core library. We do so to not have a fixed dependency on the icon DLLs from the core package. These copied icons are not meant to be used by developers and are just for internal usage. As thy are just copies, there really is no value in making them searchable/browsable

> The performance issue is something that I totally understand. But it's something that the user already has. (just less obvious) but I agree that this is an area where I could invest a bit more time to fix it. From an HTML perspective that emojis should render way faster. So their is a potential to make it a lot faster.

I don't feel speed is an impediment for this functionality. Intended audience are developers who just need to be able to find the icon or emoji they need in their own application or site. Doing that at the highest possible speed is a non-optimization.
 
> To the JsInterop, it is needed to figure out how much width a row has. I was not able to find a better solution then that. Their is sadly no API in blazor available that I'm aware of that allows C# to measure the real width of an element. That's why it's needed. But you are right the loading should be moved in the 'on after render'.
> 
> Hm I can not speak for others only about my team. But what happens way to often, you go to the fluentui blazor documentation side. Open for example the icons page you go to the explore and you sit their and question yourself why you only see 20 elements. Until you realize that you didn't scrolled long enough down to see the page navigation. Also this page navigation is really bad in getting an overview about what is possible. Because you are always interrupted by the navigation to the next page. The solution for my team was to use the icons page of the react implementation of fluent. This page has a perfomant endless scroll.
> 

We don't have an unlimited Azure subscription to host our sites on. We are not an official team and do not have official budgets to run this. Providing unlimited scrolling will lead to higher bandwidth usage and therefore higher costs. We rather use the budget we have for something else.

> Why I didn't openend an issue upfront because I initially thinked it's a minor problem. It's only a small change in docs and that's it. That I needed in the end 2 days for it I couldn't know 😅 but if you are worried about the changes in the components project I can revert them all without affecting to much the icon/emoji explorer.
> 
> I totally understand if you wanna focus your work on V5. Don't feel pressured by that PR.
> 
> > P.S.: I'm not near a PC for the next 10 days. (or I plan to)

PR #4177, opened 2025-09-30
[DataGrid] Asynchronous IQueryable based loading and error handling UI feedback

miguelhasse:
# Pull Request

## 📖 Description

This feature introduces two visual enhancements to the DataGrid component while asynchronously loading items from an IQueryable data source:

1. The 'Loading' state flag is always set to true when requesting items on first time load (or while no items have yet been loaded) providing better visual feedback for users (avoiding empty item messages before loading).
2. Exceptions that occur when asynchronously requesting items can the caught, depending on the filter parameter 'HandleLoadingError', and displayed in the body of the DataGrid. The exception error rendering can be customized using 'ErrorContent' template.

## 👩‍💻 Reviewer Notes

A working sample of the 'Loading' state flag feature can be seen in the following samples:
- https://github.com/miguelhasse/fluentui-blazor/tree/adventureworks-server-sample/examples/AdventureWorks
- https://github.com/miguelhasse/fluentui-blazor/tree/adventureworks-server-sample/examples/TripPin

These same samples can be used to see a working sample for the error handling feature by forcing data source connection errors.

### General

- [ ] I have added tests for my changes.
- [x] I have tested my changes.
- [ ] I have updated the project documentation to reflect my changes.
- [x] I have read the [CONTRIBUTING](https://github.com/microsoft/fluentui-blazor/blob/master/docs/contributing.md) documentation and followed the [standards](https://www.fast.design/docs/community/code-of-conduct/#our-standards) for this project.

### Component-specific

- [ ] I have added a new component
- [ ] I have added [Unit Tests](https://github.com/Microsoft/fluentui-blazor/blob/master/unit-tests.md) for my new component
- [x] I have modified an existing component
- [x] I have validated the [Unit Tests](https://github.com/Microsoft/fluentui-blazor/blob/master/unit-tests.md) for an existing component

PR #4218, opened 2025-10-06
[dev-v5] Pull in v4 PRs (part 2)

vnbaaij:
Part 2 of syncing V4 PR's from the last couple of months into the dev-v5 branch

This part implements 
- #4172
- #4177
- #4178 
- Remove `NoTabbing` parameter from DataGrid as it had no implementation (also not on v4!)  
- Add CustomIcon, IconsExtensions and unit tests

As of 6-10-2025 we are in sync with v4 PRs again!


src/Core/Components/DataGrid/FluentDataGrid.razor
```diff
@@ -5,260 +5,313 @@
...
-                }
-                else
-                {
-                    @LoadingContent
-                }
-            </FluentDataGridCell>
-        </FluentDataGridRow>
-    }
+	[ExcludeFromCodeCoverage(Justification = "Circumvent EFCC shortcomings when using razor.")]
+	private void RenderActualError(RenderTreeBuilder __builder, Exception ex)
```
dvoituron, line 306: To avoid using `ExcludeFromCodeCoverage`, you should move the code to the .cs file using a code like this.
Can you do that in a new PR to increase the code coverage and to have a "best clean code" ? :-)

```csharp
private void RenderActualError(RenderTreeBuilder builder)
{
    if (ErrorContent is null)
    {
        builder.AddContent(0, "An error occurred while retrieving data.");
    }
    else
    {
        builder.AddContent(1, ErrorContent(_lastError));
    }
}
```

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -323,6 +320,27 @@ public FluentDataGrid(LibraryConfiguration configuration) : base(configuration)
     [Parameter]
     public RenderFragment? LoadingContent { get; set; }
 
+    /// <summary>
+    /// Gets or sets the callback that is invoked when the asynchronous loading state of items changes and <see cref="IAsyncQueryExecutor"/> is used.
+    /// </summary>
+    /// <remarks>The callback receives a <see langword="true"/> value when items start loading
+    /// and a <see langword="false"/> value when the loading process completes.</remarks>
+    [ExcludeFromCodeCoverage(Justification = "This method requires a db connection and is to complex to be tested with bUnit.")]
```
dvoituron, line 328: You can simulate a DB using a method including a `Delay`. In a new PR?

src/Core/Components/DataGrid/FluentDataGrid.razor
vnbaaij: Will do!

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
vnbaaij: Will try!

PR #4239, opened 2025-10-13
[dev-v5] Update the doc

dvoituron:
# [dev-v5] Update the doc

- If you are using the NuGet’s Central Package Management (CPM) feature, you probably need to set `ManagePackageVersionsCentrally=false`

- You can apply a rendering mode globally to all routes by adding a similar command in the `Routes.razor` file:
 `@rendermode @(new InteractiveServerRenderMode(prerender: false))`
 (here is an example: update this line according to your project type).

### DIFF
--- a/docs/using-latest-daily.md
+++ b/docs/using-latest-daily.md
@@ -3,7 +3,8 @@
 These instructions will get you set up with the latest build of **FluentUI-Blazor**.
 
 Each time a commit is pushed into the `main` ou `dev`, the **Core** package is published on a special NuGet repository.
-Install the latest [Visual Studio 2022 Preview version](https://visualstudio.microsoft.com/vs/preview/) for the tooling.
+Install the latest [Visual Studio 2022 Preview version](https://visualstudio.microsoft.com/vs/preview/)
+or [Visual Studio 2026 Insider edition](https://visualstudio.microsoft.com/insiders/) for the tooling.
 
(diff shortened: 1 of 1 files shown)


docs/using-latest-daily.md
```diff
@@ -6,7 +6,7 @@ Each time a commit is pushed into the `main` ou `dev`, the **Core** package is p
 Install the latest [Visual Studio 2022 Preview version](https://visualstudio.microsoft.com/vs/preview/) for the tooling.
```
vnbaaij, line 8: Or the VS 2026 Insiders editon

docs/using-latest-daily.md
```diff
@@ -23,6 +23,23 @@ or else in the global NuGet.config. See [configuring NuGet behavior](https://lea
...
+
+If you are using the NuGet’s Central Package Management (CPM) feature, you probably need to set `ManagePackageVersionsCentrally=false`
+
+Example (update the version number with the lastest values):
+
+```xml
+<PropertyGroup>
+  <ManagePackageVersionsCentrally>false</ManagePackageVersionsCentrally>
+</PropertyGroup>
+
```
vnbaaij, line 37: Why would you set that to `false`? You want to use the same version in all projects in solution, right?

docs/using-latest-daily.md
dvoituron: Updated

PR #4255, opened 2025-10-21
[Combobox] Fix presetting option

### LINKED ISSUE #4250: fix: Preset option in Combobox does not work/shows blank combobox
# 🐛 Bug Report

The initial value passed to the Combobox is not shown. The Combobox is empty on first render.

## 💻 Repro or Code Sample

On the demo website: https://www.fluentui-blazor.net/combobox
Pre-selected option example is empty.

## 🤔 Expected Behavior

The initial value of the Combobox should be shown.

## 😯 Current Behavior

The initial value is not shown.

## 🌍 Your Environment

* OS & Device: Windows on PC
* Browser: Microsoft Edge
* .NET and Fluent UI Blazor library Version: 9 and 4.13.0

vnbaaij:
Fix #4250 by extending the `GetOptionSelected` code

### DIFF
--- a/src/Core/Components/Menu/FluentMenu.razor.cs
+++ b/src/Core/Components/Menu/FluentMenu.razor.cs
@@ -240,14 +240,14 @@ public partial class FluentMenu : FluentComponentBase, IAsyncDisposable
                 }
             }
 
-            if (_jsModule is not null)
+            if (_jsModule is not null && _anchoredRegionModule is not null)
             {
                 await _jsModule.InvokeVoidAsync("initialize", Anchor, Id, Open, _anchoredRegionModule, _dotNetHelper);
             }
         }
         else
         {
-            if (_jsModule is not null && _reinitializeEventListeners)
+            if (_jsModule is not null && _anchoredRegionModule is not null && _reinitializeEventListeners)
             {
                 // If the menu was closed, remove its set event listeners. If it opened (ie if the menu starts out closed),
                 // we should set them now.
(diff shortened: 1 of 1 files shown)


src/Core/Components/List/ListComponentBase.razor.cs
```diff
@@ -483,14 +483,18 @@ protected virtual bool GetOptionSelected(TOption item)
...
-            else if (SelectedOption == null)
+            else if (SelectedOption == null && string.IsNullOrEmpty(Value))
             {
                 return false;
             }
             else if (OptionValue != null && SelectedOption != null)
             {
                 return GetOptionValue(item) == GetOptionValue(SelectedOption);
             }
+            else if (!string.IsNullOrEmpty(Value) || !string.IsNullOrEmpty(InternalValue))
```
Tyme-Bleyaert: Whats the use case of the extra null check : `!string.IsNullOrEmpty(InternalValue)`?
Or were you gonna do somthing like this: `return string.Equals(GetOptionValue(item), Value ?? InternalValue, StringComparison.Ordinal);`

vnbaaij: Oops...that was a leftover of trying to solve the other issue I mentioned. Will remove it

dvoituron:
This `ListComponentBase` class is used by many other components (including Autocomplete). Can you confirm that this will not impact other components? In version 5, we need to add additional unit tests with all these edge cases (null, empty, etc.).

We should also add a unit test in the current version 4 to verify this right away.

vnbaaij:
> This `ListComponentBase` class is used by many other components (including Autocomplete). Can you confirm that this will not impact other components? In version 5, we need to add additional unit tests with all these edge cases (null, empty, etc.).
> 

This change only touches the `GetOptionSelected` method and returns true in case the `Value` is equal to the `OptionValue` (invoked on the item)

dvoituron:
> This change only touches the `GetOptionSelected` method and returns true in case the `Value` is equal to the `OptionValue` (invoked on the item)

That's not the only change. You also have a test on `IsNullOrEmpty(Value)`. This could impact other components or cases. 

How can we be sure there won't be any regressions?

vnbaaij:
Value was not being considered in this method before this change. I don't see how that could lead to a regression.

vnbaaij:

vnbaaij:

PR #4256, opened 2025-10-21
[Menu] Fix dispose error

### LINKED ISSUE #4248: fix: Error on dispose in FluentMenu
# 🐛 Bug Report

I see this error when hot reloading:

```
fail: Aspire.Hosting.Dashboard.Microsoft.AspNetCore.Components.Server.Circuits.CircuitHost[111]
      Unhandled exception in circuit 'rBZiMW0JBoyiB3tlk6cZbCzTCjG4stihwH9QbT2x3k4'.
      System.ArgumentNullException: Value cannot be null. (Parameter 'jsObjectReference')
         at System.ArgumentNullException.Throw(String paramName)
         at System.ArgumentNullException.ThrowIfNull(Object argument, String paramName)
         at Microsoft.FluentUI.AspNetCore.Components.FluentMenu.DisposeAsync()
```

## 🤔 Expected Behavior

No error

## 😯 Current Behavior

Error

## 💁 Possible Solution

Check whether these module references are null before using them:

https://github.com/microsoft/fluentui-blazor/blob/948540e5308232dafeff16b33706790945c13211/src/Core/Components/Menu/FluentMenu.razor.cs#L400-L402

IMO the fields should be nullable so nullable checks catch this. Almost everywhere they're used in this type they're checked for null before use.

## 🌍 Your Environment

vnbaaij:
Fix #4248 
Extra null checks added

### DIFF
--- a/src/Core/Components/Menu/FluentMenu.razor.cs
+++ b/src/Core/Components/Menu/FluentMenu.razor.cs
@@ -240,14 +240,14 @@ public partial class FluentMenu : FluentComponentBase, IAsyncDisposable
                 }
             }
 
-            if (_jsModule is not null)
+            if (_jsModule is not null && _anchoredRegionModule is not null)
             {
                 await _jsModule.InvokeVoidAsync("initialize", Anchor, Id, Open, _anchoredRegionModule, _dotNetHelper);
             }
         }
         else
         {
-            if (_jsModule is not null && _reinitializeEventListeners)
+            if (_jsModule is not null && _anchoredRegionModule is not null && _reinitializeEventListeners)
             {
                 // If the menu was closed, remove its set event listeners. If it opened (ie if the menu starts out closed),
                 // we should set them now.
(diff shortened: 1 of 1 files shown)


MarvinKlein1508:
I will try this fix locally on my machine.

vnbaaij:
> I will try this fix locally on my machine.

Yes, please! Based on the error, this seems to me the logical thing to check. If not, we need to go back to the drawing board...

MarvinKlein1508:
@vnbaaij can you tell me how I can build a local nupgk file for testing? Setting
grafik

doesn't do the trick for me.

Tyme-Bleyaert (changes requested):
Since the error is `ObjectDisposedException` adding null checks wont help. We either need to to supply a `_disposed` bool or set the modules to null after disposing them.

MarvinKlein1508:
@vnbaaij I can confirm this doesn't fix the issue. I believe the main cause of this is that the `dotNetReference` there has been disposed already.

Tyme-Bleyaert:
Why did this get merged? I'm confused.

vnbaaij:
I had turned on auto merge...

vnbaaij:
> @vnbaaij can you tell me how I can build a local nupgk file for testing? Setting ...
> doesn't do the trick for me.

You can just right click the project in the Solution Explorer and choose 'Pack'. It will create a NuGet package in the 'artifacts' folder, using the version number set in the `\fluentui-blazor\Directory.Build.props` file

PR #4258, opened 2025-10-21
[Menu] More fixes for dispose error

vnbaaij:
More fixes for #4248:
- Make the `IJSObjectReference`s nullable
- Set them to null in `DisposeAsync`
- Also check `DotNetObjectReference` in `OnAfterRenderAsync`

I'll wait with merging this time till we are all on the same page :)

### DIFF
--- a/src/Core/Components/List/ListComponentBase.razor.cs
+++ b/src/Core/Components/List/ListComponentBase.razor.cs
@@ -483,13 +483,17 @@ public abstract partial class ListComponentBase<TOption> : FluentInputBase<strin
             {
                 return OptionSelected.Invoke(item);
             }
-            else if (SelectedOption == null)
+            else if (SelectedOption == null && string.IsNullOrEmpty(Value))
             {
                 return false;
             }
             else if (OptionValue != null && SelectedOption != null)
             {
                 return GetOptionValue(item) == GetOptionValue(SelectedOption);
+            }
+            else if (!string.IsNullOrEmpty(Value))
+            {
+                return string.Equals(GetOptionValue(item), Value, StringComparison.Ordinal);
             }
             else
             {


MarvinKlein1508:
@vnbaaij for me it is still not working. I still get a disposed error:
```
System.ObjectDisposedException: Cannot access a disposed object.
      Object name: 'Microsoft.JSInterop.DotNetObjectReference`1[[Microsoft.FluentUI.AspNetCore.Components.FluentMenu, Microsoft.FluentUI.AspNetCore.Components, Version=4.13.1.0, Culture=neutral, PublicKeyToken=null]]'.
         at Microsoft.JSInterop.JSRuntime.TrackObjectReference[TValue](DotNetObjectReference`1 dotNetObjectReference)
         at Microsoft.JSInterop.Infrastructure.DotNetObjectReferenceJsonConverter`1.Write(Utf8JsonWriter writer, DotNetObjectReference`1 value, JsonSerializerOptions options)
         at System.Text.Json.Serialization.JsonConverter`1.TryWrite(Utf8JsonWriter writer, T& value, JsonSerializerOptions options, WriteStack& state)
         at System.Text.Json.Serialization.JsonConverter`1.TryWrite(Utf8JsonWriter writer, T& value, JsonSerializerOptions options, WriteStack& state)
         at System.Text.Json.Serialization.JsonCollectionConverter`2.OnTryWrite(Utf8JsonWriter writer, TCollection value, JsonSerializerOptions options, WriteStack& state)
         at System.Text.Json.Serialization.JsonConverter`1.TryWrite(Utf8JsonWriter writer, T& value, JsonSerializerOptions options, WriteStack& state)
         at System.Text.Json.Serialization.JsonConverter`1.WriteCore(Utf8JsonWriter writer, T& value, JsonSerializerOptions options, WriteStack& state)
         at System.Text.Json.Serialization.Metadata.JsonTypeInfo`1.Serialize(Utf8JsonWriter writer, T& rootValue, Object rootValueBoxed)
         at System.Text.Json.JsonSerializer.WriteString[TValue](TValue& value, JsonTypeInfo`1 jsonTypeInfo)
         at Microsoft.JSInterop.JSRuntime.InvokeAsync[TValue](Int64 targetInstanceId, String identifier, CancellationToken cancellationToken, Object[] args)
         at Microsoft.JSInterop.JSRuntime.InvokeAsync[TValue](Int64 targetInstanceId, String identifier, Object[] args)
         at Microsoft.FluentUI.AspNetCore.Components.FluentMenu.OnAfterRenderAsync(Boolean firstRender) in E:\fluentui-blazor\src\Core\Components\Menu\FluentMenu.razor.cs:line 245
         at Microsoft.AspNetCore.Components.RenderTree.Renderer.GetErrorHandledTask(Task taskToHandle, ComponentState owningComponentState)
```

This can be reproduced with the steps from this: https://github.com/microsoft/fluentui-blazor/issues/4248#issuecomment-3426693456

Tyme-Bleyaert:
I'm gonna check it out this evening, might join this pr for some pair programming :)

MarvinKlein1508:
@vnbaaij I set up a try catch here to debug this and the root of this issue is that the dotNetReference has been disposed but is not null.
grafik

MarvinKlein1508:
@vnbaaij I found a working solution. In line 28 add:
```cs
private bool _disposed;
```

In line 244 change to:
```cs
if (_jsModule is not null && _anchoredRegionModule is not null && _dotNetHelper is not null && !_disposed)
```

as well in line 251 to:
```cs
if (_jsModule is not null && _anchoredRegionModule is not null && _dotNetHelper is not null && !_disposed && _reinitializeEventListeners)
```

And then update the dispose method to:
```cs
public async ValueTask DisposeAsync()
{
    if (_disposed)
    {
        return;
    }

    _disposed = true;
    _dotNetHelper?.Dispose();

    try
    {
        if (_jsModule is not null)
        {
            await _jsModule.InvokeVoidAsync("dispose", Anchor);
            await _jsModule.DisposeAsync();
            _jsModule = null;
        }

        if (_anchoredRegionModule is not null)
        {
            await _anchoredRegionModule.DisposeAsync();
            _anchoredRegionModule = null;
        }
    }
    catch (Exception ex) when (ex is JSDisconnectedException ||
                               ex is OperationCanceledException)
    {
        // The JSRuntime side may routinely be gone already if the reason we're disposing is that
        // the client disconnected. This is not an error.
    }
    finally
    {
        GC.SuppressFinalize(this);
    }
}
```

MarvinKlein1508:
At least my provided test code doesn't seem to crash when we are doing it this way.

vnbaaij:
Yeah, I think this solution is in line with what @Tyme-Bleyaert  meant.

MarvinKlein1508:
@vnbaaij we should probably move the disposed check within `OnAfterRenderAsync` right to the beginning. Otherwise the component will import the modules again. Line 221 to 223. This would also explain why dotNetReference is not null - even after I set it to null in DisposeAsync in a previous try.

We should probably do
```cs
if( _disposed)
{
return;
}
```

right in the beginning and remove it from the 2 if statements

MarvinKlein1508:
nevermind. I just tried it when it's on top. This gets the error right back to me.

vnbaaij:
I change the `IJSObjectReference`s to be non nullable again. Then in line 244 and 251 we only need to check for the `!_disposed` (besides the `_reinitializeEventListeners` in 251) . Makes it a lot simpler, I think.

src/Core/Components/Menu/FluentMenu.razor.cs
```diff
@@ -401,11 +412,13 @@ public async ValueTask DisposeAsync()
             {
                 await _jsModule.InvokeVoidAsync("dispose", Anchor);
                 await _jsModule.DisposeAsync();
+                _jsModule = null;
```
MarvinKlein1508: this needs to be removed since the field is no longer nullable

src/Core/Components/Menu/FluentMenu.razor.cs
```diff
@@ -401,11 +412,13 @@ public async ValueTask DisposeAsync()
...
             }
 
             if (_anchoredRegionModule is not null)
             {
                 await _anchoredRegionModule.DisposeAsync();
+                _anchoredRegionModule = null;
```
MarvinKlein1508: this needs to be removed since the field is no longer nullable

MarvinKlein1508:
Sorry wrong button. Should we also move 
```cs
```

into the constructor?

vnbaaij:
> Sorry wrong button. Should we also move
> 
> ```cs
> _dotNetHelper = DotNetObjectReference.Create(this);
> ```
> 
> into the constructor?

No, we always do that in `OnAfterRenderAsync`

MarvinKlein1508:
@vnbaaij you are right. I was thinking wrong here. It wouldn't make a difference anyways since the method is only run on firstRender.

MarvinKlein1508:
@vnbaaij 
We need to expand the checks for the other two calls as well
```
await _jsModule.InvokeVoidAsync("addEventLeftClick", Anchor, _dotNetHelper);
```

```
await _jsModule.InvokeVoidAsync("addEventRightClick", Anchor, _dotNetHelper);
```

I tried my demo now in WebAssembly and it seems like the disposing here is a bit faster which triggers the exception earlier. 
Otherwise it seems to be working.

MarvinKlein1508 (approved):

MarvinKlein1508:
I have upgraded my project to the latest v4 preview release and I can confirm that this seems to be working now! :) No more crashes for me :)

PR #4272, opened 2025-10-29
[Chore] Prepare for release 

vnbaaij:
Update homepage and whatsnew


MarvinKlein1508:
Just because I would like to know. Do you need to publish the new templates manually or is it also being handeled by the Azure pipeline?

vnbaaij:
Our project templates you mean? Those templates are just another NuGet package that gets built by the Azure DevOps pipeline. In total the pipeline produces 6 signed packages that we then manually upload to NuGet.

PR #4277, opened 2025-10-29
Add PublishDemoLocally script

dvoituron:
# Add PublishDemoLocally script

Add a Powershell script to publish locally the demo website (before Azure publication).

image


vnbaaij (changes requested):
dotnet serve can be run with the `-b` option to use brotli compression. Might be a good idea to use that here as we are particularly interested in that version of the script being generated.

dvoituron:
This is already done, using `--brotli`. What more could you want?

vnbaaij:
Oops, misread that line...the commands are at the end of the line indeed.

PR #4286, opened 2025-11-03
[SortableList] Handle JS errors in a better way

### LINKED ISSUE #4285: fix: Uncaught JSException in FluentSortableList when navigating away
# 🐛 Bug Report

In the FluentSortableList.OnAfterRenderAsync implementation, the invocation of the JS codebehind's `init` method which instantiates the `Sortable` is wrapped in a try/catch suppressing `JSDisconnectedException` or `OperationCancelledException`. The method's [comments](https://github.com/microsoft/fluentui-blazor/blob/c5a0da5447272574027a37a002099c3f66eda442/src/Core/Components/SortableList/FluentSortableList.razor.cs#L205) indicate:

> This exception is expected when the user navigates away from the page
> and the component is disposed. We can ignore it.

However, I see this exception thrown as a JSException which will go uncaught if not wrapped in an ErrorBoundary. 
Logged as `Microsoft.JSInterop.JSException: Sortable: `el` must be an HTMLElement, not [object Null]`.

## 💻 Repro or Code Sample

I can reproduce on the [Fluent Blazor demo site](https://www.fluentui-blazor.net/SortableList) by loading the sortable list and rapidly navigating to/from another component in the nav menu.

However, I think the site is wrapping these in an error boundary. The error can be seen in console

Image

## 🤔 Expected Behavior

`Microsoft.JSInterop.JSException` would also be caught suppressed when it can be validated it is this specific error which can occur when navigating away.

## 😯 Current Behavior

Exception will be thrown.

## 💁 Possible Solution

If we could also catch the JS Exception but re-throw if it's not this specific issue, that might work. However, it seems a little flimsy to validate this based on the error message alone. 

But our other option right now is to always remember to wrap the list in a custom error boundary that accounts for this scenario, which can be difficult to manage going forward (would always need to remember to wrap a FluentSortableList,  or use a custom component that does the wrapping). 

Any recommended ways to handle the scenario best would be appreciated. Thanks in advance.
 

## 🔦 Context

Using `FluentSortableList` in an SWA. 

## 🌍 Your Environment

* OS & Device: MacOS, Windows
* Browser Microsoft Edge, Google Chrome
* .NET 9, and Fluent UI Blazor 4.12.1

vnbaaij:
Fix #4285 by implementing `IAsynDisposable` and checking disposed state in `OnAfterRenderAsync`.

The error mentioned in the issue happened when moving **from** the SortableList page to anther page. In effect the JS function was being called when the element was already disposed of. This is now checked in a better way.

### DIFF
--- a/src/Core/Components/SortableList/FluentSortableList.razor.cs
+++ b/src/Core/Components/SortableList/FluentSortableList.razor.cs
@@ -27,8 +27,6 @@ public partial class FluentSortableList<TItem> : FluentComponentBase, IAsyncDisp
     /// <summary />
     private IJSObjectReference? Module { get; set; }
 
-
-
     /// <summary>
     /// Gets or sets the template to be used to define each sortable item in the list.
     /// Use the @context parameter to access the item and its properties.


src/Core/Components/SortableList/FluentSortableList.razor.cs
```diff
@@ -23,6 +24,11 @@ public partial class FluentSortableList<TItem> : FluentComponentBase
     [Inject]
     private IJSRuntime JSRuntime { get; set; } = default!;
 
+    /// <summary />
+    private IJSObjectReference? Module { get; set; }
+
+
```
dvoituron: Can you remove the multiple lines?

vnbaaij: Yup. Jus saw it

PR #4316, opened 2025-11-12
Update the 'IDialogService.CloseAsync' methods to accept an IDialogReference interface instead of the concrete DialogReference class.

StevenRasmussen:
# Pull Request

## 📖 Description

The methods on the `IDialogService` interface that create a reference to a dialog (ShowDialog, ShowPanel etc) all return an `IDialogReference` interface.  However, the `CloseAsync` method accepts a `DialogReference` concrete type instead of the interface.  It makes sense to me that the `CloseAsync` methods should accept the interface as well.

### General

### Component-specific


vnbaaij:
I'm not opposed to this, but can you elaborate on what we gain from it?

StevenRasmussen:
As mentioned in the description, when you use one of the methods to create a reference to a dialog (e.g. `ShowDialog` or `ShowPanel`) then you get back an `IDialogReference`.  If you want to close that dialog via the `CloseDialog` method, you now have to do some checks to ensure that you actually have an instance of the concrete type `DialogReference` and not just an object that implements the `IDialogReference` interface.  So your logic is at worst:

```csharp
var dialogReference = _dialogService.ShowDialog(...);

// Not type safe and could be false and cause an error
_dialogService.CloseDialog(dialogReference as DialogReference)
```

OR

```csharp
var dialogReference = _dialogService.ShowDialog(...);

if(dialogReference is DialogReference realDialogReference)
{
    _dialogService.CloseDialog(realDialogReference);
}
```

Either way... the methods should be consistent in what type they return and accept (unless of course there is a good reason for it which doesn't appear to be the case in this instance).  You could change the `ShowDialog` and `ShowPanel` methods to return the concrete type of `DialogReference` or change the `CloseDialog` methods to accept an `IDialogReference` as is being suggested by this PR :)

Issue #4311, opened 2025-11-11
fix: Error in FluentDialogProvider when displaying dialog on page load


JamesNK:
# 🐛 Bug Report

## 💻 Repro or Code Sample

`FluentDialogProvider` throws an error if its module isn't available: https://github.com/microsoft/fluentui-blazor/blob/b77b5300ef71b93d6064799bb1c4132eee0934c6/src/Core/Components/Dialog/FluentDialogProvider.razor.cs#L64-L67

The problem with this is trying to use `IDialogService` to show a dialog immediately on app load will get an error because the module hasn't finished loading. And there is no way for a caller to know if the module is available or not.

## 🤔 Expected Behavior

`IDialogService.ShowDialog` should be safe to call when a page first loads. If JavaScript is still loading, then the service waits until JS is finished loading.

## 😯 Current Behavior

Throws an error when showing a dialog on page load. There is no way to know whether it will throw the error.

## 💁 Possible Solution

Create a `TaskCompletionSource` that methods like ShowDialog await on instead of throwing an error. TCS is completed when this is complete (or an error is set if it throws):
https://github.com/microsoft/fluentui-blazor/blob/b77b5300ef71b93d6064799bb1c4132eee0934c6/src/Core/Components/Dialog/FluentDialogProvider.razor.cs#L58

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* .NET and Fluent UI Blazor library Version [e.g. 8.0.2 and 4.4.1]

JamesNK:
Regressed here: https://github.com/microsoft/fluentui-blazor/pull/4095

dvoituron:
FYI @adamint

adamint:
Please assign this to me, I will fix today

PR #4324, opened 2025-11-17
[Dialog] Fix error when module hasn't finished loading

JamesNK:
Fixes https://github.com/microsoft/fluentui-blazor/issues/4311


JamesNK:
The problem with the TCS approach is displaying a dialog in OnInitialize + await will block OnAfterRenderAsync from being called and will deadlock.

JamesNK:
Changed to not track previously selected element if module hasn't been loaded.

vnbaaij (approved):

PR #4333, opened 2025-11-18
[dev-v5][Chore] Update SDK, Packages and TFM

vnbaaij:
- Update .NET 10 SDK
- Update project TFMs
- Update NuGet Packages


dvoituron:
I don't think we need to do that now (except for `Directory.Packages.props`).

Adding multiple frameworks will slow down our development, as it will have to be compiled in multiple frameworks. We don't use any .NET9 features. 

Until we are in GA, I think it's best to keep a single framework (the lowest one). 
I also think that in the end, we will also need to compile in .NET8/9/10.

What do you think?

vnbaaij:
I've reverted the double TFMs. Back to 9 it is (except test projects which were already on 10)

Not sure if we should keep .NET 8 for v5. There are a couple of changes (like the UseStaticFiles/MapStaticFiles) which are harder to do if we support both 8 and 9/10.

PR #4335, opened 2025-11-18
[dev-v5][Layout] Rename Menu to Navigation

vnbaaij:
Use the more appropriate term 'navigation' (or 'nav') instead of 'menu'

### DIFF
--- a/src/Core/Enums/LayoutArea.cs
+++ b/src/Core/Enums/LayoutArea.cs
@@ -10,9 +10,11 @@ namespace Microsoft.FluentUI.AspNetCore.Components;
 public enum LayoutArea
 {
     /// <summary />
+    [Description("header")]
     Header,
 
     /// <summary />
+    [Description("footer")]
     Footer,
 
     /// <summary />
@@ -20,8 +22,10 @@ public enum LayoutArea
     Navigation,
 
     /// <summary />
+    [Description("content")]
     Content,
 
     /// <summary />
+    [Description("aside")]
     Aside,
 }


examples/Demo/FluentUI.Demo.Client/Documentation/Components/Layout/Examples/LayoutDefault.razor
```diff
@@ -1,6 +1,6 @@
 ﻿<FluentSwitch @bind-Value="@UseGlobalScrollbar" Label=" GlobalScrollbar" Margin="@Margin.All2" />
 <FluentSwitch @bind-Value="@UseHeaderSticky" Label="Header Sticky" Margin="@Margin.All2" Disabled="@(!UseGlobalScrollbar)" />
-<FluentSwitch @bind-Value="@UseMenuSticky" Label="Menu Sticky" Margin="@Margin.All2" Disabled="@(!UseGlobalScrollbar)" />
+<FluentSwitch @bind-Value="@UseNavigationSticky" Label="Menu Sticky" Margin="@Margin.All2" Disabled="@(!UseGlobalScrollbar)" />
```
dvoituron: Change to `Label="Navigation Sticky"`

src/Core/Enums/LayoutArea.cs
```diff
@@ -15,7 +16,8 @@ public enum LayoutArea
     Footer,
 
     /// <summary />
-    Menu,
+    [Description("nav")]
```
dvoituron, line 21: Why do you need this attribute?

dvoituron (changes requested):

src/Core/Enums/LayoutArea.cs
vnbaaij: So navigation can be shortened to nav in CSS (AddGridAreaStyles is using the ToAttributeValue to get the name from the enum). Think in CSS it is better to use 'nav' instead of 'navigation'

dvoituron: OK. Can you add the Description attribute on the other enum values (to be coherent)?

And including an empty line between each items.

dvoituron (approved):

PR #4362, opened 2025-11-27
[dev-v5] Migrate FluentOverlay

MarvinKlein1508:
This PR is a draft for a first exchange about the component structure. It includes the basic FluentOverlay component and all demos from v4.

I will close this PR after exchange and open a new one once the component is fully ready.

## Still missing

- [ ] Tests
- [ ] Code comments
- [ ] Migration guide

### DIFF
--- a/src/Core/Components/Overlay/FluentOverlay.razor.cs
+++ b/src/Core/Components/Overlay/FluentOverlay.razor.cs
@@ -38,8 +38,7 @@ public partial class FluentOverlay : FluentComponentBase, IAsyncDisposable
     /// <summary />
     protected string? StyleValue => new StyleBuilder()
         .AddStyle("cursor", "auto", () => Transparent)
-        //.AddStyle("background-color", $"rgba({_r}, {_g}, {_b}, {Opacity.ToString()})", () => !Transparent)
-        ////.AddStyle("opacity", Opacity.ToString()!.Replace(',', '.'), CheckCSSVariableName().IsMatch(BackgroundColor))
+        .AddStyle("background-color", string.Create(CultureInfo.InvariantCulture, $"rgba({_r}, {_g}, {_b}, {Opacity})"), () => !Transparent)
         .AddStyle("cursor", "default", () => !Transparent)
         .AddStyle("position", FullScreen ? "fixed" : "absolute")
         .AddStyle("display", "flex")
--- a/src/Core/Components/Overlay/FluentOverlay.razor.cs
+++ b/src/Core/Components/Overlay/FluentOverlay.razor.cs
@@ -1,0 +1,289 @@
+// ------------------------------------------------------------------------
+// This file is licensed to you under the MIT License.
+// ------------------------------------------------------------------------
+
+using System.Globalization;
+using System.Text.RegularExpressions;
+using Microsoft.AspNetCore.Components;
+using Microsoft.AspNetCore.Components.Web;
+using Microsoft.FluentUI.AspNetCore.Components.Extensions;
+using Microsoft.FluentUI.AspNetCore.Components.Utilities;
+using Microsoft.JSInterop;
+
+namespace Microsoft.FluentUI.AspNetCore.Components;
+
+/// <summary />
+public partial class FluentOverlay : FluentComponentBase, IAsyncDisposable
+{
+    private readonly string _defaultId = Identifier.NewId();
+    private string? _color;
+    private int _r, _g, _b;
+
+    //private const string JAVASCRIPT_FILE = "./_content/Microsoft.FluentUI.AspNetCore.Components/Components/Overlay/FluentOverlay.razor.js";
+    private const string DEFAULT_NEUTRAL_COLOR = "#808080";
+
+    private DotNetObjectReference<FluentOverlay>? _dotNetHelper;
+
+    /// <summary />
+    public FluentOverlay(LibraryConfiguration configuration) : base(configuration) { }
+
+    /// <summary />
+    private IJSObjectReference? _jsModule { get; set; }
+
+    /// <summary />
+    protected string? ClassValue => new CssBuilder("fluent-overlay")
+        .AddClass("prevent-scroll", PreventScroll)
+        .Build();
+
+    /// <summary />
+    protected string? StyleValue => new StyleBuilder()
+        .AddStyle("cursor", "auto", () => Transparent)
+        //.AddStyle("background-color", $"rgba({_r}, {_g}, {_b}, {Opacity.ToString()})", () => !Transparent)
+        ////.AddStyle("opacity", Opacity.ToString()!.Replace(',', '.'), CheckCSSVariableName().IsMatch(BackgroundColor))
+        .AddStyle("cursor", "default", () => !Transparent)
+        .AddStyle("position", FullScreen ? "fixed" : "absolute")
+        .AddStyle("display", "flex")
+        .AddStyle("align-items", Alignment.ToAttributeValue())
+        .AddStyle("justify-content", Justification.ToAttributeValue())
+        .AddStyle("pointer-events", "none", () => Interactive)
+        .AddStyle("z-index", ZIndex.Overlay.ToString(CultureInfo.InvariantCulture))
+        .Build();
+
+    /// <summary />
+    protected string? StyleContentValue => new StyleBuilder()
+        .AddStyle("pointer-events", "auto", () => Interactive)
+        .Build();
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the overlay is visible.
+    /// </summary>
+    [Parameter]
+    public bool Visible { get; set; } = false;
+
+    /// <summary>
+    /// Callback for when overlay visibility changes.
+    /// </summary>
+    [Parameter]
+    public EventCallback<bool> VisibleChanged { get; set; }
+
+    /// <summary>
+    /// Callback for when the overlay is closed.
+    /// </summary>
+    [Parameter]
+    public EventCallback<MouseEventArgs> OnClose { get; set; }
+
+    /// <summary>
+    /// Gets or set if the overlay is transparent.
+    /// </summary>
+    [Parameter]
+    public bool Transparent { get; set; } = true;
+
+    /// <summary>
+    /// Gets or sets the opacity of the overlay.
+    /// Default is 0.4.
+    /// </summary>
+    [Parameter]
+    public double? Opacity { get; set; }
+
+    /// <summary>
+    /// Gets or sets the alignment of the content to a <see cref="HorizontalAlignment"/> value.
+    /// Defaults to Align.Center.
+    /// </summary>
+    [Parameter]
+    public HorizontalAlignment Alignment { get; set; } = HorizontalAlignment.Center;
+
+    /// <summary>
+    /// Gets or sets the justification of the content to a <see cref="AspNetCore.Components.JustifyContent"/> value.
+    /// Defaults to JustifyContent.Center.
+    /// </summary>
+    [Parameter]
+    public JustifyContent Justification { get; set; } = JustifyContent.Center;
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the overlay is shown full screen or bound to the containing element.
+    /// </summary>
+    [Parameter]
+    public bool FullScreen { get; set; } = false;
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the overlay is interactive, except for the element with the specified <see cref="InteractiveExceptId"/>.
+    /// In other words, the elements below the overlay remain usable (mouse-over, click) and the overlay will closed when clicked.
+    /// </summary>
+    [Parameter]
+    public bool Interactive { get; set; } = false;
+
+    /// <summary>
+    /// Gets or sets the HTML identifier of the element that is not interactive when the overlay is shown.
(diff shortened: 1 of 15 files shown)


src/Core/Components/Overlay/FluentOverlay.razor.cs
```diff
@@ -0,0 +1,269 @@
...
+using Microsoft.FluentUI.AspNetCore.Components.Utilities;
+using Microsoft.JSInterop;
+
+namespace Microsoft.FluentUI.AspNetCore.Components;
+
+/// <summary />
+public partial class FluentOverlay : FluentComponentBase
+{
+    private const string JAVASCRIPT_FILE = FluentJSModule.JAVASCRIPT_ROOT + "Overlay/FluentOverlay.razor.js";
+    private const string DEFAULT_NEUTRAL_COLOR = "#808080";
```
MarvinKlein1508: Should we use a FluentUI CSS variable here?

src/Core/Components/Overlay/FluentOverlay.razor.cs
```diff
@@ -0,0 +1,269 @@
...
+    /// </summary>
+    [Parameter]
+    public double? Opacity { get; set; }
+
+    /// <summary>
+    /// Gets or sets the alignment of the content to a <see cref="HorizontalAlignment"/> value.
+    /// Defaults to Align.Center.
+    /// </summary>
+    [Parameter]
+    public HorizontalAlignment Alignment { get; set; } = HorizontalAlignment.Center;
```
MarvinKlein1508, line 86: I exchanged this from `Align` to `HorizontalAlignment` (Breaking change)

src/Core/Components/Overlay/FluentOverlay.razor.cs
```diff
@@ -0,0 +1,269 @@
...
+
+    /// <summary>
+    /// Gets or sets the justification of the content to a <see cref="AspNetCore.Components.JustifyContent"/> value.
+    /// Defaults to JustifyContent.Center.
+    /// </summary>
+    [Parameter]
+    public JustifyContent Justification { get; set; } = JustifyContent.Center;
```
MarvinKlein1508, line 93: Values from this Enum have been changed in v5 (Breaking change)

src/Core/Components/Overlay/FluentOverlay.razor.cs
```diff
@@ -0,0 +1,269 @@
...
+        if (firstRender)
+        {
+            _dotNetHelper ??= DotNetObjectReference.Create(this);
+            await JSModule.ImportJavaScriptModuleAsync(JAVASCRIPT_FILE);
+        }
+    }
+    /// <summary />
+    protected override async Task OnParametersSetAsync()
+    {
+        if (Interactive && JSModule.Imported)
```
MarvinKlein1508, line 150: Using `JSModule.Imported` here because the component registers events based on it's parameters. Please confirm wether I'm using it right here.

src/Core/Components/Overlay/FluentOverlay.razor.cs
```diff
@@ -0,0 +1,269 @@
...
+    /// <summary />
+    protected string? ClassValue => DefaultClassBuilder
+        .AddClass("fluent-overlay")
+        .AddClass("prevent-scroll", PreventScroll)
+        .Build();
+
+    /// <summary />
+    protected string? StyleValue => DefaultStyleBuilder
+        .AddStyle("cursor", "auto", () => Transparent)
+        .AddStyle("background-color", string.Create(CultureInfo.InvariantCulture, $"rgba({_r}, {_g}, {_b}, {Opacity})"), () => !Transparent)
```
MarvinKlein1508: We should probably change this in v4 as well to save some memory allocations

src/Core/Components/Overlay/FluentOverlay.razor.cs
```diff
@@ -0,0 +1,269 @@
...
+        if (OnClose.HasDelegate)
+        {
+            await OnClose.InvokeAsync(e);
+        }
+    }
+
+    /// <summary />
+    protected override async ValueTask DisposeAsync(IJSObjectReference jsModule)
+    {
+        await InvokeOverlayDisposeAsync();
```
MarvinKlein1508, line 211: Can we do it like this?

src/Core/Components/Overlay/FluentOverlay.razor.ts
```diff
@@ -0,0 +1,57 @@
+export namespace Microsoft.FluentUI.Blazor.Overlay {
```
MarvinKlein1508, line 3: My first time typescript. Please verify. I put this together by taking a look on the other typescript files in the project.

src/Core/Microsoft.FluentUI.AspNetCore.Components.csproj
```diff
@@ -57,6 +57,10 @@
     <None Include="**\*.ts" />
   </ItemGroup>
 
+  <ItemGroup>
+    <TypeScriptCompile Remove="components\overlay\FluentOverlay.razor.ts" />
```
dvoituron: Can you delete this line?

src/Core/Microsoft.FluentUI.AspNetCore.Components.csproj
```diff
@@ -122,7 +126,7 @@
     </ItemGroup>
 
     <!-- Log before modification -->
-    <Message Importance="normal" Text="Found @(_ContentWithAbsolutePaths->Count()) Content items with absolute paths to normalize" />
+    <Message Importance="normal" Text="Found @(_ContentWithAbsolutePaths-&gt;Count()) Content items with absolute paths to normalize" />
```
dvoituron: And can you rollback this change?

src/Core/Components/Overlay/FluentOverlay.razor.cs
vnbaaij: Yes! (but don't ask which one 😁)

dvoituron: In my mind, this `#808080` value is the default `BackgroundColor` value. So this parameter should be "not null" and with this default value.

src/Core/Components/Overlay/FluentOverlay.razor.cs
dvoituron: The `OnParametersSetAsync` should be empty (for performance reason). 
All "initialization" code should be inside the `OnInitialize`. 
Why do we want to detect the changes of `Transparent`, `BackgroundColor`, `Opacity`, ...

src/Core/Components/Overlay/FluentOverlay.razor.ts
```diff
@@ -0,0 +1,57 @@
...
+  export function Initialize(dotNetHelper: any, containerId: string, id: string) {
```
dvoituron: `dotNetHelper` must be typed (not `any` type)
And `Initialize` must "returns" `void`

src/Core/Components/Overlay/FluentOverlay.razor.ts
```diff
@@ -0,0 +1,57 @@
...
+
+      // Remove the event listener
+      document.removeEventListener('click', _document.fluentOverlayData[id].clickHandler);
+
+      // Remove the data
+      _document.fluentOverlayData[id] = null;
+      delete _document.fluentOverlayData[id];
+    }
+  }
+  function isClickInsideContainer(event: MouseEvent, container: any) {
```
dvoituron: Avoid (id possible) to use `any`. This is not a good practice. So, use the `HtmlElement` or other type

src/Core/Utilities/ZIndex.cs
```diff
@@ -34,4 +34,9 @@ public static class ZIndex
     /// ZIndex for the popup components in the <see cref="FluentDataGrid{TGridItem}"/>.
     /// </summary>
     public static int DataGridHeaderPopup { get; set; } = 5;
+
+    /// <summary>
+    /// ZIndex for the <see cref="FluentOverlay" /> component.
+    /// </summary>
+    public static int Overlay { get; set; } = 9900;
```
dvoituron: The Tooltip will be above the Overlay?

dvoituron (changes requested):

src/Core/Components/Overlay/FluentOverlay.razor.ts
MarvinKlein1508: which type is it? Any other .ts file defines it as any in the project as well

src/Core/Components/Overlay/FluentOverlay.razor.ts
MarvinKlein1508: Using `HTMLElement` here breaks the code for some reason.

src/Core/Components/Overlay/FluentOverlay.razor.ts
dvoituron: Mmmm. In the **Scripts** project we can import the `Microsoft.JSInterop` but not yet in the **Components** project. I will update that in another PR.
```js
import { DotNet } from "../../d-ts/Microsoft.JSInterop";
```

dvoituron: I come back with that...

Try to add this, in top of your file:
```ts
import { DotNet } from "../../../Core.Scripts/src/d-ts/Microsoft.JSInterop";
```

And the type will be `DotNet.DotNetObject`

src/Core/Utilities/ZIndex.cs
MarvinKlein1508: Which value would you recommend? I copied this one over from v4. In v4 there is no ZINdex for tooltips

MarvinKlein1508:
@dvoituron @vnbaaij if we stick with the current code then I'm going to prepare remaining tests and documentation.

src/Core/Components/Overlay/FluentOverlay.razor.cs
MarvinKlein1508: See #4366

src/Core/Components/Overlay/FluentOverlay.razor.cs
```diff
@@ -0,0 +1,271 @@
...
+    [Parameter]
+    public bool Dismissable { get; set; } = true;
+
+    /// <summary>
+    /// Gets or sets the background color.
+    /// Needs to be formatted as an HTML hex color string (#rrggbb or #rgb)
+    /// Default NeutralBaseColor token value (#808080).
+    /// </summary>
+    [Parameter]
+    public string BackgroundColor { get; set; } = "#808080";
```
dvoituron, line 127: I never liked this requirement to use a HEX color.
Now, all modern browsers support the [color-mix](https://caniuse.com/wf-color-mix) CSS function. Can you check if the result is the same?

Example:
```html
<div id="id1">Hello</div>
<div id="id2">World</div>
<div id="id3">Nice</div>

<style>
  #id1 {
    background-color: color-mix(in srgb, #FF0000 10%, transparent);
  }
  #id2 {
    background-color: color-mix(in srgb, blue 10%, transparent);
  }
  #id3 {
    --my-color: green;
    background-color: color-mix(in srgb, var(--my-color) 10%, transparent);
  }
</style>
```

Image

dvoituron (changes requested):

src/Core/Components/Overlay/FluentOverlay.razor.cs
MarvinKlein1508: Done works fine. Introduces another breaking change because Opacity has been changed from `double?` to `int?`

dvoituron: Why? You can keep a `double?` and multiply this value to 100 to have a percentage :-)

MarvinKlein1508: Sure we could multiply it but wouldn't it be more clear to the developer to pass in the percentage directly? It's a small breaking change and we save us the computation.

vnbaaij: I'm confused. What is value now for the `BackgroundColor`  then?

MarvinKlein1508: @vnbaaij it can be anything now.
``` 
red
#FF0
#FF0000
hsl(0, 100%, 50%)
rgba(255, 0, 0, 1)
var(--my-color)
```

dvoituron: > Sure we could multiply it but wouldn't it be more clear to the developer to pass in the percentage directly? It's a small breaking change and we save us the computation.

I don't really see why it's simpler (0.5 or 50 is clear to me). Furthermore, why prevent percentages with decimal points? This is accepted by the CSS standard.
> A percentage value specifying the amount of the corresponding color to mix; can be any <percentage> value between 0% and 100%, inclusive.

But I can understand both views.

MarvinKlein1508: OK. I changed it back to double. But I leave out the part with multiplying it by 100 as this might cause confusion in which format the value has to be entered. Now it is pretts simple. You want 45.5%? Enter 40.5.

MarvinKlein1508:
I've updated the `Transparent` parameter to be `false` by default. This aligns with the removed checks in `OnInitialized` which set this value to false when an `Opacity` is provided which by default was provided anyways.

PR #4370, opened 2025-11-29
[dev-v5] [Core] Add a build-all-lib YAML to sign the RC NuGet packages

dvoituron-msft:
Add a build-all-lib YAML to sign the RC NuGet packages

### DIFF
--- a/eng/pipelines/build-all-lib.yml
+++ b/eng/pipelines/build-all-lib.yml
@@ -1,0 +1,235 @@
+# Build all projects (manually triggered or tag-based).
+name: $(FileVersion).$(Year:yy)$(DayOfYear).$(Rev:r)
+
+trigger:
+  branches:
+    include: []         # Explicitly disable branch triggers
+  tags:
+    include:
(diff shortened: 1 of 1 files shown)


eng/pipelines/build-all-lib.yml
```diff
@@ -0,0 +1,235 @@
...
+    include: []         # Explicitly disable branch triggers
+  tags:
+    include:
+      - v*              # Triggers on tags like v1.0.0, v2.1.3
+pr: none         # Disable pull request triggers.
+
+parameters:
+  - name: Projects  	    # List of projects to build
+    default: |
+      **/Microsoft.FluentUI.AspNetCore.Components.csproj
```
vnbaaij, line 15: This only builds the main components project. Is that what you intended (as it is called 'build all' )?

dvoituron: Yes. At this moment the Icons, Emojis are published using v4. And there are no Templates projects 😀

vnbaaij (approved):

PR #4371, opened 2025-11-29
[dev-v5] Sort spelling and remove duplicate entries

MarvinKlein1508:
Title speaks for itself.


spelling.dic
```diff
@@ -54,52 +59,30 @@ onclosecolumnoptions
 onclosecolumnresize
 ondialogbeforetoggle
 ondropdownchange
+oninput
 onkeydown
 onmenuitemchange
 ontabchange
 ontreechanged
 ontreetoggle
+Overscan
```
dvoituron: Can you change to lowercase. To be coherent with all words.

dvoituron (changes requested):

PR #4377, opened 2025-12-03
[dev-v5] launchBrowser true

MarvinKlein1508:
Re-enables the browser launch in dev-v5 for easier debugging


dvoituron:
I updated the `launchBrowser` flag because:

1. `launchBrowser=true` 

   The browser will be automatically opened but the Visuals Studio kill the previous process to restart a new one. The URL is `/` and the current page is the Home page.

   https://github.com/user-attachments/assets/183034d7-c831-48fc-9837-10c2dcd49c86

2. `launchBrowser=false` 

   You need to click on the link to open the browser but Visual Studio restart the process and refresh the **current page**.

   https://github.com/user-attachments/assets/8fb53b14-c4d3-46ee-9610-53bb9291d001

Do you have the same result? Did you configure VS2026 using another flag?
What is the best solution?

When debugging a single component (page), I believe the second solution is preferable.

vnbaaij:
I didn't think of using it to get the page to refresh. That is a nice thing to have!

It is just annoying it doesn't open the site automatically. I'm so used to that by now that it feels like something is going wrong if it doesn't happen.

vnbaaij:
I didn't think of using it to get the page to refresh. That is a nice thing to have!

It is just annoying it doesn't open the site automatically. I'm so used to that by now that it feels like something is going wrong if it doesn't happen.

dvoituron:
Yes I know. Maybe another flag exists in VS but not yet found 🙁

MarvinKlein1508:
@vnbaaij @dvoituron how about a second launch profile?

examples/Demo/FluentUI.Demo/Properties/launchSettings.json
```diff
@@ -1,15 +1,25 @@
...
-        "commandName": "Project",
-        "dotnetRunMessages": true,
-        "launchBrowser": false,
-        "inspectUri": "{wsProtocol}://{url.hostname}:{url.port}/_framework/debug/ws-proxy?browser={browserInspectUri}",
-        "applicationUrl": "https://localhost:7197;http://localhost:5219",
-        "environmentVariables": {
-          "ASPNETCORE_ENVIRONMENT": "Development"
-        }
+  "profiles": {
+    "WithBrowser": {
```
dvoituron: Can you keep the same name `https`

examples/Demo/FluentUI.Demo/Properties/launchSettings.json
```diff
@@ -1,15 +1,25 @@
...
+      "commandName": "Project",
+      "dotnetRunMessages": true,
+      "launchBrowser": true,
+      "inspectUri": "{wsProtocol}://{url.hostname}:{url.port}/_framework/debug/ws-proxy?browser={browserInspectUri}",
+      "applicationUrl": "https://localhost:7197;http://localhost:5219",
+      "environmentVariables": {
+        "ASPNETCORE_ENVIRONMENT": "Development"
+      }
+    },
+    "WithoutBrowser": {
```
dvoituron: Can you use a more explicit name like `launch-browser`?

MarvinKlein1508:
Done. As a bonus: you can configure the behaviour in VS.
grafik

PR #4399, opened 2025-12-10
[dev-v5] Update preview usage documentation

MarvinKlein1508:
This PR updates the documentation section for setting up the preview feed for dev-v5. In addition it also adds documentation to use preview versions which do not have a signature applied.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,213 @@
+# AGENTS.md
+
+This file provides context and instructions for AI coding agents working on the **Microsoft Fluent UI Blazor** components library.
+
+## Project Overview
+
+This is a .NET Blazor component library that provides Fluent UI design system components for Blazor applications. The library wraps Microsoft's official **Fluent UI Web Components** and provides additional components leveraging the Fluent Design System.
+
+- **Main Package**: `Microsoft.FluentUI.AspNetCore.Components`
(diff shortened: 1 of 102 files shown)


vnbaaij:
The name change is indeed not necessary. 

There are no preview packages for icons/emoji so they need to come from public NuGet feed. The library needs to come from our preview feed.

MarvinKlein1508:
I do this because of #4356

Defining the `packageSource` fixes this issue. I specified `Emoji` and `Icons` in nuget because they have the same prefix. Without this my nuget gets confused because it tries to resolve those packages from dotnet9 which do not exist there. The name change to `.org` was copied from my file over. Can revert this one if necessary.

dvoituron:
> I do this because of #4356

It has absolutely nothing to do with it: this 4356 change was made because Microsoft's ADO pipeline refuses to have two feeds in a nuget.config

The change you are proposing id the documentation is too complex for users who are just starting out. We cannot list all possible configurations for all use cases.
What problem did you encounter with this configuration alone? 
Normally, NuGet uses the first feed, and if the package is not there, it moves on to the next one. Simple and effective :-)

vnbaaij:
I'm closing this one for now. We can make some changes later if/when we have a signed version of the package and also have the icons/emoiji packages

PR #4409, opened 2025-12-14
[dev-v5] FluentOverlay

dvoituron:
# [dev-v5] FluentOverlay
The Overlays are used to temporarily overlay screen content to focus a dialog, progress or other information/interaction.

When the `FullScreen` parameter is set to `false`, the overlay will only cover the content area of its parent container.

The overlay can be customized to suit various use cases, including:

- **Full-Screen Overlay**: By setting the `FullScreen` property to `true`, the overlay will cover the entire screen.
- **Interactive Overlay**: The `Interactive` property allows user interactions with the overlay content.
- **Custom Close Modes**: The `CloseMode` property can be set to control how the overlay can be closed (e.g., by clicking outside of it or only by code).

```razor
  <FluentOverlay Style="background: transparent;"
                 Class="my-class"
                 FullScreen="@FullScreen"
                 Interactive="@Interactive"
                 @bind-Visible="@Visible"
                 CloseMode="@CloseMode">
        <FluentSpinner />
    </FluentOverlay>
```

peek_3

## Unit Tests

image

### DIFF
--- a/src/Core/Components/Base/FluentComponentBase.cs
+++ b/src/Core/Components/Base/FluentComponentBase.cs
@@ -24,6 +24,7 @@ public abstract class FluentComponentBase : ComponentBase, IAsyncDisposable, IFl
     protected FluentComponentBase(LibraryConfiguration configuration)
     {
         configuration?.DefaultValues.ApplyDefaults(this);
+        LibraryConfiguration = configuration;
     }
 
     [Inject]
@@ -36,6 +37,11 @@ public abstract class FluentComponentBase : ComponentBase, IAsyncDisposable, IFl
     /// <summary />
     [Inject]
     protected IFluentLocalizer Localizer { get; set; } = FluentLocalizerInternal.Default;
+
+    /// <summary>
+    /// Gets the configuration settings for the library instance.
+    /// </summary>
+    protected internal LibraryConfiguration? LibraryConfiguration { get; }
 
     /// <summary>
     /// Gets the JavaScript module imported with the <see cref="FluentJSModule.ImportJavaScriptModuleAsync"/> method.
--- a/src/Core/Components/DateTime/FluentCalendar.razor.cs
+++ b/src/Core/Components/DateTime/FluentCalendar.razor.cs
@@ -22,9 +22,9 @@ public partial class FluentCalendar<TValue> : FluentCalendarBase<TValue>
 {
     private ElementReference _calendarReference = default!;
     private const string JAVASCRIPT_FILE = FluentJSModule.JAVASCRIPT_ROOT + "DateTime/FluentCalendar.razor.js";
-    
-    internal static string ArrowUp = "<svg width=\"24\" height=\"24\" viewBox=\"0 0 24 24\" xmlns=\"http://www.w3.org/2000/svg\"><path d=\"M4.2 10.73a.75.75 0 001.1 1.04l5.95-6.25v14.73a.75.75 0 001.5 0V5.52l5.95 6.25a.75.75 0 001.1-1.04l-7.08-7.42a1 1 0 00-1.44 0L4.2 10.73z\"/></svg>";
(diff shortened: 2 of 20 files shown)


src/Core.Scripts/src/Components/Overlay/FluentOverlay.ts
```diff
@@ -70,6 +70,42 @@ export namespace Microsoft.FluentUI.Blazor.Components.Overlay {
       shadow.appendChild(this.dialog);
     }
 
+    /*************** 
+      Attributes
+    ****************/
+
+    static get observedAttributes() { return ['dialogStyle', 'dialogClass', 'visible']; }
```
MarvinKlein1508: should contain background as it is observed below as well

src/Core.Scripts/src/Components/Overlay/FluentOverlay.ts
```diff
@@ -331,6 +380,24 @@ export namespace Microsoft.FluentUI.Blazor.Components.Overlay {
 
   }
 
+  /**
+   * Display the fluent-overlay with the given id
+   * @param id The id of the fluent-overlay to display
+   */
+  export function Show(id: string): void {
+    const element = document.getElementById(id) as any;
```
MarvinKlein1508: can we define it as FluentOverlay?

src/Core/Components/Overlay/FluentOverlay.razor.cs
```diff
@@ -0,0 +1,128 @@
...
+    protected string? ClassValue => DefaultClassBuilder
+        .Build();
+
+    /// <summary />
+    protected string? StyleValue => DefaultStyleBuilder
+        .Build();
+
+    /// <inheritdoc cref="IFluentComponentElementBase.Element" />
+    [Parameter]
+    public ElementReference Element { get; set; }
```
MarvinKlein1508: what is the purpose of this? It isn't used in the FluentOverlay.razor

src/Core/Components/Overlay/FluentOverlay.razor.cs
```diff
@@ -0,0 +1,128 @@
...
+    /// </summary>
+    [Parameter]
+    public string BackgroundColor { get; set; } = "var(--colorBackgroundOverlay)";
+
+    /// <summary>
+    /// Gets or sets the opacity level of the overlay background.
+    /// Default is `0.2` (20%).
+    /// </summary>
+    [Parameter]
+    public double Opacity { get; set; } = 0.4;
```
MarvinKlein1508: I would consider changing this to 40 and remove the `* 100` logic. It is a breaking change but it makes it much clearer on how to use the component. If you want 40.5% you provide the value 40.5.

src/Core/Components/Overlay/FluentOverlay.razor.cs
```diff
@@ -0,0 +1,128 @@
...
+            if (VisibleChanged.HasDelegate)
+            {
+                await VisibleChanged.InvokeAsync(Visible);
+            }
+        }
+    }
+
+    private string GetBackground()
+    {
+        return $"color-mix(in srgb, {BackgroundColor} {Opacity * 100}%, transparent)";
```
MarvinKlein1508: you need to use string.create here with localization support. Otherwise this won't work if you get a decimal number and your decimal separtor is a `,` like in Germany.

src/Core.Scripts/src/Components/Overlay/FluentOverlay.ts
dvoituron: Done

src/Core/Components/Overlay/FluentOverlay.razor.cs
dvoituron: Done

src/Core/Components/Overlay/FluentOverlay.razor.cs
dvoituron: Done

src/Core/Components/Overlay/FluentOverlay.razor.cs
dvoituron: it is no longer useful

src/Core.Scripts/src/Components/Overlay/FluentOverlay.ts
dvoituron: Done

MarvinKlein1508 (approved):

PR #4417, opened 2025-12-20
Remove NotFound components from the .NET 10 Microsoft.NET.Sdk.BlazorW…

WesHaze:
# Remove NotFound components from the .NET 10 Microsoft.NET.Sdk.BlazorWebAssembly template
## 📖 Description
Remove NotFound components from the .NET 10 Microsoft.NET.Sdk.BlazorWebAssembly template App.Razor implementation because dual use in combination with the Router's NotFoundPage prop is not supported

I chose this implementation over the NotFoundComponent because there is a routablecomponent/page for the notfound in this version of the template, the NotfoundPage is more aligned with the other template version however

Before
image

After
image

### 🎫 Issues
No relevant issues found

## 👩‍💻 Reviewer Notes
The reboot.css is present when building using so I presume its not an issue:
dotnet publish -c Release -r win-x64 --self-contained true

The issue itself is not present in the net 9 template

## 📑 Test Plan
There seemingly are not affected tests.

### General
- [ x] I have tested my changes.
- [ x] I have read the [CONTRIBUTING](https://github.com/microsoft/fluentui-blazor/blob/master/docs/contributing.md) documentation and followed the [standards](https://www.fast.design/docs/community/code-of-conduct/#our-standards) for this project.

## ⏭ Next Steps
None

### DIFF
--- a/examples/Demo/Shared/Microsoft.FluentUI.AspNetCore.Components.xml
+++ b/examples/Demo/Shared/Microsoft.FluentUI.AspNetCore.Components.xml
@@ -7264,6 +7264,13 @@
             Most important info to be shown in the message bar.
             </summary>
         </member>
+        <member name="P:Microsoft.FluentUI.AspNetCore.Components.Message.UseMarkupString">
+            <summary>
+            Indicates whether the title and body should be rendered as markup string.
+            Using MarkupString can introduce XSS vulnerabilities because it renders unencoded HTML.
+            Only use it with fully trusted, sanitized content.
+            </summary>
+        </member>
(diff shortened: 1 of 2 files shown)


vnbaaij:
I won't be merging this as it is now. Each .NET version has its own template. I changed that a while ago. Before, I solved it by adding conditions and exceptions in the templates source code.
So the .NET 10 templates should use the .NET 10 components like NotFound. If there are errors we need to work on fixing those, not removing the components or reverting the changes.

**Update**:
Hmm, I don't see the NotFound component in the aspnetcore repo either... Will investigate further.

Issue #2473, opened 2024-07-31
feat: add notion of parent/child in the menu data structures


MarvinKlein1508:
# 🙋 Feature Request

The FluentNav component should know which child items are active to expand the navigation bar correctly when browsing the page through a bookmark or external link.

## 🔦 Context
Related to #2297 & #2142

vnbaaij:
This is going to take a significant amount of work, I believe. We will investigate for the next major version

MarvinKlein1508:
I assign you since denis tagged you for FluentMenu in v5. Perhaps this is something we can consider.

MarvinKlein1508:
Done in #4419

Issue #2142, opened 2024-06-03
fix: navigation does not expand when navigating through the page by bookmark or link


MarvinKlein1508:
# 🐛 Bug Report

When navigating to a certain page which is nested underneith a `FluentNavGroup` the correct item will get the active class but the `FluentNavGroup` will be collapsed by default. This is really annoying when you come from bookmarks or external links to your website.

Please see this video for a demonstration:
https://github.com/microsoft/fluentui-blazor/assets/32510006/cb3d385d-4de1-4a79-81cf-0232f2bc3a40

## 💻 Repro or Code Sample
I'Ve created a small sample repo. Simply click on Show weather in Home.razor

https://github.com/MarvinKlein1508/ExpandBugFluentUi

## 🤔 Expected Behavior

I would expect `FluentNavGroup` to be expaned when visiting /weater through the link on Home.razor

## 😯 Current Behavior

More is collapsed but the item underneith More has active classes.

## 🌍 Your Environment

.NET 8.0.6
Firefox latest version

vnbaaij:
We have no notion of a 'parent/child' relationship in the current `NavMenu`. I don't see a way how we could do this at the moment.
Any ideas?

vnbaaij:
Closing this as we do not have a solution for this for the foreseeable future

PR #4419, opened 2025-12-22
[dev-v5][Nav] Part 1 - Add the FluentNav components 

vnbaaij:
Add the `FluentNav` family of components including docs and tests (100% coverage). This is part 1 containing docs, components and tests. Next part will contain the changes to use this component in the demo site.

This is an implementation of the Nav as described in the [FluentDesign System](https://fluent2.microsoft.design/components/web/react/core/nav/usage) (and implemented in [React](https://storybooks.fluentui.dev/react/?path=/docs/components-nav--docs)). 

image

There is no upgrade/migration path vor the v4 FluentNavMenu and associated components as those offer multi level navigation. This implementation is simpler and only has one-level of nesting.

Unit tests Line coverage :
image
This pull request introduces a new `FluentNav` navigation component for Fluent UI, including its implementation, styling, documentation, and usage example. The component provides a flexible, collapsible navigation menu with support for icons, categories, section headers, and programmatic control. Additionally, new icon classes are added to the icon library to support navigation visuals.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,19 +1,19 @@
 <Project>
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
-    <RuntimeVersion9>9.0.9</RuntimeVersion9>
-    <AspNetCoreVersion9>9.0.9</AspNetCoreVersion9>
-    <EfCoreVersion9>9.0.9</EfCoreVersion9>
-    <RuntimeVersion10>10.0.0</RuntimeVersion10>
-    <AspNetCoreVersion10>10.0.0</AspNetCoreVersion10>
-    <EfCoreVersion10>10.0.0</EfCoreVersion10>
+    <RuntimeVersion9>9.0.11</RuntimeVersion9>
+    <AspNetCoreVersion9>9.0.11</AspNetCoreVersion9>
+    <EfCoreVersion9>9.0.11</EfCoreVersion9>
+    <RuntimeVersion10>10.0.1</RuntimeVersion10>
+    <AspNetCoreVersion10>10.0.1</AspNetCoreVersion10>
+    <EfCoreVersion10>10.0.1</EfCoreVersion10>
   </PropertyGroup>
   <!-- Independent from TargetFrameWorks-->
   <ItemGroup>
     <!-- For Sample Apps -->
-    <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components" Version="4.13.1" />
-    <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components.Icons" version="4.13.1" />
-    <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components.Emoji" Version="4.13.1" />
+    <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components" Version="4.13.2" />
+    <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components.Icons" version="4.13.2" />
+    <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components.Emoji" Version="4.13.2" />
     <PackageVersion Include="LoxSmoke.DocXml" Version="3.8.0" />
     <!-- Test dependencies -->
     <PackageVersion Include="bunit" Version="2.2.2" />
--- a/src/Core.Scripts/package-lock.json
+++ b/src/Core.Scripts/package-lock.json
@@ -1,6 +1,5 @@
 {
   "name": "microsoft.fluentui.aspnetcore.components.assets",
-  "version": "1.0.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
@@ -8,19 +7,19 @@
       "name": "microsoft.fluentui.aspnetcore.components.assets",
       "license": "ISC",
       "dependencies": {
-        "@fluentui/web-components": "3.0.0-beta.124",
+        "@fluentui/web-components": "3.0.0-rc.1",
         "imask": "^7.6.1"
       },
       "devDependencies": {
-        "@typescript-eslint/eslint-plugin": "8.44.0",
-        "@typescript-eslint/parser": "8.44.0",
-        "esbuild": "0.25.10",
+        "@typescript-eslint/eslint-plugin": "8.51.0",
+        "@typescript-eslint/parser": "8.51.0",
+        "esbuild": "0.27.2",
         "esbuild-plugin-inline-css": "0.0.1",
-        "eslint": "9.36.0",
-        "glob": "^11.1.0",
-        "rimraf": "6.0.1",
-        "tabbable": "6.2.0",
-        "typescript": "5.9.2"
+        "eslint": "9.38.0",
+        "glob": "^13.0.0",
+        "rimraf": "6.1.2",
+        "tabbable": "6.3.0",
+        "typescript": "5.9.3"
       }
     },
     "node_modules/@babel/runtime-corejs3": {
@@ -36,9 +35,9 @@
       }
     },
     "node_modules/@esbuild/aix-ppc64": {
-      "version": "0.25.10",
-      "resolved": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/@esbuild/aix-ppc64/-/aix-ppc64-0.25.10.tgz",
-      "integrity": "sha1-7mtxY6E1KOCZ7PViuXLyvOvgqpc=",
+      "version": "0.27.2",
+      "resolved": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/@esbuild/aix-ppc64/-/aix-ppc64-0.27.2.tgz",
+      "integrity": "sha1-Uhy9lo3PNiCUA0lH92+hsY0tQDw=",
       "cpu": [
         "ppc64"
       ],
@@ -53,9 +52,9 @@
       }
     },
     "node_modules/@esbuild/android-arm": {
-      "version": "0.25.10",
-      "resolved": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/@esbuild/android-arm/-/android-arm-0.25.10.tgz",
-      "integrity": "sha1-jVgRkS2nf2FTmGEeW7wTM/4yGqk=",
+      "version": "0.27.2",
+      "resolved": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/@esbuild/android-arm/-/android-arm-0.27.2.tgz",
+      "integrity": "sha1-VUiHgh4AndbYU/ly/ebFFD8d4UI=",
       "cpu": [
         "arm"
       ],
@@ -70,9 +69,9 @@
       }
     },
     "node_modules/@esbuild/android-arm64": {
-      "version": "0.25.10",
-      "resolved": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/@esbuild/android-arm64/-/android-arm64-0.25.10.tgz",
-      "integrity": "sha1-EV/HZjHoLdBoEb+vLbDUl5wW4ss=",
+      "version": "0.27.2",
+      "resolved": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/@esbuild/android-arm64/-/android-arm64-0.27.2.tgz",
+      "integrity": "sha1-YepVCWLYqhKpszGUOU4Adlem31c=",
       "cpu": [
         "arm64"
       ],
@@ -87,9 +86,9 @@
       }
     },
     "node_modules/@esbuild/android-x64": {
-      "version": "0.25.10",
-      "resolved": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/@esbuild/android-x64/-/android-x64-0.25.10.tgz",
-      "integrity": "sha1-4+llFrLVDXQQW7kllMRz4w3cFrE=",
+      "version": "0.27.2",
+      "resolved": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/@esbuild/android-x64/-/android-x64-0.27.2.tgz",
+      "integrity": "sha1-p86dByGCX8V4+SkqdtnlMzRIC6I=",
       "cpu": [
         "x64"
       ],
@@ -104,9 +103,9 @@
       }
     },
     "node_modules/@esbuild/darwin-arm64": {
-      "version": "0.25.10",
-      "resolved": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/@esbuild/darwin-arm64/-/darwin-arm64-0.25.10.tgz",
-      "integrity": "sha1-ava7HQWIfaxRXeGxYrWdxxIS7XY=",
+      "version": "0.27.2",
+      "resolved": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/@esbuild/darwin-arm64/-/darwin-arm64-0.27.2.tgz",
+      "integrity": "sha1-LLdlm9XRCYA8WTz8QURQ1UMMglY=",
       "cpu": [
         "arm64"
       ],
@@ -121,9 +120,9 @@
       }
     },
     "node_modules/@esbuild/darwin-x64": {
-      "version": "0.25.10",
-      "resolved": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/@esbuild/darwin-x64/-/darwin-x64-0.25.10.tgz",
-      "integrity": "sha1-ma6CNH+9M2/C0o/9TwVpTm5bcj0=",
+      "version": "0.27.2",
+      "resolved": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/@esbuild/darwin-x64/-/darwin-x64-0.27.2.tgz",
+      "integrity": "sha1-50H6axq7DNA2QSa6NMoX/V579Qk=",
       "cpu": [
         "x64"
       ],
@@ -138,9 +137,9 @@
       }
     },
     "node_modules/@esbuild/freebsd-arm64": {
-      "version": "0.25.10",
-      "resolved": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/@esbuild/freebsd-arm64/-/freebsd-arm64-0.25.10.tgz",
-      "integrity": "sha1-DG1VWKYyKwvbF/cCXBm9fSNZQ30=",
+      "version": "0.27.2",
+      "resolved": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/@esbuild/freebsd-arm64/-/freebsd-arm64-0.27.2.tgz",
+      "integrity": "sha1-K2TnEWhlyhctTOA0EUwh88k+OXw=",
       "cpu": [
         "arm64"
       ],
@@ -155,9 +154,9 @@
       }
     },
     "node_modules/@esbuild/freebsd-x64": {
-      "version": "0.25.10",
-      "resolved": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/@esbuild/freebsd-x64/-/freebsd-x64-0.25.10.tgz",
-      "integrity": "sha1-jDWHP6uMCFenUwCj3M5DJMoLmEQ=",
+      "version": "0.27.2",
+      "resolved": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/@esbuild/freebsd-x64/-/freebsd-x64-0.27.2.tgz",
+      "integrity": "sha1-5SUlUeZvSZ5JNO+2EYEvOCDpkLs=",
       "cpu": [
         "x64"
       ],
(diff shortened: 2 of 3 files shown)


src/Core/Components/Nav/FluentNavItem.razor.cs
```diff
@@ -0,0 +1,364 @@
...
+            // This is because the router accepts an absolute URI value of "same
+            // as base URI but without trailing slash" as equivalent to "base URI",
+            // which in turn is because it's common for servers to return the same page
+            // for http://host/vdir as they do for host://host/vdir/ as it's no
+            // good to display a blank page in that case.
+            if (hrefAbsolute[^1] == '/' &&
+                currentUriAbsolute.SequenceEqual(hrefAbsolute[..^1], CaseInsensitiveComparer))
+            {
+                return true;
+            }
```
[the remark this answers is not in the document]
vnbaaij: Keep as-is (copied from Blazor

src/Core/Components/Nav/FluentNavItem.razor.cs
```diff
@@ -0,0 +1,364 @@
...
+    /// </summary>
+    [Parameter]
+    public bool Disabled { get; set; }
+
+    /// <summary>
+    /// Gets or sets the target attribute that specifies where to open the group, if Href is specified.
+    /// Possible values: _blank | _self | _parent | _top.
+    /// </summary>
+    [Parameter]
+    public string? Target { get; set; }
```
dvoituron: You can use the existing enum `LinkTarget`.

src/Core/Components/Nav/FluentNavItem.razor.cs
```diff
@@ -0,0 +1,364 @@
...
+    /// </summary>
+    /// <remarks>This property is typically set automatically by the Blazor framework when the component is
+    /// used within a <see cref="FluentNav"/>. It enables the component to access shared state or functionality from
+    /// its parent navigation menu.</remarks>
+    [CascadingParameter(Name = "Category")]
+    public FluentNavCategory? Category { get; set; }
+
+    /// <summary />
+    [Inject]
+    public required NavigationManager NavigationManager { get; set; }
```
dvoituron: Can you move all injected properties before all other properties (top of the file)?

src/Core/Components/Nav/FluentNavItem.razor.cs
```diff
@@ -0,0 +1,364 @@
...
+    public RenderFragment? ChildContent { get; set; }
+
+    /// <summary>
+    /// Gets or sets the parent <see cref="FluentNav"/> component for this instance.
+    /// </summary>
+    /// <remarks>This property is typically set automatically by the Blazor framework when the component is
+    /// used within a <see cref="FluentNav"/>. It enables the component to access shared state or functionality from
+    /// its parent navigation menu.</remarks>
+    [CascadingParameter]
+    public required FluentNav Owner { get; set; }
```
dvoituron: Can you set `internal` for all `CascadingParameter]? This property will be set only because the component will be included inside a parent.

```csharp
[CascadingParameter]
internal FluentNav Owner { get; set; } = default!;

[Inject]
internal NavigationManager NavigationManager { get; set; } = default!;

[CascadingParameter(Name = "Category")]
internal FluentNavCategory? Category { get; set; }
```

src/Core/Components/Nav/FluentNavItem.razor.cs
```diff
@@ -0,0 +1,364 @@
...
+
+    /// <summary />
+    protected string? StyleValue => DefaultStyleBuilder
+        .Build();
+
+    /// <summary>
+    /// Gets or sets the icon of the nav menu item.
+    /// </summary>
+    [Parameter]
+    public Icon? Icon { get; set; }
```
dvoituron, line 56: I think you should have two parameters: an `Icon` parameter and an `IconSelected` parameter. 
1. You will encounter issues with the `GetActiveIcon` feature and the Trimming WASM feature.
2. You don't always have a filled icon for the `similar` regular icon.
3. You should be able to select a different icon or color when an item is selected.

src/Core/Components/Nav/FluentNavItem.razor
```diff
@@ -0,0 +1,61 @@
...
+			rel="@(!string.IsNullOrWhiteSpace(Target) ? "noopener noreferrer" : string.Empty)"
+			@onclick="OnClickHandlerAsync"
+			@attributes="@AdditionalAttributes">
+		@RenderIcon
+		@ChildContent
+	</button>
+}
+
+@code {
+	private RenderFragment RenderIcon => __builder =>
```
dvoituron: Can you move this method to the code-behind?
See `AddTag.BuildRenderTree` method, for an example

src/Core/Components/Nav/FluentNav.razor.cs
```diff
@@ -0,0 +1,257 @@
...
+    /// Gets or sets whether to allow just one expanded category or multiple
+    /// </summary>
+    [Parameter]
+    public bool UseSingleExpanded { get; set; } = false;
+
+    /// <summary>
+    /// Gets or sets the density of the nav menu item.
+    /// </summary>
+    [Parameter]
+    public NavDensity? Density { get; set; }
```
dvoituron, line 100: The `Density` is a Nav parameter, so the `density` HTML attribute must be on this `.fluent-nav` element (and not on ALL **navitems**. The single difference will be an update in the CSS.

src/Core/Components/Nav/FluentNav.razor.css
```diff
@@ -0,0 +1,156 @@
...
+
+  [dir="rtl"] .fluent-nav-container.collapsed {
+    transform: translate3d(var(--nav-width), 0, 0);
+  }
+
+.fluent-nav {
+  overflow: hidden;
+  max-width: 100vw;
+  height: auto;
+  max-height: 100vh;
```
dvoituron, line 30: I don't think this max-height must be defined: this limit must be defined by the container (or not defined if the dev prefers to display a very long navigation panel)

src/Core/Components/Nav/FluentNav.razor
```diff
@@ -0,0 +1,36 @@
...
+			{
+				<header role="none" class="header">
+				<FluentButton Appearance="@ButtonAppearance.Transparent"
+							  IconStart="@ToggleIcon"
+							  Title="@Localizer[Localization.LanguageResource.Nav_ToggleTitle]"
+							  IconOnly="true"
+							  OnClick="@ToggleNavAsync"/>
+				</header>
+			}
+			<div class="body">
```
dvoituron: `body` is too common a name. There is too much risk that a class already has this name in the dev project => `nav-body`?

src/Core/Components/Nav/FluentNav.razor
```diff
@@ -0,0 +1,36 @@
...
+<CascadingValue Value="this" IsFixed="true">
+	<div class="fluent-nav-container@(_navOpen ? "" : " collapsed")">
+		<div id="@Id"
+			role="navigation"
+			class="@ClassValue"
+			style="@StyleValue"
+			@attributes="@AdditionalAttributes">
+			@if (UseHeader)
+			{
+				<header role="none" class="header">
```
dvoituron: Same as `body` class below, `header` is too common a name.

And as already discussed. I think this header can be removed at this moment: the FluentLayout already manage this feature.

src/Core/Components/Nav/FluentNavDivider.razor.cs
```diff
@@ -0,0 +1,35 @@
...
+public partial class FluentNavDivider
+{
+    /// <summary>
+    /// Gets or sets the parent <see cref="FluentNav"/> component for this instance.
+    /// </summary>
+    /// <remarks>This property is typically set automatically by the Blazor framework when the component is
+    /// used within a <see cref="FluentNav"/>. It enables the component to access shared state or functionality from
+    /// its parent navigation menu.</remarks>
+    [CascadingParameter]
+    public required FluentNav Owner { get; set; }
```
dvoituron: Can you use `internal required FluentNav Owner { get; set; } = default!` ?

src/Core/Components/Nav/FluentNavDivider.razor.cs
```diff
@@ -0,0 +1,35 @@
...
+
+    /// <summary>
+    /// Validates that this component is used within a FluentNav.
+    /// </summary>
+    protected override void OnParametersSet()
+    {
```
dvoituron: This test can be moved to the OnInitialized.
Try to avoid to use `OnParametersSet`

src/Core/Components/Nav/FluentNavDivider.razor.css
```diff
@@ -0,0 +1,3 @@
+.fluent-nav .fluent-navdivider {
+    margin: 4px 0;
```
dvoituron: Why not to use `--spacingVerticalXS`?

src/Core/Components/Nav/FluentNavCategory.razor.cs
```diff
@@ -0,0 +1,310 @@
...
+    /// </summary>
+    protected override void OnInitialized()
+    {
+        Owner?.RegisterCategory(this);
+    }
+
+    /// <summary>
+    /// Disposes of the component and unregisters from the owner.
+    /// </summary>
+    public void Dispose()
```
dvoituron: The `FluentComponentBase` already contains an overridable `DisposeAsync` method. You can use it. Example: `FluentOption.DisposeAsync`

src/Core/Components/Nav/FluentNavDivider.razor.cs
```diff
@@ -0,0 +1,35 @@
...
+// ------------------------------------------------------------------------
+
+using Microsoft.AspNetCore.Components;
+
+namespace Microsoft.FluentUI.AspNetCore.Components;
+
+/// <summary>
+/// A divider for use within a <see cref="FluentNav"/>
+/// </summary>
+public partial class FluentNavDivider
```
dvoituron: All Fluent components must inherit from `FluentComponentBase`. In this case, I think you could skip creating it and just use `FluentDivider` directly in the examples. The result is the same, and there's no need to redevelop it.

src/Core/Components/Nav/FluentNavCategory.razor.cs
```diff
@@ -0,0 +1,310 @@
...
+
+namespace Microsoft.FluentUI.AspNetCore.Components;
+
+/// <summary>
+/// Represents a group of related navigation items for use in a fluent navigation interface.
+/// </summary>
+/// <remarks>Use this class to organize navigation elements into logical groups when building fluent or
+/// hierarchical navigation structures. Grouping navigation items can improve usability and clarity in user interfaces
+/// that support complex navigation scenarios.</remarks>
+public partial class FluentNavCategory : FluentComponentBase, IDisposable
```
dvoituron: See my other similar comment: `FluentComponentBase` already implements `IDisposableAsync`.

src/Core/Components/Nav/FluentNavSectionHeader.razor.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// <remarks>This property is typically set automatically by the Blazor framework when the component is
+    /// used within a <see cref="FluentNav"/>. It enables the component to access shared state or functionality from
+    /// its parent navigation menu.</remarks>
+    [CascadingParameter]
+    public required FluentNav Owner { get; set; }
+
+    /// <summary>
+    /// Validates that this component is used within a FluentNav.
+    /// </summary>
+    protected override void OnParametersSet()
```
dvoituron: You can test that in the `OnInitialized` method

src/Core/Components/Nav/FluentNavItem.razor.cs
MarvinKlein1508: @dvoituron consider making them `protected` instead so you can access them in derived components.

MarvinKlein1508:
@vnbaaij have you considered #2473 & #2142 for this as well?

vnbaaij:
> @vnbaaij have you considered #2473 & #2142 for this as well?

The Nav will expand to the item navigated to, yes

src/Core/Components/Nav/FluentNavItem.razor.cs
vnbaaij: Yes, I agree. I will do that in part 3 (together with the toggle stuff)

src/Core/Components/Nav/FluentNavItem.razor.cs
vnbaaij: When they are made internal you get error: required member 'Owner' cannot be less visible or have a setter less visible than the containing type

src/Core/Components/Nav/FluentNav.razor.css
vnbaaij: This is needed so that height of the nav part can be 'contained' and the nav in total does not grow larger than the available screen height

Without the max-height:
image

With the max-height:
image

(notice the scroll bar on the nav

src/Core/Components/Nav/FluentNav.razor.cs
vnbaaij: Done

dvoituron (changes requested):

src/Core/Components/Nav/FluentNavDivider.razor.cs
vnbaaij: CascadingParameter cannot be made internal. Will give errors like 'Required member 'Owner' cannot be less visible or have a setter less visible than the containing type'

src/Core/Components/Nav/FluentNavDivider.razor.cs
vnbaaij: Done

src/Core/Components/Nav/FluentNavDivider.razor.css
vnbaaij: Done

src/Core/Components/Nav/FluentNavCategory.razor.cs
vnbaaij: Done

src/Core/Components/Nav/FluentNavDivider.razor.cs
vnbaaij: This is just an 'override' for FluentDivider so the correct styling (when used in nav) will be applied. Also, because we assing a class, this allows for the dev to overrule the styling when they want/need to do that.
Becasue it just renders a FluentDivider, this component itself does not have to implement FluentComponentBase.

src/Core/Components/Nav/FluentNavCategory.razor.cs
vnbaaij: Done

src/Core/Components/Nav/FluentNavSectionHeader.razor.cs
vnbaaij: Done

src/Core/Components/Nav/FluentNavDivider.razor.cs
vnbaaij: Fixed with later commit

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Nav/FluentNav.md
```diff
@@ -0,0 +1,60 @@
...
+In the example below, the first (and third and fifth) item has an `Href` that navigates to an external site, while the second (and fourth) item has an `OnClick` handler that triggers a method in the component.
+
+{{ NavDefault }}
+
+## API FluentNav
+
+{{ API Type=FluentNav }}
+
+## API FluentNavCategory
+{{ API Type=FluentNavCategory }}
```
MarvinKlein1508, line 50: We should put another line break here as we always have 2 in v5

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Nav/Examples/NavDefault.razor
```diff
@@ -0,0 +1,108 @@
...
+			<FluentButton OnClick="ExpandRetirementAsync">Expand 'Retirement'</FluentButton>
+			<FluentButton OnClick="CollapseRetirementAsync">Collapse 'Retirement'</FluentButton>
+		</span>
+	</FluentStack>
+
+</FluentStack>
+
+@code
+{
+	private FluentNav? nav;
```
MarvinKlein1508: should we do `private FluentNav nav = default!;` instead of it beaing nullable. This way we don't have to check suppress the null warning everytime

src/Core/Components/Nav/FluentNav.razor.cs
```diff
@@ -0,0 +1,257 @@
...
+    /// </summary>
+    [Parameter]
+    public string AppLink { get; set; } = "/";
+
+    /// <summary>
+    /// Gets or sets whether to use the header with the hamburger icon.
+    /// Defaults to false until we have a good way to make this work with the LayoutHamburger component.
+    /// </summary>
+    [Parameter]
+    public bool UseHeader { get; set; } = false;
```
MarvinKlein1508: `false` is the default value and don't need to be set

src/Core/Components/Nav/FluentNav.razor.cs
```diff
@@ -0,0 +1,257 @@
...
+    /// Gets or sets whether to enable using icons in the nav items.
+    /// </summary>
+    [Parameter]
+    public bool UseIcons { get; set; } = true;
+
+    /// <summary>
+    /// Gets or sets whether to allow just one expanded category or multiple
+    /// </summary>
+    [Parameter]
+    public bool UseSingleExpanded { get; set; } = false;
```
MarvinKlein1508: `false` is the default value and don't need to be set

src/Core/Components/Nav/FluentNavCategory.razor.cs
```diff
@@ -0,0 +1,303 @@
...
+
+namespace Microsoft.FluentUI.AspNetCore.Components;
+
+/// <summary>
+/// Represents a group of related navigation items for use in a fluent navigation interface.
+/// </summary>
+/// <remarks>Use this class to organize navigation elements into logical groups when building fluent or
+/// hierarchical navigation structures. Grouping navigation items can improve usability and clarity in user interfaces
+/// that support complex navigation scenarios.</remarks>
+public partial class FluentNavCategory : FluentComponentBase, IAsyncDisposable
```
MarvinKlein1508: Since you inherit from `FluentComponentBase` you don't need to specify `IAsyncDisposable` here again

src/Core/Components/Nav/FluentNavItem.razor.cs
```diff
@@ -0,0 +1,392 @@
...
+using Microsoft.AspNetCore.Components.Routing;
+using Microsoft.AspNetCore.Components.Web;
+using Microsoft.FluentUI.AspNetCore.Components.Utilities;
+
+namespace Microsoft.FluentUI.AspNetCore.Components;
+
+/// <summary>
+/// Represents a navigation menu item that renders content within a Fluent UI styled navigation link.
+/// </summary>
+public partial class FluentNavItem : FluentComponentBase, IAsyncDisposable
```
MarvinKlein1508: Since you inherit from `FluentComponentBase` you don't need to specify `IAsyncDisposable` here again

MarvinKlein1508 (changes requested):

src/Core/Components/Nav/FluentNavDivider.razor.cs
dvoituron: I don't really see the difference: **FluentDivider** does all that and can be used directly. There are no styles applied except for a “margin". In this case, you can add a style in you FluentNav.razor.css file:
```css
.fluent-nav fluent-divider {
    margin: var(--spacingVerticalXS) 0;
}
```

But in your `FluentNavDivider` component doesn't inherit the `FluentDivider`. And at this moment, the dev cannot apply a custom `Style` or `Class` :-(

src/Core/Components/Nav/FluentNavDivider.razor.cs
dvoituron: Not fixed

src/Core/Components/Nav/FluentNav.razor.css
dvoituron: Yes. I've seen that before. And that's my point: by default, the component should behave like any other element. That is, it should take the place of the content. That's what DIV does, for example. If you want a scrollbar, you have to specify it.

Currently, the dev cannot “not have a scrollbar.” In some cases, I would like the content to extend below the visible part.

Give me your opinion :-)

src/Core/Components/Nav/FluentNavItem.razor.cs
dvoituron: > @dvoituron consider making them `protected` instead so you can access them in derived components.

I don't see in what case you would need to access this property from an inherited class. And this may cause problems for Unit Tests that can access `internal`

@vnbaaij The `Category` property is still `public`

src/Core/Components/Nav/FluentNavDivider.razor.cs
vnbaaij: I like the css solution. Will do that.

The FluentNavDivider not being customizable was intentional :)

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Nav/Examples/NavDefault.razor
vnbaaij: Done

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Nav/FluentNav.md
vnbaaij: Done

src/Core/Components/Nav/FluentNav.razor.cs
vnbaaij: Done

src/Core/Components/Nav/FluentNav.razor.cs
vnbaaij: Done

src/Core/Components/Nav/FluentNav.razor.css
vnbaaij: I agree that that is sometimes what a dev wants but that is not the case we want to have (or make possible) with a nav. I modelled this based on the Fluent Design and React implementation. For now it looks as it should. We can see if we want to change it in the part 3 PR that I am doing later

src/Core/Components/Nav/FluentNavCategory.razor.cs
vnbaaij: Done

src/Core/Components/Nav/FluentNavDivider.razor.cs
vnbaaij: Done

src/Core/Components/Nav/FluentNav.razor.css
dvoituron: I don't like that you say “for part 3.” For me, we can choose not to develop something and do it later, but if we do develop it, it has to be correct :-)

So if we want a scrollbar, why use a `max-height: 100vh` and not `100%`? Or `calc(100vh-<HeaderHeight>-<FooterHeight>)`? Or ??? Do you see the problem? ;-)

src/Core/Components/Nav/FluentNavItem.razor.cs
vnbaaij: Done

src/Core/Components/Nav/FluentNavItem.razor.cs
vnbaaij: Category done!

dvoituron (approved):

PR #4421, opened 2025-12-22
[AppBar] Fix AppBarItem OnClick not being handeled in popover

### LINKED ISSUE #4418: fix: FluentAppBar - Popover - OnClick
Have an AppBar as from documentation, with OnClick execute mode:

Image

All is working OK.

But if you resize available area:

Image

The OnClick event **is not called** when FluentAppBarItem is **inside Popover**.

Make official demo OnClick/AppBar resizeable to test, latest version, NET10.

vnbaaij:
Fix #4418. In popover, an AppItem's OnClick was not being passed through.

### DIFF
--- a/examples/Demo/Shared/Microsoft.FluentUI.AspNetCore.Components.xml
+++ b/examples/Demo/Shared/Microsoft.FluentUI.AspNetCore.Components.xml
@@ -7264,6 +7264,13 @@
             Most important info to be shown in the message bar.
             </summary>
         </member>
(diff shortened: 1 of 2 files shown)


dvoituron (approved):

LucaCris:
The action is now executed, but the panel remains open...

immagine

It should be click&close.

vnbaaij:
@LucaCris not sure if it should _always_ be click and close but we can take a look on how to solve that. 

Please create a new issue for this as we can't do work on closed PRs and issues.

vnbaaij:
After some internal discussion, we agree. Should be click and close. Will create a PR for that.

PR #4455, opened 2026-01-09
[dev-v5][Nav] Replace Icon parameter with IconRest/IconActive

vnbaaij:
Previous implementation used reflection to find a filled variant of a specified icon. This doesn't work together with the icon trimming we have when doing a release build. With this PR it is now possible to specify the rest state icon and the active state icon separately.

- Replace Icon with IconRest and IconActive parameters
- Remove unused code and tests 
- Alter and add new tests

### DIFF
--- a/examples/Demo/FluentUI.Demo.Client/Documentation/Components/Nav/Examples/NavDefault.razor
+++ b/examples/Demo/FluentUI.Demo.Client/Documentation/Components/Nav/Examples/NavDefault.razor
@@ -6,7 +6,7 @@
     <FluentNav @ref="@nav"
                UseIcons="@UseIcons"
                UseSingleExpanded="@(!UseMultipleExpanded)"
-               OnItemClick="@(e => Console.WriteLine("NavItem clicked."))"
+			   OnItemClick="@(e => Console.WriteLine($"NavItem clicked: '{e.IconRest?.Name}'."))"
                Width="260px"
                Density="@Density">
 
--- a/examples/Demo/FluentUI.Demo.Client/Layout/DemoNav.razor
+++ b/examples/Demo/FluentUI.Demo.Client/Layout/DemoNav.razor
@@ -23,16 +23,16 @@
 			@if (!sitePage.Items.Any())
 			{
 				<FluentNavItem Href="@sitePage.Route"
-							   IconRest="@(GetIconFromName(sitePage.Icon, false))"
-							   IconActive="@(GetIconFromName(sitePage.Icon, true))">
+							   IconRest="@(GetIconFromName(sitePage.Icon, isActive: false))"
+							   IconActive="@(GetIconFromName(sitePage.Icon, isActive: true))">
 					@sitePage.Title
 				</FluentNavItem>
 			}
 			else
 			{
 				<FluentNavCategory Title="@sitePage.Title"
-								   IconRest="@(GetIconFromName(sitePage.Icon, false))"
-								   IconActive="@(GetIconFromName(sitePage.Icon, true))">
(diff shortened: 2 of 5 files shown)


examples/Demo/FluentUI.Demo.Client/Documentation/Components/Nav/Examples/NavDefault.razor
```diff
@@ -3,49 +3,48 @@
 @inject IDialogService DialogService
 
 <FluentStack Orientation="Orientation.Horizontal" HorizontalGap="15">
-
     <FluentNav @ref="@nav"
                UseIcons="@UseIcons"
                UseSingleExpanded="@(!UseMultipleExpanded)"
-               OnItemClick="@(e => Console.WriteLine($"NavItem clicked: Icon '{e.Icon?.Name}'."))"
+               OnItemClick="@(e => Console.WriteLine("NavItem clicked."))"
```
dvoituron: You need to display an information to know which item is clicked. Why the Icon name is not valid?

examples/Demo/FluentUI.Demo.Client/Layout/DemoNav.razor
```diff
@@ -2,46 +2,50 @@
...
+		   Density="@NavDensity.Medium">
 
-    <FluentNavItem Href="/"
-                   Match="NavLinkMatch.All"
-                   Icon="@(new Icons.Regular.Size20.Home())">
-        Home
-    </FluentNavItem>
+	<FluentNavItem Href="/"
+				   Match="NavLinkMatch.All"
+				   IconRest="@(new Icons.Regular.Size20.Home())"
```
dvoituron, line 10: Is it possible to use the same Icon, without to replicate the code?
- if IconRest is null => use IconActive
- if IconActive is null => use IconRest

examples/Demo/FluentUI.Demo.Client/Layout/DemoNav.razor
```diff
@@ -2,46 +2,50 @@
...
+				<FluentNavItem Href="@sitePage.Route"
+							   IconRest="@(GetIconFromName(sitePage.Icon, false))"
+							   IconActive="@(GetIconFromName(sitePage.Icon, true))">
+					@sitePage.Title
+				</FluentNavItem>
+			}
+			else
+			{
+				<FluentNavCategory Title="@sitePage.Title"
+								   IconRest="@(GetIconFromName(sitePage.Icon, false))"
```
dvoituron: Please, add `isActive: false` to have a clearer code (always interesting for boolean arguments).

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Nav/Examples/NavDefault.razor
vnbaaij: I'm adding it back.

examples/Demo/FluentUI.Demo.Client/Layout/DemoNav.razor
vnbaaij: Not sure what you mean.
Rest icon is the default icon. If not set then do not display an icon (not in rest nor in active state). If an active icon is set and the item is active then use the active icon. Fall back to the rest icon if no active icon set

dvoituron: Oh... My bad. I thought the IconRest and IconActive were the same values.
This if correct.

dvoituron (approved):

Issue #4295, opened 2025-11-07
[dev-v5] fix: selection of custom items in FluentSelect


MarvinKlein1508:
# 🐛 Bug Report

Our Customize the items for FluentSelect in v5 demo is broken.

When loading the page the selected item is not shown in the component itself.

<img width="227" height="199" alt="Image" src="https://github.com/user-attachments/assets/0ead4eed-0ec8-480c-9773-992621af3064" />

When you select a value then the wrong one will be displayed.

<img width="194" height="128" alt="Image" src="https://github.com/user-attachments/assets/b3b69825-744b-4934-ab85-03342ed4f044" />

Please assign me in the PR for this as well. I'd like to understand the root of this issue.

## 💻 Repro or Code Sample
https://fluentui-blazor-v5.azurewebsites.net/List/Select

## 🤔 Expected Behavior

I expect the item to be displayed right away and the correct item when I select one.

## 😯 Current Behavior

No item will be displayed on initialization and the wrong one once you select an item.

## 🔦 Context

Perhaps the underlying webcomponent has been changed? I've noticed that the component uses
```
protected virtual string DropdownType => "dropdown";
```

although it is a select.

The official component is divided into three different types:
https://storybooks.fluentui.dev/web-components/?path=/docs/components-dropdown--docs

<img width="1200" height="107" alt="Image" src="https://github.com/user-attachments/assets/2d9419f0-bb84-433f-bf83-f8ce9b1f2dc1" />

PR #4456, opened 2026-01-09
[dev-v5] Add a generic FluentOption<TOption> 

dvoituron:
# [dev-v5] Add a generic FluentOption<TOption> 

This pull request introduces a generic `FluentOption<TOption>` component, enabling type-safe options for FluentSelect, FluentCombobox, and related components, and updates supporting infrastructure and samples accordingly.

Example:
```razor
<!-- Using Items -->
<FluentSelect Label="Language"
              Items="@Languages"
              @bind-Value="@SelectedId1"
              OptionText="@(i => i.Name)"
              OptionValue="@(i => i.Id)" />

<!-- Using FluentOptions -->
<FluentSelect Label="Language"
              TOption="Language"
              TValue="int?"
              @bind-Value="@SelectedId2">
    <FluentOption TValue="int?" Value="@(null)" Text="[Select]" />
    @foreach (var language in Languages)
    {
        <FluentOption TValue="int?" Value="@language.Id">
            @language.Name
        </FluentOption>
    }
</FluentSelect>

<div>
    IDs: (@SelectedId1, @SelectedId2)
</div>

@code {

    int SelectedId1 = 3;
    int? SelectedId2 = 3;

    static Language[] Languages = new Language[]
    {
        new(1, "C#"),
        new(2, "Java"),
        new(3, "Python"),
        new(4, "JavaScript"),
        new(5, "Go")
    };

    record Language(int Id, string Name);
}
```

## Unit Tests

`100%`: Updated

### DIFF
--- a/examples/Demo/FluentUI.Demo.Client/Documentation/Components/List/Select/FluentSelect.md
+++ b/examples/Demo/FluentUI.Demo.Client/Documentation/Components/List/Select/FluentSelect.md
@@ -93,6 +93,10 @@ Demonstrates two-way data binding with `FluentSelect` using `@bind-Value` to bin
 
 {{ SelectLanguage }}
 
+**Note:** If you programmatically change the value of a manually constructed list (using `FluentOption`),
+you will likely need to add an attribute `@key=“@SelectedId”` to force Blazor to redraw the component correctly.
+This is the case in the above example: `@key="@(SelectedId2 ?? 0)"`.
+
 ## Many items
 
 You can use this component with a large number of items.
--- a/tests/Core/Components/List/FluentComboboxTests.razor
+++ b/tests/Core/Components/List/FluentComboboxTests.razor
@@ -281,36 +281,4 @@
         Assert.Equal(expectedParsedValue, parsedValue);
         Assert.Equal(expectedValidationErrorMessage, validationErrorMessage);
     }
-
-//     [Theory]
-//     [InlineData("InvalidId")]
-//     [InlineData(null)]
-//     [InlineData("FIRST_EXISTING_ID")]
-//     public void FluentCombobox_RemoveUnknownOption(string? id)
-//     {
-// #pragma warning disable BL0005
-
-//         // Arrange
-//         var cut = Render(@<FluentCombobox Items="@Digits" TOption="string" TValue="string" />);
-//         var ids = cut.FindAll("fluent-option").Select(i => i.GetAttribute("id"));
-//         var data = (object?)null;
-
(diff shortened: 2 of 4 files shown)


MarvinKlein1508:
@dvoituron the current code works with non nullable properties. However selection via code does not work properly yet.

Add 

```razor
<FluentButton OnClick="() => SelectedLanguageId = 1">Select C#</FluentButton>
```

to the sample and then hit it. The value changes but not the displayed one on the dropdown.

Using null as option breaks the component. See this code for reference:
```razor
<FluentSelect Label="Language" @bind-Value="@SelectedLanguageId">
    <FluentOption TOption="int?" Value="null"></FluentOption>
    @foreach (var language in Languages)
    {
        <FluentOption Value="@language.Id">
            @language.Name
        </FluentOption>
    }
</FluentSelect>

<div>
    Selected: @(SelectedLanguageId?.ToString() ?? "<NULL>")
</div>

@code {

    int? SelectedLanguageId;

    static Language[] Languages = new Language[]
    {
        new(1, "C#"),
        new(2, "Java"),
        new(3, "Python"),
        new(4, "JavaScript"),
        new(5, "Go")
    };

    record Language(int Id, string Name);
}
```

dvoituron:
> @dvoituron the current code works with non nullable properties. However selection via code does not work properly yet.
> Using null as option breaks the component. See this code for reference:

The `Value` is always nullable, because it's possible to "select" nothing (that was already the case before).
You don't need to specify that your type is nullable.

MarvinKlein1508:
> The `Value` is always nullable, because it's possible to "select" nothing (that was already the case before). You don't need to specify that your type is nullable.

When I want to provide a manual item then I have to.
```razor
<FluentOption TOption="int?" Value="null"></FluentOption>
```

This item should mimic the default state and allows the user to unselect a previous selected option.

MarvinKlein1508:
btw. these changes also fix #4295 for me

MarvinKlein1508:
@dvoituron there is still an issue when setting the second language example via code.
```razor
<FluentButton OnClick="() => SelectedId1 = 1">Set first to C#</FluentButton>
<FluentButton OnClick="() => SelectedId2 = 1">Set second to C#</FluentButton>
<FluentButton OnClick="() => SelectedId2 = null">Set second to null</FluentButton>
```

First button works fine. The other two are changing the value but the dropdown will still display the wrong one.

dvoituron:
> @dvoituron there is still an issue when setting the second language example via code.

No. This is a Blazor behavior: you need to add ’@key’ to fix that. I added  this info in the markdown page (below the example)

PR #4472, opened 2026-01-20
Add FluentValidationMessage

vnbaaij:
With this PR you can use `FluentValidationMessage` to display the validation message like the standard Blazor `ValidationMesage`
Also, the `FluentField` has been adapted so the validation messages will be displayed in the message part of the field.

Unit Test need to be validated still


MarvinKlein1508:
@vnbaaij what is the state of this PR? If I remember correctly then something needs to be done in the web components first, am I right?

vnbaaij:
Yes, that is correct. Or, at least, I need to get some info from the web components people

PR #4477, opened 2026-01-22
[dev-v5] More Unit Tests

vnbaaij:
Bring more components to 100% (or at least above 98%)coverage by adding additional tests and enhancing existing tests:

- FluentCheckBox (99.1%)
- FluentMultiSplitterPane (100%)
- FluentTabs (100%)
- FluentErrorBoundary (100%)

### DIFF
--- a/src/Tools/McpServer/Extensions/ServiceCollectionExtensions.cs
+++ b/src/Tools/McpServer/Extensions/ServiceCollectionExtensions.cs
@@ -42,6 +42,10 @@ internal static class ServiceCollectionExtensions
         // Falls back to embedded resource if no external file is found
         services.AddSingleton(_ => new FluentUIDocumentationService(externalJsonPath));
 
+        // Documentation service
+        // Excludes the 'mcp' folder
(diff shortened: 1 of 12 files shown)


src/Core/Components/ErrorBoundary/FluentErrorBoundary.razor.cs
```diff
@@ -75,13 +75,7 @@ public FluentErrorBoundary(LibraryConfiguration configuration) : base(configurat
     /// </summary>
     public void Recover()
     {
-        try
```
dvoituron, line 78: Why did you remove that?
I've already encountered several cases where recovery fails, and it's not a good idea to throw an exception in that case.

dvoituron (changes requested):

src/Core/Components/ErrorBoundary/FluentErrorBoundary.razor.cs
vnbaaij: I couldn't get it to throw in a Unit Test. But will revert and then we'll be a bit under 100% for this one. Can live with that.

Issue #4495, opened 2026-01-29
fix: FluentDatePicker- Didnt ignore or format date 01012000


SharmaAishna:
# 🐛 Bug Report
While traversing by Tab key  or directly selecting DateInput and user fills the date. The user typed 01012000 this doesn't get formatted . It should get clean up like other date format or converted to correct date format . 

https://github.com/[SharmaAishna/FluentDatePickerSample](https://github.com/SharmaAishna/FluentDatePickerSample)

## 🤔 Expected Behavior

Like other dates are being type they are formatted based on localization . This format should be clean

## 😯 Current Behavior

traversing to other fields the date doesn't get rejected or formatted . No event method can hit the issue. 

## 🌍 Your Environment

Visual Studio 2026 professional . 
Fluent 4.0
Blazor 9.0
Browser Microsoft Edge, Chrome has  same issue

Tyme-Bleyaert:
There is currently a bug where parsing failures do not surface a validation error on the field.
I’ll fix this.

Setting that aside, I’m not convinced we should allow a default parsing fallback for date inputs such as:
- ddMMyyyy
- MMddyyyy
- yyyyMMdd

We could derive the expected format from `CultureInfo.DateTimeFormat.ShortDatePattern` and build a culture-specific numeric pattern, but the open question is whether this behavior should be implicit at all, or whether we should instead expose an explicit way for consumers to define supported formats for the component.

Implicit fallback parsing increases convenience, but it also introduces ambiguity and makes failures harder to reason about, especially with purely numeric inputs that are valid under multiple formats. From an API design perspective, this pushes hidden behavior into what is otherwise a deterministic component.

For v4, I don’t see us implementing this. But it could be worth revisiting for v5.

vnbaaij:
We should definitely not do an implicit behavior and make the assumption that that is what a consumer wants. (For example I use English language and region for my OS but date format is set to Dutch...)
Adding a way to set the format explicitly is a good addition.

SharmaAishna:
Thank you for looking into this issue and for the detailed explanation around the parsing behavior and design considerations.

I’ll wait for the fix in the upcoming release and look forward for v5 release. Thanks again for the quick response and ongoing work on this.

PR #4496, opened 2026-01-29
[DatePicker] Prevent silently ignoring date formatting failures

Tyme-Bleyaert:
# Pull Request

## 📖 Description

This pr will prevent the silent ignore of failed formatting, so the user can correct the input when they type `02022000` instead of `02/02/2000`.

### 🎫 Issues

#4495 

### General

### Component-specific


src/Core/Components/DateTime/FluentCalendar.razor.cs
```diff
@@ -293,9 +293,9 @@ private async Task PickerYearSelectAsync(DateTime? year)
     /// <summary />
     protected override bool TryParseValueFromString(string? value, out DateTime? result, [NotNullWhen(false)] out string? validationErrorMessage)
     {
-        BindConverter.TryConvertTo(value, Culture, out result);
-        validationErrorMessage = null;
-        return true;
+        bool success = BindConverter.TryConvertTo(value, Culture, out result);
+        validationErrorMessage = success? null : $"The {DisplayName ?? (FieldBound ? FieldIdentifier.FieldName : UnknownBoundField)} field is not in a valid format.";
```
dvoituron: The message need to be translated for non English people

Check the other components, but maybe using
```csharp
    /// <summary>
    /// Gets or sets the error message to show when the field can not be parsed.
    /// </summary>
    [Parameter]
    public string ParsingErrorMessage { get; set; } = "The {0} field must be a valid format.";
```

src/Core/Components/Base/FluentInputBase.cs
```diff
@@ -132,6 +132,12 @@ public abstract partial class FluentInputBase<TValue> : FluentComponentBase, IDi
     [Parameter]
     public virtual bool Embedded { get; set; } = false;
 
+    /// <summary>
+    /// Gets or sets the error message to show when the field can not be parsed.
+    /// </summary>
+    [Parameter]
+    public virtual string ParsingErrorMessage { get; set; } = "The {0} field must be a valid format.";
```
vnbaaij: ... must have a valid format

src/Core/Components/Base/FluentInputBase.cs
```diff
@@ -132,6 +132,12 @@ public abstract partial class FluentInputBase<TValue> : FluentComponentBase, IDi
     [Parameter]
     public virtual bool Embedded { get; set; } = false;
 
+    /// <summary>
+    /// Gets or sets the error message to show when the field can not be parsed.
+    /// </summary>
+    [Parameter]
+    public virtual string ParsingErrorMessage { get; set; } = "The {0} field must have a valid format.";
```
dvoituron: 😱 I don't think it's a good idea to add a parameter to `InputBase`. Because that means you have to manage it for all child components. There are apparently more than 15 of them.

Image

dvoituron (changes requested):

src/Core/Components/Base/FluentInputBase.cs
Tyme-Bleyaert: Every derived component already must deal with parsing errors via:
```csharp
/// <summary>
/// Parses a string to create an instance of <typeparamref name="TValue"/>. Derived classes can override this to change how
/// <see cref="CurrentValueAsString"/> interprets incoming values.
/// </summary>
/// <param name="value">The string value to be parsed.</param>
/// <param name="result">An instance of <typeparamref name="TValue"/>.</param>
/// <param name="validationErrorMessage">If the value could not be parsed, provides a validation error message.</param>
/// <returns>True if the value could be parsed; otherwise false.</returns>
protected abstract bool TryParseValueFromString(string? value, [MaybeNullWhen(false)] out TValue result, [NotNullWhen(false)] out string? validationErrorMessage);
```
So the claim “you now have to manage it for all child components” is slightly misleading. They already do, just in a fragmented way.

Meanwhile this parameter:

- is optional
- has a sane default
- does not require overrides
- does not introduce breaking changes

And before this, we got hardcoded text sitting in either `FluentInputExtensions.TryParseSelectableValueFromString()` .
Or in any derived component where we had the formatting actually implemented.

However, we could implement interfaces such as `IParsableComponent` and `ICultureSensitiveComponent` in relevant derived components and move the parsing logic there.
And move the whole parsing part to `IParsableComponent`
This way we can use simple checks in an extension method eg:
```csharp
public static bool TryParseValueFromString<TInput, [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] TValue>(
    this TInput input, string? value,
    [MaybeNullWhen(false)] out TValue result,
    [NotNullWhen(false)] out string? validationErrorMessage) where TInput : FluentInputBase<TValue>, IParsableComponent
{
    try
    {
        // We special-case bool values because BindConverter reserves bool conversion for conditional attributes.
        if ((typeof(TValue) == typeof(bool) && TryConvertToBool(value, out result)) ||
            (typeof(TValue) == typeof(bool?) && TryConvertToNullableBool(value, out result)))
        {
            validationErrorMessage = null;
            return true;
        }

        var culture =
            input is ICultureSensitiveComponent c
                ? c.Culture
                : CultureInfo.CurrentCulture;

        if (BindConverter.TryConvertTo<TValue>(value, culture, out var parsedValue))
        {
            result = parsedValue;
            validationErrorMessage = null;
            return true;
        }

        result = default;
        validationErrorMessage = string.Format(input.ParsingErrorMessage, input.FieldDisplayName);
        return false;
    }
    catch (InvalidOperationException ex)
    {
        throw new InvalidOperationException($"{input.GetType()} does not support the type '{typeof(TValue)}'.", ex);
    }
}
```
This approach would require additional work and testing. However, since this PR targets v4 and not the upcoming v5, which is where we’re planning bigger changes, I don’t see a strong benefit to implementing it right now.

vnbaaij: No, We should not do that for v4. Having a sane default is enough, I think.

PR #4507, opened 2026-02-05
[dev-v5] Centralize .NET version management

dvoituron:
# Centralize .NET version management

Introduce a `NetVersion` property in `Directory.Build.props` to manage .NET versioning across project files. Update project files to reference this property, ensuring consistency and simplifying future updates. This change enhances maintainability and reduces the risk of version mismatches.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,8 +2,17 @@
   <PropertyGroup>
 
     <!-- .NET Framework Version -->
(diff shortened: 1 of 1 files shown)


vnbaaij (approved):

dvoituron:
> Now that we have the publish bug resolved with the workaround in the csproj file, should we not default to net10.0?

Yes, we can do that. I will do that in another PR.

PR #4523, opened 2026-02-13
[dev-v5] Preparation for public RC packages

dvoituron:
# [dev-v5] Preparation for public RC packages

- Update .NET versions and package dependencies for multi-targeting support
- Add support for deterministic builds in CI and Release configurations
- Update package versions
- Update .NET version and adjust multi-targeting settings; change build mode to Debug


Directory.Build.props
```diff
@@ -12,7 +11,13 @@
          - examples\Tools\FluentUI.Demo.DocApiGen.IntegrationTests\FluentUIComponentsIntegrationTests.cs
          - tests\Integration\WebServer\StartServerFixture.cs
     -->
+    <NetVersion>net9.0</NetVersion>
 
+    <!-- Target .NET versions for multi-targeting -->
+    <!-- Used with the published libraries: Core; McpServer -->
+    <!-- Debug: single target for faster builds; Release: full multi-targeting -->
+    <TargetNetVersions Condition="'$(Configuration)' == 'Release'">net9.0;net10.0</TargetNetVersions>
+    <TargetNetVersions Condition="'$(Configuration)' != 'Release'">$(NetVersion)</TargetNetVersions>
```
vnbaaij, line 20: Guidance is to not use TargetNetVersion**s** when there is only one version specified

vnbaaij (changes requested):

Directory.Build.props
dvoituron: I know, but it's the only way I've found to avoid compiling in all versions of dotnet when developing locally (in Debug mode). It reduces compilation time by 50% (or even more) and is not used on Build servers (ADO and Github).

I also deliberately put net9 (and not 10) to ensure that we were developing under the worst possible conditions.

If you need to develop later, or test locally in multiple versions, just put `net9.0;net10.0` and that's it.

vnbaaij (approved):

PR #4542, opened 2026-02-19
[dev-v5] [DatePicker] Forward Width to FluentTextInput

MarvinKlein1508:
# Pull Request

## 📖 Description

I came accross this one during development. The `FluentDatePicker` is doing nothing with the value `Width`. This PR addresses this and forwards this values down to `FluentTextInput`.

Before when you add `Width="100%"`:
before

After when you add `Width="100%"`:
after

### General

### Component-specific

## ⏭ Next Steps

The `FluentTimePicker` is suffering from the same issue. Fixing it for native is pretty easy but when `IsFluentUIStyle` is true it isn't working properly yet.


src/Core/Components/DateTime/FluentDatePicker.razor
```diff
@@ -26,6 +26,9 @@
                  MessageState="@MessageState"
                  MessageTemplate="@MessageTemplate"
                  Placeholder="@GetPlaceholderAccordingToView()"
+                 Width="@Width"
+                 Padding="@Padding"
```
dvoituron: OK for `Width` but why for `Padding` and `Margin`?

dvoituron (changes requested):

src/Core/Components/DateTime/FluentDatePicker.razor
MarvinKlein1508: because you can specify them on the `FluentTextInput` as well. This allows the developer to move the field inside.

dvoituron: The Width is a special case because the value is assigned to the FluentTextInput and the sub-component. But the Padding and Margin should be included in the existing StyleValue. So why did you add that? What was your issue about Margin/Padding?

dvoituron (approved):

PR #4558, opened 2026-02-22
[dev-v5] Fix List Widths

dvoituron:
# [dev-v5] Fix List Widths

- Add a Debug page to test and validate all cases
- Fix the List Width
- Fix the FluentField.LabelStyle

image


vnbaaij (approved):

MarvinKlein1508:
@dvoituron for me it is still not clear when I even should use `LabelWidth`. In which cases do you want a different `Width` for the label which differs from `Width` ?

dvoituron:
`LabelWidth` is used only for the **Field label** 😆 this is interesting when you set the label position to `Before` or `After`, or if the label content is too long vs the TextInput.

Issue #1050, opened 2023-12-01
fix: double submit with Enter


iotalambda:
# 🐛 Bug Report

When navigating to an EditForm's submit button by keyboard and hitting Enter, submit occurs twice.

## 💻 Repro or Code Sample

`dotnet new install Microsoft.FluentUI.AspNetCore.Templates::4.1.1`
`dotnet new fluentblazor`

Replace Counter.razor content with the following, tab your way to the FluentButton and hit Enter -> current count is incremented by 2.
```razor
@page "/counter"
@rendermode InteractiveServer

<PageTitle>Counter</PageTitle>

<h1>Counter</h1>

<p role="status">Current count: <FluentBadge Appearance="Appearance.Neutral">@currentCount</FluentBadge></p>

<EditForm OnValidSubmit=@IncrementCount Model="new { }">
    <button type="submit">Submit with normal button -> +1</button>
    <FluentButton Appearance="Appearance.Accent" Type=@ButtonType.Submit>Submit with FluentButton -> +2</FluentButton>
</EditForm>

@code {
    private int currentCount = 0;

    private void IncrementCount()
    {
        currentCount++;
    }
}
```

## 🌍 Your Environment

Occurs at least in Edge and Firefox

vnbaaij:
I just tried it with same template and your exact code. Not seeing this behavior:
![counter-#1050](https://github.com/microsoft/fluentui-blazor/assets/1761079/b98180da-396d-4f52-be0a-75fe72db5332)

iotalambda:
[...]

Issue #2726, opened 2024-09-27
fix: Hitting enter doesn't submit in FluentCombobox


JamesNK:
# 🐛 Bug Report

Hitting enter key when focus is inside a FluentCombobox doesn't submit the form.

## 🤔 Expected Behavior

I would expect that hitting enter would close the combobox if it is open (selecting a value, if a value is select).
I would then expect that hitting enter again (while it is closed) would submit the form.

## 😯 Current Behavior

Nothing happens.

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* .NET and Fluent UI Blazor library Version [e.g. 8.0.2 and 4.4.1]

vnbaaij:
There are some known issues with input  (TextField, TextArea, etc) web components and enter key behavior. See also #1050 and #1085. As the Combobox uses a Fluent text field internally, I can imagine that this component is affected as well.

I'll do some more investigating.

vnbaaij:
This behavior is referred to as implicit form submission: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#implicit-submission
 
The good news is this is built into the `<fluent-text-input>` v3 component. And the team said they will make sure to do the same for combobox.
The bad news is it won't get fixed here...

vnbaaij:
Closing this as it is fixed in v5

PR #4574, opened 2026-02-24
[dev-v5] Update debug page for EditForm

MarvinKlein1508:
This PR is updating the debug page for `EditForm`.

It adds two debug toggles to switch between native rendering for inputs and buttons. In addition it adds a test model which can be expanded with more properties for testing. Right now the debug page is showcasing an issue which we thought has been fixed already #1050 but seems to be broken again.


MarvinKlein1508:
I'm also planning on adding more test cases from common issues here as well. For example #2726

dvoituron (approved):

vnbaaij (approved):

PR #4588, opened 2026-02-28
[dev-v5][DataGrid] width issues

### LINKED ISSUE #4582: [dev-v5] Multiple issues with `FluentDataGrid` when using `Width` property
# 🐛 Bug Report

I found multiple issues when using `Width` parameter on `PropertyColumn` and `TemplateColumn` within `FluentDataGrid`

## 💻 Repro or Code Sample
I created this small test repository
https://github.com/MarvinKlein1508/FluentUIv5-datagrid-issues

### Header not applying width twice

1. Run the project
2. Enter 10px in `width column 1`
3. Press enter
4. Change value to 50px
5. Press enter
6. See that the column will be updated but not the column header

I found out that this is happening because of the `width:100%`

Image

### Additional space when sortable is enabled

1. Run the project
2. switch to table mode
3. Enter 50px for `column width 1`
4. Press enter
5. Enable sorting
6. See that the column is not 50px in size. Instead it is applying the `min-width` of `96px` from the `:host` of the `FluentButton` in the column header

Image

### Wrong sizing

1. Run the project
2. Enter 50px for `column width 1`
3. Confirm in dev tools that cell is 50px
4. Switch to table mode
5. See in dev tools that the cell is no longer 50px in size

In Grid mode:

Image

In Table mode:

Image

## 🌍 Your Environment

* Browser latest Firefox and Edge
* .NET 10 and latest dev-v5 build

vnbaaij:
Fixes issues mentioned in #4582 

1. Header not applying width twice
Yup, the style=100% is the culprit here. I'll remove it and verify.

2. Additional space when sortable is enabled
The optimal min-width is arbitrary and depends mostly on what title you are using. With the word year and applying a sort (so the arrow gets shown (with its own margin/padding)) the minimal with could be brought down to around 80 pixels. I'll set the min-width to 75px in the CSS.
There was also a calc that subtracted 20px from the width. I'll remove that also

3. Wrong sizing
When switching to table layout, the table uses the default table-layout: auto, the column width is being determined by the content of the entire column (not just the ). The browser will expand the column to fit it's content. I'll change it so that it uses table-layout: fixed in that mode. It already sets it to that when resizing is enabled anyway

### DIFF
--- a/examples/Demo/FluentUI.Demo.Client/Documentation/Components/DataGrid/FluentDataGrid.md
+++ b/examples/Demo/FluentUI.Demo.Client/Documentation/Components/DataGrid/FluentDataGrid.md
@@ -45,6 +45,8 @@ again will toggle the sort direction. When `HeaderCellAsButtonWithMenu` is true,
 A sort can be removed by right clicking (or by pressing <kbd>Shift</kbd> + <kbd>r</kbd>) on the header column (with exception of
 the default sort).
 
+_The minimal width for a sortable column is 75 pixels._
+
 ## Row size
 
 The DataGrid offers a `RowSize` parameter which allows you to use different preset row heights. The value uses the `DataGridRowSize`
--- a/src/Core/Components/DataGrid/Columns/ColumnBase.razor
+++ b/src/Core/Components/DataGrid/Columns/ColumnBase.razor
@@ -118,8 +118,6 @@
 
                 @if (Sortable.HasValue ? Sortable.Value : IsSortableByDefault())
                 {
-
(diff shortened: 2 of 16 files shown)


dvoituron (approved):

MarvinKlein1508:
I'm going to review and test on monday. Currently I'm not at home

MarvinKlein1508:
@vnbaaij I did some testing and the only strange thing I could find is that the width is now being applied to the header as well which results in the button overflows in the next column. See:
grafik

MarvinKlein1508:
@vnbaaij I think we should mention the minimum of `75px` for `Width` when `Sortable` is enabled in the docs.

MarvinKlein1508 (approved):

PR #4620, opened 2026-03-12
[dev-v5][Calendar] Add MinDate and MaxDate

### LINKED ISSUE #4611: feat: add the capability to "deactivate" the header in the FluentCalendar
# 🙋 Feature Request

Adds a boolean to deactivate the header in the FluentCalendar...

## 🤔 Expected Behavior

As we can see here, the header of the calendar is "active" : a click on it show a new view and we have up and down buttons...

Image

For the FluentCxScheduler, where you can have a sample here :

https://github.com/Community-Toolkit-for-Fluent-UI-Blazor/fluentui-blazor-community-extensions/tree/feat/migrate-fluentcx-scheduler

I must recreate the same calendar with a deactivated header like that :

Image

The same calendar without the up / down buttons and without a clickeable header...

## 💁 Possible Solution

A simple boolean like this : 

```[Parameter]
public bool IsHeaderActive { get; set; }

And on the razor side : 

@if (IsHeaderActive) 
{
    Same code as original
}
else
{
     <!-- I don't know the component used but let say that it is -->
     <FluentText>@Header</FluentText>
}```


## 🔦 Context

<!--- What are you trying to accomplish? -->
<!--- How has not having this feature affected you? -->
<!--- What alternatives have you considered? -->

To reduce the complexity and duplication of some code for the Scheduler, it could be useful to be able to use the FluentCalendar instead of rewrote it with a deactivated header...

## 💻 Examples

<!-- Examples help us understand the requested feature better -->
<!-- Attach screenshots or images if they would add detail to your request -->

### PR DESCRIPTION
**Pull request by vnbaaij**
- Add MinDate and MaxDate to `FluentCalendar`

`MinDate` and `MaxDate` allow you to specify a range in which the calendar can render selectable dates (independently from other influencers like `DisabledDateFunc`)

Example
- Left calendar: Only a date in the current month can be selected. Title is not clickable but only current month is enabled, same goes for year)
- Middle calendar: `Mindate="@(new DateTime(2026, 3, 1))"` and `MaxDate="@(new DateTime(Today.Year+1, 7, 30))"` so january and february of this year can not be selected anymore. Clicking on the year header shows the range with only 2026 and 2027 enabled (second image)
- Right calendar. Only `MaxDate="@(new DateTime(2030, 1, 1))"` set. When reaching the end of the range, Arrow down (next) is not rendered as those years are out of range.

image

image

Unit tests have been added

### CODE REVIEW

File: src/Core/Components/DateTime/Calendar/CalendarTitles.cs
Comment (dvoituron) [inline src/Core/Components/DateTime/Calendar/CalendarTitles.cs:1]: This file seems empty

File: src/Core/Components/DateTime/FluentCalendar.razor.cs [hunk 1]
```diff
@@ -244,45 +244,100 @@ internal CalendarTitles<TValue> GetTitles()
...
+        var candidatePickerMonth = View switch
+        {
+            CalendarViews.Days => PickerMonth.AddMonths(+1, Culture),
+            CalendarViews.Months => PickerMonth.AddYears(+1, Culture),
+            CalendarViews.Years => PickerMonth.AddYears(+12, Culture),
+            _ => PickerMonth,
+        };
 
-        switch (View)
+        var maxDateTime = MaxDate.ConvertToDateTime();
```
dvoituron: Can you move this part of the code to an extension method such as `AdaptedForMaxDate`
to get something like this in the switch?
```csharp
 PickerMonth = PickerMonth.AddMonths(+1, Culture).AdaptedForMaxDate(MaxDate, Culture);` 
```

src/Core/Components/DateTime/FluentCalendar.razor.cs
```diff
@@ -244,45 +244,100 @@ internal CalendarTitles<TValue> GetTitles()
...
+            CalendarViews.Days => PickerMonth.AddMonths(-1, Culture),
+            CalendarViews.Months => PickerMonth.AddYears(-1, Culture),
+            CalendarViews.Years => PickerMonth.AddYears(-12, Culture),
+            _ => PickerMonth,
+        };
 
-            case CalendarViews.Months:
-                PickerMonth = PickerMonth.AddYears(-1, Culture);
-                break;
+        var minDateTime = MinDate.ConvertToDateTime();
```
dvoituron: Can you move this part of the code to an extension method such as `AdaptedForMinDate`
to get something like this in the switch?
```csharp
 PickerMonth = PickerMonth.AddMonths(-1, Culture).AdaptedForMinDate(MinDate, Culture);` 
```

src/Core/Components/DateTime/FluentCalendarBase.cs
```diff
@@ -76,6 +77,71 @@ protected FluentCalendarBase(LibraryConfiguration configuration) : base(configur
     [Parameter]
     public virtual CalendarViews View { get; set; } = CalendarViews.Days;
 
+    /// <summary>
+    /// Gets or sets whether the header of the calendar (title and prev/next arrows) is displayed interactively.
+    /// True (default) renders clickable title and prev/next arrows. False renders just the title as static text.
+    /// </summary>
+    [Parameter]
+    public virtual bool HeaderInteractive { get; set; } = true;
```
dvoituron: I find the name “HeaderReadOnly” more appropriate. No?

src/Core/Components/DateTime/FluentCalendarBase.cs
```diff
@@ -76,6 +77,71 @@ protected FluentCalendarBase(LibraryConfiguration configuration) : base(configur
...
+
+        if (max.HasValue && date > max.Value)
+        {
+            return true;
+        }
+
+        return false;
+    }
+
+    internal bool IsPeriodOutsideRange(DateTime periodStart, DateTime periodEnd)
```
dvoituron: Can you add these methods `IsSelectionOutsideRange`, `IsPeriodOutsideRange` and `IsOutsideRange` to `RangeOfDates`?

dvoituron (changes requested):
I need to run some more tests, but CalendarMonth is no longer working properly with the keyboard: the years scroll by, but I can no longer select the month.

In my opinion, it would be better to create two PRs: one for the header and one for the min/max (which is much more complex).

src/Core/Components/DateTime/FluentCalendarBase.cs
vnbaaij: No, interactive is more clear in this case

src/Core/Components/DateTime/Calendar/CalendarTitles.cs
vnbaaij: Ok, removed with next commit

src/Core/Components/DateTime/FluentCalendarBase.cs
vnbaaij: Moved these in next commit

vnbaaij:
> In my opinion, it would be better to create two PRs: one for the header and one for the min/max (which is much more complex).

Getting the `HeaderInteractive` change out of this PR doesn't really make sense anymore now. It is only a tiny thing in this whole PR

src/Core/Components/DateTime/FluentCalendar.razor.cs
vnbaaij: Done

src/Core/Components/DateTime/FluentCalendar.razor.cs
vnbaaij: Done

vnbaaij:
Using CalenderMonth with keyboard works fine on my side:

CalenderMonth-keyboard select

dvoituron:
> Using CalenderMonth with keyboard works fine on my side:
> 
> 
> 
> CalenderMonth-keyboard select
> 
> 

You need to press KeyDown multiple times to go to the next year. The first month don't keep the focus (previously that's worked)

dvoituron (changes requested):
The `HeaderInteractive` parameter doesn't work with FluentDatePicker
```razor
<FluentDatePicker Label="Days view"
                    HeaderInteractive="false"
                    @bind-Value="@SelectedValue" />
```

examples/Demo/FluentUI.Demo.Client/Documentation/Components/DateTime/Calendar/FluentCalendar.md
```diff
@@ -16,6 +16,17 @@ The user can switch between these views using the title item: the month name or
 
 {{ CalendarDefault }}
 
+### Setting a minimum and maximum date
```
dvoituron-msft: Can you use the H2 and not H3?

src/Core/Components/DateTime/FluentCalendarBase.cs
```diff
@@ -76,6 +76,20 @@ protected FluentCalendarBase(LibraryConfiguration configuration) : base(configur
     [Parameter]
     public virtual CalendarViews View { get; set; } = CalendarViews.Days;
 
+    /// <summary>
+    /// Gets or sets the minimum date that can be selected in the calendar. If not set, there is no minimum date.
+    /// </summary>
+    [Parameter]
+    public TValue? MinDate { get; set; }
```
dvoituron-msft, line 83: This is not implemented for the DatePicker :-(

src/Core/Components/DateTime/RangeOfDates.cs
```diff
@@ -38,4 +39,50 @@ public override string ToString()
     {
         return $"From {Start?.ToString("yyyy-MM-dd", CultureInfo.InvariantCulture) ?? "<null>"} to {End?.ToString("yyyy-MM-dd", CultureInfo.InvariantCulture) ?? "<null>"}.";
     }
+
+    internal bool IsOutsideRange(DateTime value)
```
dvoituron-msft, line 43: This method must be tested in RangeOfDatesTests.cs

src/Core/Components/DateTime/RangeOfDates.cs
```diff
@@ -38,4 +39,50 @@ public override string ToString()
...
+
+        if (max.HasValue && date > max.Value)
+        {
+            return true;
+        }
+
+        return false;
+    }
+
+    internal bool IsPeriodOutsideRange(DateTime periodStart, DateTime periodEnd)
```
dvoituron-msft, line 62: This method must be tested in RangeOfDatesTests.cs

src/Core/Components/DateTime/RangeOfDates.cs
```diff
@@ -38,4 +39,50 @@ public override string ToString()
...
+
+        if (max.HasValue && periodStart.Date > max.Value)
+        {
+            return true;
+        }
+
+        return false;
+    }
+
+    internal bool IsSelectionOutsideRange(DateTime value, CalendarViews view, CultureInfo culture)
```
dvoituron-msft, line 80: This method must be tested in RangeOfDatesTests.cs

src/Core/Components/DateTime/FluentCalendarMonth.cs
```diff
@@ -59,7 +59,9 @@ internal FluentCalendarMonth(FluentCalendar<TValue> calendar, DateTime month)
     /// <summary>
     /// Gets the title of the month in the format [month] [year].
     /// </summary>
+#pragma warning disable MA0076 // Do not use implicit culture-sensitive ToString in interpolated strings
     public string Title => $"{_calendar.CalendarExtended.GetMonthName(Month)} {Month.GetYear(_calendar.Culture):0000}";
```
dvoituron-msft: You should be able to use the Culture argument:
```
public string Title => $"{_calendar.CalendarExtended.GetMonthName(Month)} {Month.GetYear(_calendar.Culture):0000, _calendar.Culture}";
```

src/Core/Components/DateTime/CalendarTValue.cs
```diff
@@ -152,6 +152,122 @@ public static TValue AddYears<TValue>(this TValue? value, int years, CultureInfo
         return DateTimeExtensions.AddYears(value.ConvertToRequiredDateTime(), years, culture).ConvertToTValue<TValue>();
     }
 
+    internal static TValue AdaptedForMinDate<TValue>(this TValue value, TValue? minDate, CultureInfo culture)
```
dvoituron-msft: All of these new methods must be tested in CalendarTValueTests.cs

dvoituron-msft (changes requested):

src/Core/Components/DateTime/FluentCalendarMonth.cs
vnbaaij: Done

src/Core/Components/DateTime/CalendarTValue.cs
vnbaaij: These methods have been removed and replaced with logic that resides in `FluentCalendar.razor.cs`. Tests for this new code have been added

src/Core/Components/DateTime/FluentCalendarBase.cs
vnbaaij: Forwarding the MinDate and MaxDate to the calendar popup now

PR #4632, opened 2026-03-16
[DragContrainer][DropZone] Switch to `EventCallback`

### LINKED ISSUE #4598: Port over #4590 to v4
Need this in our current v4 app. I'm going to port this change over to v4 as well.

MarvinKlein1508:
# Pull Request

## 📖 Description

This PR updates the `FluentDragContainer` and `FluentDropZone` components and exchanges all `Action<FluentDragEventArgs<TItem>>` with `EventCallback<FluentDragEventArgs<TItem>>`. This makes it possible to provide both `Task` and `void` methods as callback.

This three methods are now possible:
```cs
private void OnRowDropEnd(FluentDragEventArgs<FormRow> e) ...

private Task OnRowDropEnd(FluentDragEventArgs<FormRow> e) ...

private async Task OnRowDropEnd(FluentDragEventArgs<FormRow> e) ...
```

This should not introduce any breaking change.

These changes are the same as we have already done in v5. See: #4590

### 🎫 Issues

Fix: #4598 

### General

### Component-specific

### DIFF
--- a/examples/Demo/Shared/Microsoft.FluentUI.AspNetCore.Components.xml
+++ b/examples/Demo/Shared/Microsoft.FluentUI.AspNetCore.Components.xml
@@ -3936,6 +3936,7 @@
             left (full height), right (full height)
             or screen middle (using Width and Height properties).
             HorizontalAlignment.Stretch is not supported for this property.
+            NOTE: Left/Right only works for Panels, not Dialogs, which are always centered.
             </summary>
         </member>
         <member name="P:Microsoft.FluentUI.AspNetCore.Components.DialogParameters.Title">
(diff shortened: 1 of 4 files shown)


dvoituron:
It is a breaking change, but only for specific usage patterns.

## Breaking scenarios

### 1. Null checks in consumer code

```csharp
// Before (works with Action?)
if (OnDragEnter != null) { ... }

// After (EventCallback is a struct, never null)
if (OnDragEnter.HasDelegate) { ... }
```

### 2. Direct C# assignment (e.g., in tests or programmatic usage)

```csharp
// Before
component.OnDragEnter = (args) => { ... };  // ✅

// After — doesn't compile, EventCallback can't be assigned this way
component.OnDragEnter = (args) => { ... };  // ❌
```

### 3. Invocation inside the component

```csharp
// Before
OnDragEnter?.Invoke(args);

// After
await OnDragEnter.InvokeAsync(args);
``` 

### NOT breaking (most common case)
Typical Razor attribute binding works the same for both:

```xml
<FluentSomething OnDragEnter="HandleDrag" />
<FluentSomething OnDragEnter="@((args) => DoSomething(args))" />
```

### Conclusion

I don't think it's necessary to apply this change in v4

MarvinKlein1508:
> It is a breaking change, but only for specific usage patterns.
> ## Breaking scenarios
> ### 1. Null checks in consumer code
> 
> ```cs
> // Before (works with Action?)
> if (OnDragEnter != null) { ... }
> 
> // After (EventCallback is a struct, never null)
> if (OnDragEnter.HasDelegate) { ... }
> ```

I don't think that is something we should be bothered about.  I can't think of any good reason why someone would check externally for the internal events.
 
> ### 2. Direct C# assignment (e.g., in tests or programmatic usage)
> 
> ```cs
> // Before
> component.OnDragEnter = (args) => { ... };  // ✅
> 
> // After — doesn't compile, EventCallback can't be assigned this way
> component.OnDragEnter = (args) => { ... };  // ❌
> ```

You are right about this one. Since no changes were required in v5 for this I assumed that it doesn't break anything. But these are valid definitions (although I can't think of any reason why someone would define it like this outside from tests). This would require the developer to update the syntax to: `EventCallback.Factory.Create(this, Method);`
 
> ### 3. Invocation inside the component
> 
> ```cs
> // Before
> OnDragEnter?.Invoke(args);
> 
> // After
> await OnDragEnter.InvokeAsync(args);
> ```

Yes the inner part of the component has been updated but it behaves exactly the same.

> ### NOT breaking (most common case)
> 
> Typical Razor attribute binding works the same for both:
> 
> ```
> <FluentSomething OnDragEnter="HandleDrag" />
> <FluentSomething OnDragEnter="@((args) => DoSomething(args))" />
> ```

That's the way it's done like 99% of the time.

> ### Conclusion
> 
> I don't think it's necessary to apply this change in v4

The main reason I did this change is to support async code. Right now you can only forward `void` or `async void` and you can't catch any exceptions within `async void`. So I do believe that this change might be beneficial for v4 as well. Even though it might break some edge cases. 

Anyways I will create a second PR for v5 to mention this changes in the migration docs.

dvoituron:
I don't think this is the right time to introduce breaking changes in v4.
In v4, developers can still call an asynchronous method via `ComponentBase.InvokeAsync`

For v5, it's clearly an advantage.

MarvinKlein1508:
@dvoituron I can live with that! :)

PR #4641, opened 2026-03-22
[Docs] Fix incorrect Icon in search after Navigation

sebbogle:
# Pull Request

## 📖 Description

Fix an issue where the 'X' icon would remain visible on the [Docs](https://www.fluentui-blazor.net/) Search input after a user has selected an item and navigated away. This is due to the check that the Autocomplete component does where with this component usage, this.SelectedOption is not null at this point.

Updating the component to use `Multiple="false"` fixes this issue and still allows for correct handling of navigation.

I had originally ran into this issue myself as when implementing a similar Navigation Search for my own project where I used the docs as a starting point for how best to achieve this with FluentUI.

## 👩‍💻 Reviewer Notes

Current Behaviour:
fluentUI-Search-Icon-Before

Behaviour After Changes:
fluentUI-Search-Icon-After

## 📑 Test Plan

Manually testing after making the change to see that the icon correctly updates to the search icon when navigating matching the expended behaviour as the search input text clears. Checked with both mouse clicks, and keyboard arrow keys + enter.

As this change is just to the docs page rather than a change to the component library itself, I don't believe there is any unit testings that covers this. However if I'm wrong, please just let me know and I'd be happy to add something in.

### General

### DIFF
--- a/examples/Demo/Shared/Shared/DemoSearch.razor
+++ b/examples/Demo/Shared/Shared/DemoSearch.razor
@@ -1,13 +1,13 @@
 ﻿<FluentAutocomplete @ref="_searchAutocomplete"
                     TOption="NavItem"
+                    Multiple="false"
                     Width="250px"
                     AutoComplete="off"
                     Placeholder="Type / to search..."
-                    MaximumSelectedOptions="1"
                     OptionText="@(item => item.Title)"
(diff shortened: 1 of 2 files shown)


sebbogle:

vnbaaij (approved):

PR #4653, opened 2026-03-30
[dev-v5][DataGrid] Add `DisableCellFocus` parameter

### LINKED ISSUE #4468: feat: add ability to turn off (or remove entirely) focusing of data grid cells in DataGrid
# 🙋 Feature Request

I would like the ability to be able to turn off cells in a data grid being focusable. We have some data grids in our app that contain focusable elements such as buttons and text fields, and because of the data grid cells also being focusable, it is taking double the amount of tab key presses to move between the focusable elements in the grid.

To be honest I don't see data grid cells being focusable as particularly useful, as when for example the data grid cell contains text you can't seem to do anything useful with it such as press CTRL-C to copy the text. But maybe it's an a11y thing. However, being able to focus the column headers is useful, as then you can change for example the sort order using the keyboard.

## 🔦 Context

We have some data grids in our app that have columns that contain text but also has a column that is a text field. Product management is complaining that it takes two tab key presses to move between the text fields (caused by the data grid cells also being focusable). If we could turn off data grid cells being focusable then we could move between text fields in the grid with one tab key press.

MarvinKlein1508:
# Pull Request

## 📖 Description

This PR introduces a new parameter `DisableCellFocus` to `ColumnBase<TGridItem>` which makes it possible to disable cell focus for each column individually.

To achieve this I have removed the hardcoded `tabindex="0"` from `FluentDataGridCell.razor` and instead I'm sending `tabindex` via `AdditionalAttributes` from `FluentDataGrid`.

### 🎫 Issues

Fix #4468

### General

### Component-specific


src/Core/Components/DataGrid/FluentDataGrid.razor
```diff
@@ -117,7 +117,7 @@
 
 				string? tooltip = col.Tooltip ? @col.RawCellContent(item) : null;
 
-				<FluentDataGridCell GridColumn=@(colIndex + 1) Class="@ColumnJustifyClass(col)" Style="@col.Style" @key="@col" TGridItem="TGridItem" Item="@item" title="@tooltip" aria-label="@tooltip">
+				<FluentDataGridCell GridColumn=@(colIndex + 1) Class="@ColumnJustifyClass(col)" Style="@col.Style" @key="@col" TGridItem="TGridItem" Item="@item" title="@tooltip" aria-label="@tooltip" tabindex="@(col.DisableCellFocus ? -1 : 0)">
```
vnbaaij: Would it not be better to use `null` in case the `DisableCellFocus` is set?

src/Core/Components/DataGrid/FluentDataGrid.razor
```diff
@@ -246,7 +246,7 @@
 		}
 
 		<FluentDataGridRow Class="@EMPTY_CONTENT_ROW_CLASS" TGridItem="TGridItem">
-			<FluentDataGridCell Class="empty-content-cell" Style="@style" colspan="@colspan">
+			<FluentDataGridCell Class="empty-content-cell" Style="@style" colspan="@colspan" tabindex="0">
```
vnbaaij: I don't think the empty content cell needs to be focusable

src/Core/Components/DataGrid/FluentDataGridCell.razor
```diff
@@ -9,7 +9,6 @@
         class="@ClassValue"
         style="@StyleValue"
         role="gridcell"
-        tabindex="0"
```
vnbaaij, line 12: It is still possible to define a DataGrid declaratively (see manual example) so tabindex should not just be removed. Maybe pass it through with the InternalGridContext instead of via the change in `FluentDataGrid.razor`?

src/Core/Components/DataGrid/Columns/ColumnBase.razor.cs
```diff
@@ -203,6 +203,12 @@ public abstract partial class ColumnBase<TGridItem>
     [Parameter]
     public bool HierarchicalToggle { get; set; }
 
+    /// <summary>
+    /// Disables the ability for cells to receive focus.
+    /// </summary>
+    [Parameter]
+    public bool DisableCellFocus { get; set; }
```
vnbaaij, line 210: What happens with a `SelectColumn` if this is set to true?

vnbaaij (changes requested):

src/Core/Components/DataGrid/FluentDataGridCell.razor
MarvinKlein1508: This shouldn't be a problem because you can just do the following:
`<FluentDataGridCell tabindex="0">` 

`AdditionalAttributes` will be passed down. It's the same approach which the `FluentDataGrid` uses.

src/Core/Components/DataGrid/Columns/ColumnBase.razor.cs
MarvinKlein1508: Tested this as well. The cell won't receive focus. Instead the input within the cell will receive it (what is excatly the user wants)

src/Core/Components/DataGrid/FluentDataGrid.razor
MarvinKlein1508: I agree. I put it here because it was part of the unit test. I can remove it and update the test.

vnbaaij: Yes, please update the test then

src/Core/Components/DataGrid/FluentDataGrid.razor
MarvinKlein1508: done.

vnbaaij (approved):

PR #4662, opened 2026-03-31
[dev-v5] Add FluentAutocomplete

### LINKED ISSUE #4187: [dev-v5] Migrate AutoComplete component

dvoituron:
# [dev-v5] Add FluentAutocomplete

peek_1

## Unit Tests

image

## TODO

- Accessibility
- `Virtualize = true`

Close #4187


examples/Demo/FluentUI.Demo.Client/Documentation/Components/List/Autocomplete/Examples/AutocompleteCustomized.razor
```diff
@@ -0,0 +1,54 @@
...
+<div>Selected: <b>@string.Join("; ", SelectedCountries.Select(c => c.Name))</b></div>
+<div>Search Text: <b>@SearchText</b></div>
+
+<FluentAutocomplete TOption="Country"
+                    TValue="string"
+                    Width="100%"
+                    Label="Select countries"
+                    Placeholder="Type to search..."
+                    ShowProgressIndicator="@ShowProgressIndicator"                    
+                    MaxAutoHeight="@((MaxAutoHeight) ? "150px" : null)"
```
MarvinKlein1508: you probably wanna use `auto` here instead of a fixed height. If you define a too small value then the component will render kind of strang. `auto` is probably the case must users want by default as it allows for natural growing

examples/Demo/FluentUI.Demo.Client/Documentation/Components/List/Autocomplete/FluentAutocomplete.md
```diff
@@ -0,0 +1,46 @@
...
+It combines a free-text input with a filtered list of options, allowing users to either select from the suggestions or type their own value.
+
+It is particularly useful when the list of options is large, as the user can narrow down choices without scrolling through all available items.
+
+## Keyboard interaction
+
+| Key | Behavior |
+|---|---|
+| **Type text** | Filters the list of options and triggers the `OnSearchAsync` method to fetch matching results. |
+| **Arrow Down / Arrow Up** | Navigates through the items in the suggestion list. |
```
MarvinKlein1508: Also opens the suggestion list.

src/Core/Components/List/FluentAutocomplete.razor
```diff
@@ -0,0 +1,136 @@
...
+                         MessageTemplate="@MessageTemplate"
+                         MessageCondition="@MessageCondition"
+                         MessageIcon="@MessageIcon"
+                         MessageState="@MessageState"
+                         Placeholder="@Placeholder"
+                         Style="@StyleValue"
+                         Name="@Name"
+                         auto-complete="true"
+                         Width="@Width"
+                         AutoComplete="false"
```
MarvinKlein1508, line 44: in v4 this is a parameter.

src/Core/Components/List/FluentAutocomplete.razor
```diff
@@ -0,0 +1,136 @@
...
+                         Required="@Required"
+                         Message="@Message"
+                         MessageTemplate="@MessageTemplate"
+                         MessageCondition="@MessageCondition"
+                         MessageIcon="@MessageIcon"
+                         MessageState="@MessageState"
+                         Placeholder="@Placeholder"
+                         Style="@StyleValue"
+                         Name="@Name"
+                         auto-complete="true"
```
MarvinKlein1508, line 41: do we need this line if we pass in the other autocomplete parameter

MarvinKlein1508: nevermind. This one is to apply CSS rules. But kinda confusing though.

src/Core/Components/List/FluentAutocomplete.razor.cs
```diff
@@ -0,0 +1,326 @@
...
+
+    /// <summary>
+    /// Gets or sets the short hint displayed in the input before the user enters a value.
+    /// </summary>
+    [Parameter]
+    public string? Placeholder { get; set; }
+
+    /// <summary>
+    /// Gets or sets the delay, in milliseconds, before to raise the event.
+    /// Default is 400 milliseconds.
```
MarvinKlein1508, line 62: 200

MarvinKlein1508:
I tried this code:
```razor
<div>Selected: <b>@string.Join("; ", Selected.Select(x => x.Name))</b></div>

<FluentAutocomplete TOption="MyUser"
                    TValue="int"
                    Width="100%"
                    Label="Name"
                    Placeholder="Name"
                    OptionText="(option) => option.Name"
                    OptionValue="(option) => option.UserId"
                    OptionSelectedComparer="CompareUser"
                    OnOptionsSearch="@OnSearchAsync"
                    @bind-SelectedItems="@Selected" />
@code
{
    IEnumerable<MyUser> Selected { get; set; } = [];

    Task OnSearchAsync(OptionsSearchEventArgs<MyUser> e)
    {
        e.Items =
        [
            new MyUser
            {
                UserId = 1,
                Name = "Marvin Klein"
            },
            new MyUser
            {
                UserId = 2,
                Name = "Denis Voituron"
            },
    ];

        return Task.CompletedTask;
    }

    public bool CompareUser(int checkItemId, int selectedItemId)
    {
        return checkItemId == selectedItemId;
    }

    public class MyUser
    {
        public int UserId { get; set; }
        public string Name { get; set; } = string.Empty;
    }
}
```

The behaviour of the component gets confused by this.

grafik

We should also add documentation to `OptionSelectedComparer` to show the user which parameter is the current checking item and which one is the value from the object the user selects. Also there is a breaking change from v4 as I can no longer pass in an `IEqualityComparer<T>`

examples/Demo/FluentUI.Demo.Client/Documentation/Components/List/Autocomplete/Examples/AutocompleteCustomized.razor
dvoituron: Correct

src/Core/Components/List/FluentAutocomplete.razor
dvoituron: I know but this is an internal attribute used for my CSS. And I cannot use `autocomplete` already defined by default in HTL elements :-( Or do you have another suggestion for a name?

src/Core/Components/List/FluentAutocomplete.razor
dvoituron: Yes, but obsolete now because you cannot activate the HTML `autocomplete` and the `FluentAutocomplete` in the same time :-)

dvoituron:
@MarvinKlein1508 The `OptionSelectedComparer` is not yet used.
But do you think better to continue to use a `IEqualityComparer` and not a `Func`?
I prefer to continue to use this IEqualityComparer, so I need to create another PR to update all list components

MarvinKlein1508:
@dvoituron what is missing here? Or is it a leftover from the v4 docs?
image

dvoituron:
> @dvoituron what is missing here? Or is it a leftover from the v4 docs? 
Keyboard navigation is complete, but NVDA (text-to-speech) features are not yet available.

examples/Demo/FluentUI.Demo.Client/Documentation/GetStarted/Migration/MigrationFluentAutocomplete.md
```diff
@@ -0,0 +1,94 @@
+---
+title: Migration FluentAutocomplete
+route: /Migration/Autocomplete
+hidden: true
+---
+
```
MarvinKlein1508, line 6: We should also describe that the component is now growing automatically when horizontal space runs out due to the missing `FluentFlipper` component. This could break layouts if you switch from v4 to v5. A simple solutions would be to use the `MaxAutoHeight` parameter set to `unset` or `auto`.

examples/Demo/FluentUI.Demo.Client/Documentation/Components/List/Autocomplete/Examples/AutocompleteCustomized.razor
```diff
@@ -0,0 +1,83 @@
+﻿@using System.Transactions
```
MarvinKlein1508: using is not necessary

examples/Demo/FluentUI.Demo.Client/Documentation/Components/List/Autocomplete/DebugPages/AutocompleteComparer.razor
```diff
@@ -0,0 +1,66 @@
+@page "/Lists/Autocomplete/Debug/Comparer"
+
+<div>Selected: <b>@string.Join("; ", Selected.Select(x => x.Name))</b></div>
+
+<FluentAutocomplete TOption="MyUser"
```
MarvinKlein1508: Can we include this example in the main docs? I regulary need this sample and it makes it easier to copy & paste it for me.

In v4 we had this section:

Image

I wouldn't mind if we hide it behind a link for _Advanced examples_

examples/Demo/FluentUI.Demo.Client/Documentation/Components/List/Autocomplete/DebugPages/AutocompleteDebug.razor
```diff
@@ -0,0 +1,110 @@
+﻿@page "/Lists/Autocomplete/Debug"
+
+@using static FluentUI.Demo.SampleData.Olympics2024
```
MarvinKlein1508, line 3: Can we hide this using behind a `_Import.razor` ? I remember some people were getting confused by it when they copied over the sample from the docs.

src/Core/Components/List/FluentAutocomplete.razor
```diff
@@ -0,0 +1,199 @@
...
+                                }
+                            }
+                        }
+                    </div>
+                }
+                else
+                {
+                    @if (_internalSelectedItem is not null)
+                    {
+                        <div tabindex="0" role="button" aria-haspopup="listbox" aria-expanded="@_isOpen" style="cursor: pointer; width: 100%;">
```
MarvinKlein1508, line 88: Just a question: wouldn't it make sense to use a button here for accessability reasons? Is there any reason this has to be a div? 

This way we wouldn't have to specifiy `tabindex="0" role="button"` & `style="cursor: pointer;`

examples/Demo/FluentUI.Demo.Client/Documentation/GetStarted/Migration/MigrationFluentAutocomplete.md
dvoituron: Added
>In v4, the component used horizontal scroll navigation (`FluentFlipper`) when selected items exceeded the available width.
In v5, this horizontal navigation has been removed, and the component **grows vertically** to display all selected items:
set the `MaxAutoHeight` parameter to `unset` or a specific value. 
You can also use `MaxSelectedWidth` to truncate long selected item labels, reducing the horizontal space each badge occupies.
This may break existing layouts if you relied on the fixed-height behavior.

examples/Demo/FluentUI.Demo.Client/Documentation/Components/List/Autocomplete/DebugPages/AutocompleteComparer.razor
dvoituron: Added

examples/Demo/FluentUI.Demo.Client/Documentation/Components/List/Autocomplete/DebugPages/AutocompleteDebug.razor
dvoituron: In fact, if I remove the `using` statement, the code will be limited to just `Country`, which will “come out of nowhere.” Having this `using` statement actually allows us to show exactly where it comes from.

src/Core/Components/List/FluentAutocomplete.razor
dvoituron: It won't make much of a difference: when using a `button`, I have to add a ton of styles to remove the button's visual appearance. Here, I just need to specify the role so that accessibility tools recognize it as a button

examples/Demo/FluentUI.Demo.Client/Documentation/Components/List/Autocomplete/FluentAutocomplete.md
```diff
@@ -0,0 +1,67 @@
...
+title: Autocomplete
+route: /Lists/Autocomplete
+---
+
+# Autocomplete
+
+An **Autocomplete** component is a text input that provides real-time suggestions as the user types.
+It combines a free-text input with a filtered list of options, allowing users to either select from the suggestions or type their own value.
+
+It is particularly useful when the list of options is large, as the user can narrow down choices without scrolling through all available items.
```
vnbaaij: This is particularly useful when the list of options is large, as the user can narrow down the list op options without needing to scroll through all available items.

src/Core/Components/Icons/FluentIcon.razor.cs
vnbaaij, line 1: Should these (valuable) new capabilities not be mentioned on the `FluentIcon` doc page?

vnbaaij (changes requested):

src/Core/Components/Icons/FluentIcon.razor.cs
dvoituron: There's no need for a dedicated section for this, but I see that the API isn't included in the Icons documentation... I'll create another PR for that.

vnbaaij (approved):

MarvinKlein1508 (approved):

PR #4677, opened 2026-04-07
[MessageBar] Fix NullReferenceException and thread safety in MessageService

### LINKED ISSUE #4674: fix: NullReferenceException in MessageService.ShowMessageBarAsync
# 🐛 Bug Report

I see about 30 NullReferenceException errors per week from MessageService.ShowMessageBarAsync in my app.

I have a component that references IMessageService and can call ShowMessageBarAsync. But this sometimes errors. I think the problem is ShowMessageBarAsync assumes that `OnMessageItemsUpdatedAsync` has been subscribed to:

https://github.com/microsoft/fluentui-blazor/blob/0625bb0d7a6e0909bebab4ca9cbf98d586b8817d/src/Core/Components/MessageBar/Services/MessageService.cs#L266

This happens in `FluentMessageBarProvider.OnInitialize`:

https://github.com/microsoft/fluentui-blazor/blob/0625bb0d7a6e0909bebab4ca9cbf98d586b8817d/src/Core/Components/MessageBar/FluentMessageBarProvider.razor.cs#L93

## 🤔 Expected Behavior

No error.

## 😯 Current Behavior

There is a race between OnInitialize in the two controls causing NullReferenceExceptions.

## 💁 Possible Solution

MessageService doesn't call the event if it isn't registered?

Also, isn't [AllMessages](https://github.com/microsoft/fluentui-blazor/blob/0625bb0d7a6e0909bebab4ca9cbf98d586b8817d/src/Core/Components/MessageBar/Services/MessageService.cs#L42-L56) not thread safe? The lock around it just locks while accessing the property. A copy isn't being returned so a message could be added to it while the message bar provider is enumerating the collection, causing an error. Either a copy needs to be made here, or a copy is made each time the message list is modifed.

## 🌍 Your Environment

JamesNK:
## Summary

Fix #4674

### NullReferenceException in `ShowMessageBarAsync`

`ShowMessageBarAsync` called `await OnMessageItemsUpdatedAsync!.Invoke()` without a null check. If `ShowMessageBarAsync` is called before `FluentMessageBarProvider.OnInitialized` subscribes to the event, this throws a `NullReferenceException`. The delegate is now captured in a local variable before invoking for thread safety:

```csharp
if (OnMessageItemsUpdatedAsync is { } handler)
{
    await handler.Invoke();
}
```

### Thread safety of `AllMessages` and `MessagesToShow`

Both properties/methods acquired a read lock but returned the live `MessageList` reference. The lock was released before the caller could enumerate, so concurrent writes (e.g. adding a message) could modify the collection during enumeration, causing `InvalidOperationException`.

Both now return a snapshot copy via `.ToList()` while still holding the read lock.

### DIFF
--- a/src/Core/Components/DataGrid/FluentDataGrid.razor.cs
+++ b/src/Core/Components/DataGrid/FluentDataGrid.razor.cs
@@ -372,7 +372,7 @@ public partial class FluentDataGrid<TGridItem> : FluentComponentBase, IHandleEve
 
     // Returns Loading if set (controlled). If not controlled,
     // we assume the grid is loading until the next data load completes
-    internal bool EffectiveLoadingValue => Loading ?? ItemsProvider is not null;
+    internal bool EffectiveLoadingValue => Loading ?? (ItemsProvider is not null);
 
     private ElementReference? _gridReference;
     //private DotNetObjectReference<Type>? _dotNetObjectReference;
--- a/src/Core/Components/MessageBar/Services/MessageService.cs
+++ b/src/Core/Components/MessageBar/Services/MessageService.cs
@@ -66,9 +66,12 @@ public class MessageService : IMessageService, IDisposable
         MessageLock.EnterReadLock();
         try
         {
-            IEnumerable<Message> messages = string.IsNullOrEmpty(section)
-                       ? MessageList
-                       : MessageList.Where(x => x.Section == section);
+            IEnumerable<Message> messages = MessageList;
+
+            if (!string.IsNullOrEmpty(section))
+            {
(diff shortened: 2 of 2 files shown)


vnbaaij (dismissed):

vnbaaij:
Hi @JamesNK please re-target this PR to go to the dev branch

dvoituron (approved):

dvoituron (changes requested):
As Vincent pointed out, the PR should be merged into `dev`, not `main`

vnbaaij:
Rolled up into #4680. Thanks again for your contribution!

PR #4678, opened 2026-04-07
[DataGrid] Skip debounce delay on first provider call when using virtualize

### LINKED ISSUE #4676: FluentDataGrid initial load performance
I'm using a FluentDataGrid with ItemsProvider and Virtualize=true.

I noticed there is a delay between the grid header being displayed and grid rows being displayed. The ItemsProvider callback isn't async. I would expect data to be displayed immediately.

I investigated and noticed that the ItemsProvider is always called multiple times on page load. And I noticed the second call is behind a Task.Delay:

https://github.com/microsoft/fluentui-blazor/blob/0625bb0d7a6e0909bebab4ca9cbf98d586b8817d/src/Core/Components/DataGrid/FluentDataGrid.razor.cs#L891-L912

Ideally there shouldn't be:
- Multiple calls to ItemsProvider (it doesn't matter much in my case because the data isn't that large or async)
- A forced 100ms delay before items are displayed.

JamesNK:
# Fix FluentDataGrid initial load performance

Fixes #4676

## Problem

`ProvideVirtualizedItemsAsync` unconditionally calls `await Task.Delay(100)` before every request, including the very first one on page load. This introduces an artificial 100ms delay before grid rows are displayed, even when the `ItemsProvider` callback is synchronous.

## Solution

Added a `_isFirstVirtualizeProviderCall` boolean field that skips the debounce delay on the first call. Subsequent calls (e.g., during scrolling) continue to be debounced as before.

Also fixed a pre-existing IDE0048 build error (missing parentheses for clarity) on the same file.

## Changes

- `FluentDataGrid.razor.cs`: Skip `Task.Delay(100)` on the first `ProvideVirtualizedItemsAsync` call
- `FluentDataGrid.razor.cs`: Add parentheses to `EffectiveLoadingValue` expression to fix IDE0048

### DIFF
--- a/src/Core/Components/DataGrid/FluentDataGrid.razor.cs
+++ b/src/Core/Components/DataGrid/FluentDataGrid.razor.cs
@@ -372,7 +372,7 @@ public partial class FluentDataGrid<TGridItem> : FluentComponentBase, IHandleEve
 
     // Returns Loading if set (controlled). If not controlled,
     // we assume the grid is loading until the next data load completes
-    internal bool EffectiveLoadingValue => Loading ?? ItemsProvider is not null;
+    internal bool EffectiveLoadingValue => Loading ?? (ItemsProvider is not null);
 
     private ElementReference? _gridReference;
     //private DotNetObjectReference<Type>? _dotNetObjectReference;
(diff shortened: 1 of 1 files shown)


vnbaaij (dismissed):

vnbaaij:
Hi @JamesNK please re-target this PR to go to the `dev` branch

vnbaaij:
Rolled up into #4680. Thanks again for your contribution!

PR #4684, opened 2026-04-08
[dev-v5] Fix publish demo locally script

dvoituron:
# [dev-v5] Fix publish demo locally script

- Update the script to build and next to publish
- Add PropertyParameterOnly option to SummaryDocumentationGenerator
- Add option to fix static assets manifest for Content-Length mismatches

### DIFF
--- a/_PublishDemoLocally.ps1
+++ b/_PublishDemoLocally.ps1
@@ -146,6 +146,45 @@ if ($fullBuild) {
 Write-Host "👉 Publishing demo..." -ForegroundColor Yellow
 dotnet publish "./examples/Demo/FluentUI.Demo/FluentUI.Demo.csproj" -c Release -o "./examples/Demo/FluentUI.Demo/bin/Publish" -f $NetVersion
 
+# Fix the static assets manifest to match actual file sizes on disk.
+# Brotli compression is non-deterministic, so the manifest Content-Length values
+# recorded during publish may not match the final compressed files.
+$fixManifest = Read-Host "❓ Do you want to fix the static assets manifest (Content-Length mismatches)? (y/n) [default: y]"
+if ($fixManifest -eq "n") {
+    Write-Host "⏩ Skipping static assets manifest fix." -ForegroundColor Yellow
+} else {
+    Write-Host "👉 Fixing static assets manifest..." -ForegroundColor Yellow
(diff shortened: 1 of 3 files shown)


_PublishDemoLocally.ps1
```diff
@@ -120,26 +122,69 @@ if ($node.InnerText -ne $NetVersion) {
...
-    dotnet build "./src/Core/Microsoft.FluentUI.AspNetCore.Components.csproj" -c Release -o "./src/Core/bin/Publish"  -f $NetVersion
-
-    # Generate API documentation file
-    Write-Host "👉 Generating API documentation..." -ForegroundColor Yellow
-    dotnet run --project ".\examples\Tools\FluentUI.Demo.DocApiGen\FluentUI.Demo.DocApiGen.csproj" --xml "$RootDir/src/Core/bin/Publish/Microsoft.FluentUI.AspNetCore.Components.xml" --dll "$RootDir/src/Core/bin/Publish/Microsoft.FluentUI.AspNetCore.Components.dll" --output "$RootDir/examples/Demo/FluentUI.Demo.Client/wwwroot/api-comments.json" --format json
+    dotnet build "./src/Core/Microsoft.FluentUI.AspNetCore.Components.csproj" -c Release -o "./src/Core/bin/Publish" -f $NetVersion
 
     # Build the MCP Server project
     Write-Host "👉 Building MCP Server project..." -ForegroundColor Yellow
+    dotnet build "./src/Core/Microsoft.FluentUI.AspNetCore.Components.csproj" -c Release -f $NetVersion
```
vnbaaij: Why are there now 2 builds of the same project immediately after each other

vnbaaij (changes requested):

_PublishDemoLocally.ps1
dvoituron: The output are not the same: first one to **Publish** folder and the second to **bin/Release** and used by the MCP Server

vnbaaij: Seems wasteful/time consuming to do it twice. Maybe just copy the DLL from the first step for the second step?

dvoituron: This is probably better. Updated.

vnbaaij (approved):

PR #4687, opened 2026-04-10
[dev-v5][NumberInput] Add FluentNumberInput component

dvoituron:
# [dev-v5] FluentNumber

The **FluentNumber** component enables a user to enter and edit numeric values.
It supports both integer and decimal numbers, with configurable `Min`, `Max`, and `Step` constraints.

**Value** can be one of the following **Numeric Types**:
`byte`, `sbyte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `float`, `double`, `decimal`.

Use the `StepButtons` parameter to control the visibility of the up/down step buttons.
It accepts a `NumberStepVisibility` value:
- **Visible**: The step buttons are always visible.
- **Hidden**: The step buttons are always hidden.
- **Auto** *(default)*: The step buttons are shown only when the user hovers over the input or when it has focus.

## Culture and Decimal Separator

Use the `Culture` parameter to control how numbers are formatted and parsed, including the decimal separator, group separator, and number of decimal digits. By default, the component uses the `FluentNumberCultureInfo` class (dot `.` as decimal separator). You can pass any `CultureInfo` instance (e.g. `fr`, `de`, `ja`) to match the user's locale.

https://github.com/user-attachments/assets/346c7706-7f16-4b94-8fee-99e121368f0e

## Unit Tests

image


MarvinKlein1508:
@dvoituron I look at this again and it is working perfectly fine for me. Great work! :)

src/Core/Components/Number/FluentNumber.razor
```diff
@@ -0,0 +1,60 @@
...
+            <div slot="@FluentSlot.Start">
+                @StartTemplate
+            </div>
+        }
+
+        <div slot="@FluentSlot.End">
+            @if (EndTemplate != null)
+            {
+                @EndTemplate
+
```
MarvinKlein1508: remove empty line

src/Core/Components/NumberInput/FluentNumberInput.razor.cs
```diff
@@ -0,0 +1,408 @@
...
+        return Task.CompletedTask;
+    }
+
+    /// <summary>
+    /// Increments the current value by the defined <see cref="Step"/>.
+    /// If the new value exceeds <see cref="Max"/>, it will be set to <see cref="Max"/>.
+    /// </summary>
+    private void IncrementValue()
+    {
+        if (Disabled == true || ReadOnly || Step == null || IsEquals(Step, ZeroValue))
```
MarvinKlein1508, line 273: we don't need `== true`

src/Core/Components/NumberInput/FluentNumberInput.razor.cs
```diff
@@ -0,0 +1,408 @@
...
+        CurrentValue = newValue;
+    }
+
+    /// <summary>
+    /// Decrements the current value by the defined <see cref="Step"/>.
+    /// If the new value is less than <see cref="Min"/>, it will be set to <see cref="Min"/>.
+    /// </summary>
+    private void DecrementValue()
+    {
+        if (Disabled == true || ReadOnly || Step == null || IsEquals(Step, ZeroValue))
```
MarvinKlein1508, line 294: we don't need `== true`

MarvinKlein1508 (changes requested):

src/Core/Enums/NumberInputStepVisibility.cs
vnbaaij, line 1: In line with the name change for the main component, this should be `NumberInputStepVisisbility`

src/Core/Components/NumberInput/FluentNumberInputCultureInfo.cs
vnbaaij, line 1: Should also be renamed (-> `FluentNumberInputCultureInfo`)

vnbaaij (changes requested):
We should call this component `FluentNumberInput` to make clear it is an input and acts as a counterpart of the `FluentTextInput`

src/Core/Components/NumberInput/FluentNumberInput.razor.cs
dvoituron-msft: Yes we need it, because Disabled is nullable

src/Core/Components/NumberInput/FluentNumberInput.razor.cs
dvoituron: Yes we need it, because Disabled is nullable

vnbaaij (approved):

MarvinKlein1508 (approved):

Issue #4380, opened 2025-12-05
fix: SelectedItem can't be set to option not in Items list in AutoComplete


larahina25:
# 🐛 Bug Report

I'm not totally sure if this should better be reported as Bug or Feature Request, but I think the current implementation leads to undesired behavior in some cases, so I chose the bug...

I'm using the `FluentAutocomplete` component bound to some value, that can also be changed programmatically from the outside of the component. This works well as long as the Autocomplete component hasn't been touched by the user (in this case the `OnOptionsSearch` method hasn't been called and the `Items` list that holds all available option is null). But when the user opens the option list by clicking into the component or even reduces the option list by typing some search term, it is no longer possible to programmatically set the SelectedItem to an option that is not in the current `Items` list.

## 💻 Repro or Code Sample

A repo for reproducing the issue is provided here:
https://github.com/larahina25/FluentUIBlazorTests/tree/autocomplete-selecteditemnotinitems

## 🤔 Expected Behavior

I expected that it would be still possible to set the SelectedItem programmatically.

## 😯 Current Behavior

Described in description and in sample repo in more detail.

## 💁 Possible Solution

I think the key method of this behavior is the `SetParametersAsync` overload in the `ListComponentBase` class. Here's the section I'm talking about:

``` csharp
if (isSetSelectedOption && !Equals(_currentSelectedOption, newSelectedOption))
                {
                    if (Items != null)
                    {
                        if (Items.Contains(newSelectedOption))
                        {
                            _currentSelectedOption = newSelectedOption;
                            // Make value follow new selected option
                            Value = GetOptionValue(_currentSelectedOption);
                            await ValueChanged.InvokeAsync(Value);
                        }
                        else
                        {
                            // If the selected option is not in the list of items, reset the selected option
                            _currentSelectedOption = SelectedOption = default;
                            // and also reset the value
                            Value = null;
                            await SelectedOptionChanged.InvokeAsync(SelectedOption);
                        }
                    }
                    else
                    {
                        // If Items is null, we don't know if the selected option is in the list of items, so we just set it
                        _currentSelectedOption = newSelectedOption;
                    }
                }
```

I think this behavior makes sense for other components, but for the Autocomplete it leads to undesired behavior as described. Maybe it can be changed for the autocomplete component only. Or an addition parameter added to the Autocomplete that disables this behavior which can be activated. This way the current behavior would stay the default, but could be changed if one runs into issues like me.

## 🌍 Your Environment

[...]

PR #4689, opened 2026-04-11
Autocomplete SelectedOption issue #4380

ehsangfl:
# Pull Request

## 📖 Description
If the user clicks on autocomplete (or any action that invokes InvokeOptionsSearchAsync), then the SelectedOption is set outside of the component, and SelectedOptions does not exist in the top options (default 9 top options), the component clears the SelectedOption

## 📑 Test Plan

I updated an existing example for a test 
You can find it [here](https://github.com/ehsangfl/fluentui-blazor/blob/ehsangfl-patch-1/examples/Demo/Shared/Pages/List/Autocomplete/Examples/AutoCompleteMaxSingleItem.razor)

### General

### Component-specific

### DIFF
--- a/src/Core/Components/List/FluentAutocomplete.razor.cs
+++ b/src/Core/Components/List/FluentAutocomplete.razor.cs
@@ -27,6 +27,7 @@ public partial class FluentAutocomplete<TOption> : ListComponentBase<TOption> wh
     private readonly Debounce _debounce = new();
     private bool _shouldRender = true;
     private bool _inProgress;
+    private IEnumerable<TOption> _allItems;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="FluentAutocomplete{TOption}"/> class.
@@ -279,6 +280,38 @@ public partial class FluentAutocomplete<TOption> : ListComponentBase<TOption> wh
     /// <summary />
     protected override bool ShouldRender() => _shouldRender;
 
+    public override async Task SetParametersAsync(ParameterView parameters)
+    {
+        if (!Multiple)
+        {
+            bool isSetSelectedOption = false;
+            TOption? newSelectedOption = default;
+
+            foreach (var parameter in parameters)
+            {
+                if (parameter.Name.Equals(nameof(SelectedOption)))
+                {
+                    isSetSelectedOption = true;
+                    newSelectedOption = (TOption?)parameter.Value;
+                    break;
+                }
+            }
+
+            if (newSelectedOption is not null)
+            {
+                if (isSetSelectedOption && !Equals(_currentSelectedOption, newSelectedOption))
+                {
+                    if (_allItems is not null && Items is not null
+                        && _allItems.Contains(newSelectedOption) && !Items.Contains(newSelectedOption))
+                    {
+                        Items = Items.Append(newSelectedOption)!;
+                    }
+                }
+            }
+        }
+        await base.SetParametersAsync(parameters);
+    }
+
     /// <summary>
     /// Closes the multiselect dropdown.
     /// </summary>
@@ -341,8 +374,17 @@ public partial class FluentAutocomplete<TOption> : ListComponentBase<TOption> wh
         };
 
         await OnOptionsSearch.InvokeAsync(args);
+        _allItems = args.Items;
 
-        Items = args.Items?.Take(MaximumOptionsSearch);
+        var topItems = _allItems.Take(MaximumOptionsSearch);
+        if (!Multiple && SelectedOption is not null && _allItems is not null && Items is not null)
+        {
+            if (_allItems.Contains(SelectedOption) && !topItems.Contains(SelectedOption))
+            {
+                topItems = _allItems.Take(MaximumOptionsSearch - 1).Append(SelectedOption);
+            }
+        }
+        Items = topItems;
 
         SelectableItem = Items != null
             ? Items.FirstOrDefault(i => OptionDisabled is null ? true : OptionDisabled.Invoke(i) == false)


dvoituron (changes requested):
1. I need to analyze more in details what is the problem (the related issue is closed since 4 months)

2. Using `SetParameterAsync` is dangerous because this method is raised each time a parameter is updated. So you can decrease the performances. 

3. A unit tests must be added 😊

MarvinKlein1508:
@dvoituron I have taken a look on the sample he has provided. 

```razor
@inject DataSource Data

<FluentAutocomplete TOption="Country"
                    AutoComplete="off"
                    Label="Select a country"
                    Width="250px"
                    Placeholder="Select a country"
                    OnOptionsSearch="@OnSearchAsync"
                    OptionDisabled="@(e => e.Code == "au")"
                    Multiple=false
                    OptionText="@(item => item.Name)"
                    @bind-SelectedOption=SelectedItem />

<p>
    <b>Selected</b>: @(SelectedItem?.Name)
</p>

<FluentButton OnClick="SetSelectedItem">Set Zambia</FluentButton>

@code
{
    Country? SelectedItem = null;
    IEnumerable<Country>? AllItems = null;

    protected override async Task OnInitializedAsync()
    {
        AllItems = await Data.GetCountriesAsync();
        await base.OnInitializedAsync();
    }

    private void OnSearchAsync(OptionsSearchEventArgs<Country> e)
    {
        e.Items = AllItems?.Where(i => i.Name.StartsWith(e.Text, StringComparison.OrdinalIgnoreCase))
                              .OrderBy(i => i.Name);
    }

    void SetSelectedItem()
    {
        SelectedItem = AllItems?.Last();
        //StateHasChanged();
    }
}
```

Basically the following is happening:

1. When you click the button first then the value will be updated to Zambia
2. Once you click on the `FluentAutocomplete` component and the menu has been showed then you can no longer set the value from outside of the component again. 
3. If you click on the component before you click on the button then you cannot set the value from the button as well
4. If you change the code from `.Last` to `.First` then the `FluentAutocomplete` component will update correctly in both cases.

MarvinKlein1508:
This is happening because of this code block here:
https://github.com/microsoft/fluentui-blazor/blob/64d0d50b9cd211b2a576793b93b377cb02c3c0f5/src/Core/Components/List/ListComponentBase.razor.cs#L265

The base class has figured out that the `SelectedOption` has been changed. Therefore we are landing in this block

```cs
if (isSetSelectedOption && !Equals(_currentSelectedOption, newSelectedOption))
{
    if (Items != null)
    {
        if (Items.Contains(newSelectedOption))
        {
            _currentSelectedOption = newSelectedOption;
            // Make value follow new selected option
            Value = GetOptionValue(_currentSelectedOption);
            await ValueChanged.InvokeAsync(Value);
        }
        else
        {
            // If the selected option is not in the list of items, reset the selected option
            _currentSelectedOption = SelectedOption = default;
            // and also reset the value
            Value = null;
            await SelectedOptionChanged.InvokeAsync(SelectedOption);
        }
    }
    else
    {
        // If Items is null, we don't know if the selected option is in the list of items, so we just set it
        _currentSelectedOption = newSelectedOption;
    }
}
```
In this case `Items` is only then `null` as long as the user haven't interacted with the component in anyway. Providing all Items via `Items` parameter is fixing this (but most likely not capable as you have to store all items in memory then). We want to reach the else block here 
```cs
_currentSelectedOption = newSelectedOption;
```

which is the reason why it's working in the first place before the user interacts with the component.

ehsangfl:
@MarvinKlein1508 
In the sample I provided, you should interact with component, the dropdown is shown. So, the "zambia" item does not exists in the list (because the list is top 9). then click on "Set Zambia" button to select . then AutoComplete sets the "Zambia" as the SelectedOption
in previous version, we can just set SelectedOption from "Top(9)" (Maximum option)

ehsangfl:
> This is happening because of this code block here:
> 
> https://github.com/microsoft/fluentui-blazor/blob/64d0d50b9cd211b2a576793b93b377cb02c3c0f5/src/Core/Components/List/ListComponentBase.razor.cs#L265
> 
> The base class has figured out that the `SelectedOption` has been changed. Therefore we are landing in this block
> 
> ```cs
> if (isSetSelectedOption && !Equals(_currentSelectedOption, newSelectedOption))
> {
>     if (Items != null)
>     {
>         if (Items.Contains(newSelectedOption))
>         {
>             _currentSelectedOption = newSelectedOption;
>             // Make value follow new selected option
>             Value = GetOptionValue(_currentSelectedOption);
>             await ValueChanged.InvokeAsync(Value);
>         }
>         else
>         {
>             // If the selected option is not in the list of items, reset the selected option
>             _currentSelectedOption = SelectedOption = default;
>             // and also reset the value
>             Value = null;
>             await SelectedOptionChanged.InvokeAsync(SelectedOption);
>         }
>     }
>     else
>     {
>         // If Items is null, we don't know if the selected option is in the list of items, so we just set it
>         _currentSelectedOption = newSelectedOption;
>     }
> }
> ```
> 
> In this case `Items` is only then `null` as long as the user haven't interacted with the component in anyway. Providing all Items via `Items` parameter is fixing this (but most likely not capable as you have to store all items in memory then). We want to reach the else block here
> 
> ```cs
> _currentSelectedOption = newSelectedOption;
> ```
> 
> which is the reason why it's working in the first place before the user interacts with the component.

I dont change the base class because its behavior is right. in the Autocomplete, the Items property becomes Top(9), and this issue happens just for the AutoCompele.

vnbaaij:
Closing this as the current SetParameterAsync  approach is invalid and unit tests are not there

PR #4719, opened 2026-04-16
[dev-v5][Docs] Extend the DocApiGen and DocViewer projects

vnbaaij:
This pull request refactors the documentation generator to support multiple documentation inputs and assemblies. You can now specify multiple dll and xml sources to generate one combined api-comments.json file. 
It replaces the old single-assembly generator creation with a new approach using `DocumentationInput` and `GenerationMode`. Some utility methods and test setup logic are also improved for better robustness. For instance, we now use a specialized (isolated) AssemblyLoadContext so errors about mismatching assembly manifests are now prevented.
The PR also updates the `MarkdownViewer` to read from multiple assembly documentation sources

**Test Infrastructure Improvements:**

* Refactored test setup to use `DocumentationInput` and `GenerationMode` with the `DocumentationGeneratorFactory.Create` method, replacing the previous single-assembly generator creation methods. This enables easier extension to multiple assemblies and modes. 

* Added a new test (`AllGenerator_WithMultipleDocumentationInputs_ShouldIncludeChartsAssemblyTypes`) that verifies the generator can handle multiple assemblies and documentation files, ensuring types from both the core and charts assemblies are included in the output.

**Utility and Robustness Improvements:**

* Centralized assembly path resolution with a new `GetAssemblyPath` method, supporting both Debug and Release builds, improving reliability when locating assemblies for tests.

* Updated project root directory detection to look for `.slnx` files instead of `.sln`, reflecting the current solution file extension.

**Demo Service Registration:**

* Updated `AddFluentUIDemoServices` to use the `ApiAssemblies` property (now an array) instead of the deprecated `ApiAssembly` property, aligning with new API expectations.

### DIFF
--- a/examples/Tools/FluentUI.Demo.DocApiGen/DocumentationAssemblyLoadContext.cs
+++ b/examples/Tools/FluentUI.Demo.DocApiGen/DocumentationAssemblyLoadContext.cs
@@ -1,0 +1,83 @@
+// ------------------------------------------------------------------------
+// This file is licensed to you under the MIT License.
+// ------------------------------------------------------------------------
+
+using System.Reflection;
+using System.Runtime.Loader;
+
+namespace FluentUI.Demo.DocApiGen;
+
+/// <summary>
+/// Loads documentation target assemblies and their dependencies in an isolated context.
+/// </summary>
+public sealed class DocumentationAssemblyLoadContext : AssemblyLoadContext
+{
+    private readonly Dictionary<string, string> _inputAssemblyPaths;
+    private readonly IReadOnlyList<AssemblyDependencyResolver> _resolvers;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="DocumentationAssemblyLoadContext"/> class.
+    /// </summary>
+    /// <param name="assemblyPaths">The assembly paths that can be loaded as documentation inputs.</param>
+    public DocumentationAssemblyLoadContext(IEnumerable<string> assemblyPaths)
+        : base(nameof(DocumentationAssemblyLoadContext), isCollectible: false)
+    {
+        _inputAssemblyPaths = assemblyPaths
+            .ToDictionary(
+                path => Path.GetFileNameWithoutExtension(path),
+                StringComparer.OrdinalIgnoreCase);
+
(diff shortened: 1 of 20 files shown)


examples/Tools/FluentUI.Demo.DocApiGen/DocumentationAssemblyLoadContext.cs
```diff
@@ -0,0 +1,83 @@
...
+    /// <param name="assemblyPaths">The assembly paths that can be loaded as documentation inputs.</param>
+    public DocumentationAssemblyLoadContext(IEnumerable<string> assemblyPaths)
+        : base(nameof(DocumentationAssemblyLoadContext), isCollectible: false)
+    {
+        _inputAssemblyPaths = assemblyPaths
+            .ToDictionary(
+                path => Path.GetFileNameWithoutExtension(path),
+                StringComparer.OrdinalIgnoreCase);
+
+#pragma warning disable CA1416 // Validate platform compatibility
```
dvoituron, line 30: What will be the result when this part of code will be executed on Linux (during the publication process in ADO) ?
`AssemblyDependencyResolver` seems not acceptable on Linux (not tested)

examples/Tools/FluentUI.Demo.DocApiGen/Program.cs
```diff
@@ -145,6 +142,9 @@ private static void ShowHelp()
         Console.WriteLine("  # Generate All mode JSON");
         Console.WriteLine("  DocApiGen --xml MyApp.xml --dll MyApp.dll --output api-all.json --mode all");
         Console.WriteLine();
+        Console.WriteLine("  # Generate Summary mode JSON from multiple assemblies");
+        Console.WriteLine("  DocApiGen --xml Core.xml;Charts.xml --dll Core.dll;Charts.dll --output api-summary.json");
```
dvoituron, line 146: Why not to use another mode `--mode chart` like for Icons, Emojis or MCP?
That will use the same logic as already used.
And in the ADO Pipelinem we need to add a new task to run the Chart generation. That's all :-)

dvoituron (changes requested):

PR #4843, opened 2026-05-06
[dev-v5][Dialog] Add PreventDissmissOnEscape parameter to dialog options and dialog to…

### LINKED ISSUE #4830: feat: add conditional closing to/in FluentDrawer/FluentDialog
# 🙋 Feature Request

Create CanClose event to prevent closing drawer/dialog.

## 🤔 Expected Behavior

When user/program want to close drawer/dialog there should be central event - f.e. CanClose. If we decide that closing is not allowed, drawer/dialog can't be closed. This even should fire on all closing actions (ESC, click outside - overlay, click on top header closing action or calling  dialogInstance methods (CancelAsync, CloseAsync, etc). 

## 😯 Current Behavior

When user press ESC or click outside drawer or top right "X" button, drawer is closed without possibility to stop closing. 

## 💁 Possible Solution

Create event CanClose(bool canClose, string result). This event sould be fired on all closing actions.
By default canClose is true. If we set this value to false closing execution is stopped and drawer/dialog remains open.

## 🔦 Context

Drawer/dialog could be containers of forms. Exiting without control causes data loss in editforms.
We need such event to test if data were changed. And if data were changed in need a way to prevent closing
if user decide so.

Like github in this dialog.

For our future application, this is a feature that ranks very high in priority, as data entry is a very important feature.

## 💻 Examples

Image

vnbaaij:
…...prevent dismiss on escape key

Plus add unit tests and extend panel example

Fix #4830


src/Core/Components/Dialog/FluentDialog.razor.cs
```diff
@@ -71,6 +71,14 @@ public FluentDialog(LibraryConfiguration configuration) : base(configuration)
     [Parameter]
     public bool Modal { get; set; } = true;
 
+    /// <summary>
+    /// Gets or sets a value indicating whether pressing the ESC key should be prevented from closing the dialog.
+    /// By default, the ESC key closes the dialog (<c>false</c>).
+    /// When using the <see cref="IDialogService"/>, set <see cref="DialogOptions.PreventDismissOnEscape"/> instead.
+    /// </summary>
+    [Parameter]
+    public bool PreventDismissOnEscape { get; set; } = false;
```
MarvinKlein1508: We don't need to specify `false` - this is the default value.

dvoituron (changes requested):
1. Have you verified this behavior when `DialogOptionsFooterAction.ShortCut` is set?
2. Why did you removed this test?

src/Core/Components/Dialog/FluentDialog.razor.cs
vnbaaij: FWNC

vnbaaij:
> 1) Have you verified this behavior when DialogOptionsFooterAction.ShortCut is set?

Yes, tested with `PrevenDismissOnEscpe=true` and `options.Footer.SecondaryAction.ShortCut = "Ctrl+Enter";` results in the dialog **not** being closed with pressing **ESC** but **closed** with **Ctrl+Enter**

> 2) Why did you removed this test?

That was a pasting error. Test is restored

dvoituron (approved):

PR #4923, opened 2026-06-13
[dev-v5] Add FluentValidationMessage

### LINKED ISSUE #4728: [dev-v5] fix: change does not trigger validation in `FluentSelect`
# 🐛 Bug Report
When switching the value in `FluentSelect` then the validation for the current bound value will not retrigger.

Please take a look on this video:

https://github.com/user-attachments/assets/6b70fdc7-744b-4f50-9b3e-84536ac02564

## 💻 Repro or Code Sample

```razor
@page "/Debug/Test"
@layout FluentUI.Demo.Client.Layout.EmptyLayout


<div style="padding: 1rem">
    <FluentButton OnClick="Save">Submit</FluentButton>
    <EditForm @ref="_form" Model="Input" OnValidSubmit="OnValidSubmit">
        <DataAnnotationsValidator />
        <FluentValidationSummary />
        <FluentSelect Label="Location"
                      TOption="CompanyLocation"
                      TValue="int?"
                      @bind-Value="@Input.LocationId">
            <FluentOption TValue="int?" Value="@(null)" Text="--- SELECT ---" />
            @foreach (var item in Locations)
            {
                <FluentOption TValue="int?" Value="@item.Id">
                    @item.Name
                </FluentOption>
            }
        </FluentSelect>
        <ValidationMessage For="() => Input.LocationId" />

        <div>
            <label>Location 2</label>
            <InputSelect @bind-Value="@Input.LocationId2">
                <option value="">--- SELECT ---</option>
                @foreach (var item in Locations)
                {
                    <option value="@item.Id">
                        @item.Name
                    </option>
                }
            </InputSelect>
            <ValidationMessage For="() => Input.LocationId2" />
        </div>
    </EditForm>
</div>

@code {
    private EditForm _form = default!;
    public MyInput Input { get; set; } = new();

    Task Save()
    {
        _form.EditContext!.Validate();

        return Task.CompletedTask;
    }

    Task OnValidSubmit()
    {
        return Task.CompletedTask;
    }

    public IEnumerable<CompanyLocation> Locations { get; set; } = new List<CompanyLocation>
    {
        new CompanyLocation { Id = 1, Name = "Germany" },
        new CompanyLocation { Id = 2, Name = "Netherlands" },
        new CompanyLocation { Id = 3, Name = "Belgium" }
    };

    public class MyInput
    {
        [System.ComponentModel.DataAnnotations.Required]
        public int? LocationId { get; set; }

        [System.ComponentModel.DataAnnotations.Required]
        public int? LocationId2 { get; set; }
    }

    public class CompanyLocation
    {
        public int Id { get; set; }
        public required string Name { get; set; }
    }
}
```

1. Open the page
2. Click submit
3. Notice two validation errors
4. Switch the value in the native control
5. See validation message is disappearing
6. Switch the value in `FluentSelect`
7. Notice the validation message is still there.

## 🤔 Expected Behavior

I expect the `FluentSelect` to behave exactly like `InputSelect` in this case.

## 🌍 Your Environment

* Browser latest Firefox, Chrome & Edge
* latest dev-v5 branch

### LINKED ISSUE #4908: fix: No show error validation message in Field
# 🐛 Bug Report

In V4, if a form field is not valid, FluentValidationMessage could be used to display the validation error. Also, the input field itself was wrapped in a red frame. However, in V5, the FluentValidationMessage component was removed and, according to the documentation https://fluentui-blazor-v5.azurewebsites.net/Migration/Field , you need to use the FluentField component with the ValidationFor parameter. However, there is no such parameter, and the validation error text does not appear under the field, and the text field does not turn into a red border.

## 💻 Repro or Code Sample
```razor
@page "/test"
@using System.ComponentModel.DataAnnotations

<EditForm Model="@EditForm" OnValidSubmit="@HandleValidSubmit" FormName="starship_fluent_entry" novalidate>
    <DataAnnotationsValidator />
    <FluentValidationSummary />

    <FluentStack Orientation="Orientation.Vertical">
        <div>
            <FluentField Label="Name"
                         Required="true"
                         ValidationFor="@(() => EditForm.Name)">
                <FluentTextInput @bind-Value="EditForm.Name" Width="100%" />
            </FluentField>
        </div>
        <FluentButton Type="ButtonType.Submit" Appearance="ButtonAppearance.Primary">Submit</FluentButton>
    </FluentStack>

    
</EditForm>

@code {
    [SupplyParameterFromForm]
    private TestEditForm EditForm { get; set; } = new();

    class TestEditForm
    {
        [Required]
    }
    
    private void HandleValidSubmit()
    {
        Console.WriteLine("HandleValidSubmit called");
    }
}
```
## 🤔 Expected Behavior

The field will turn into a red frame and a text with a validation error will appear below it.

## 😯 Current Behavior

None of the above. The text of the form validation error is displayed only in FluentValidationSummary.

## 🌍 Your Environment

* OS & Device: Linux
* Browser Google Chrome
* .NET and Fluent UI Blazor library Version: .net - 10.0.5, Fluent UI - 5.0.0-rc.3-26138.1

vnbaaij:
Fix #4728 

- Includes tests and adjustments to make it work with more inputs. 
- Use `DataAnnotations` message instead of default message when set.
- Re-evaluates state on change/blur/loose focus

## Before:

image

## After
image

Because hooking up the message is currently in `ChangeHandlerAsync`, that only runs on the component’s change/input event.
If you just submit a form without changing anything, on submit Blazor does:
- EditContext.Validate() (DataAnnotations, etc.)
- updates validation messages in EditContext
- 
But it does not fire each input’s native change event.

So untouched fields never execute ChangeHandlerAsync → You do not get a red border.
For untouched fields, the web components also often won’t set `:user-invalid` until it considers the control “user interacted,” which submit alone may not satisfy. So the behavior is expected with the current implementation location.


src/Core/Components/Field/FluentField.razor.cs
```diff
@@ -49,6 +67,22 @@ public FluentField(LibraryConfiguration configuration) : base(configuration) { }
     [Parameter]
     public IFluentField? InputComponent { get; set; }
 
+    /// <summary>
+    /// Gets or sets the <see cref="FieldIdentifier"/> for which validation messages should be displayed.
+    /// If set, this parameter takes precedence over <see cref="For"/>.
+    /// </summary>
+    [Parameter]
+    public FieldIdentifier? Field { get; set; }
```
dvoituron: Why did you add `Field` as a parameter? I can't find any component that uses it as a parameter.
Maybe you can use a private property?

src/Core/Components/Field/FluentField.razor
```diff
@@ -60,6 +60,12 @@ else
                 }
             }
 
+            @if (HasValidationMessages)
+            {
+                <FluentValidationMessage TValue="object"
```
dvoituron, line 65: Shouldn't we add a slot to specify where this `FluentValidationMessage` should be placed? Or combine it with `<FluentText slot="@FluentSlot.FieldMessage"`

src/Core/Components/Field/FluentField.razor.cs
```diff
@@ -49,6 +67,22 @@ public FluentField(LibraryConfiguration configuration) : base(configuration) { }
...
+
+    /// <summary>
+    /// Gets or sets the field expression for which validation messages should be displayed.
+    /// </summary>
+    [Parameter]
+    public LambdaExpression? For { get; set; }
```
dvoituron: I think that name is weird. 
Is it required to have a public parameter (in which situation the dev will use it)?

I found only 1 usage in `FluentTextInput` and you have a reference to the component via the property`InputComponent="@this"`.
It might be possible to avoid confusion among developers if this parameter were internal only.

src/Core/Components/Base/FluentInputBase.cs
```diff
@@ -209,6 +220,31 @@ protected virtual async Task ChangeHandlerAsync(ChangeEventArgs e)
         {
             // TODO
         }
+
+        await ReportValidityAsync();
```
dvoituron, line 224: Do we need to run this method and call the JS function also when the `isValid == true` (in all cases, and not only when the validation fails)?

dvoituron (changes requested):

src/Core/Components/Field/FluentField.razor.cs
vnbaaij: Field is not needed. Removed in next commit

src/Core/Components/Field/FluentField.razor
vnbaaij: The '<FluentText slot="@FluentSlot.FieldMessage"' is what is being used inside the `FluentValidationMessage`

vnbaaij:
Comments addressed. I have not done the border thing on immediate submit as it will then deviate from how the web components work.

dvoituron (approved):

PR #4934, opened 2026-06-16
[dev-v5][Templates] 

vnbaaij:
- Add separate Templates solution file
- Fix WASM templates MainLayout


Microsoft.FluentUI-v5.Templates.slnx
```diff
@@ -0,0 +1,5 @@
+<Solution>
```
dvoituron, line 1: Can you rename this file to `Microsoft.FluentUI-v5.Templates.slnx` to keep the file name coventions?

dvoituron (changes requested):

dvoituron (approved):

Issue #4763, opened 2026-04-29
[dev-v5] fix: `LabelPosition` in `FluentTextInput`


MarvinKlein1508:
# 🐛 Bug Report
When using `LabelPosition.Before` in any input component then the inputs will overlap each other when you use the `Width` parameter.

## 💻 Repro or Code Sample
```razor
@page "/Debug/Test"
@layout FluentUI.Demo.Client.Layout.EmptyLayout

<h3>Test</h3>


<div style="padding: 1rem">
    <FluentStack HorizontalGap="4px">
        <FluentTextInput @bind-Value="test"
                         Label="Werk"
                         LabelPosition="LabelPosition.Before"
                         Width="100px"
                         Placeholder="Name" />

        <FluentTextInput @bind-Value="test2"
                         Label="Suche"
                         LabelPosition="LabelPosition.Before"
                         Width="300px"
                         Placeholder="Artikelnummer, Auftragsnummer"
                         AutoComplete="off" />

        <FluentTextInput @bind-Value="test2"
                         Label="Company"
                         LabelPosition="LabelPosition.Before"
                         Width="200px"
                         Placeholder="Name" />
    </FluentStack>
</div>
@code {
    public string test { get; set; } = "test";
    public string test2 { get; set; } = "test2";
    public string test3 { get; set; } = "test3";
}
```

## 🤔 Expected Behavior

I expect all of the items to be displayed next to each other.

## 😯 Current Behavior

<img width="787" height="187" alt="Image" src="https://github.com/user-attachments/assets/f4b10b12-2651-43e6-816a-2d1d4c173abb" />

## 🌍 Your Environment

* OS & Device: Windows 11 64 Bit
* Browser latest Firefox, Chrome & Edge
*  latest dev-v5 branch

PR #4942, opened 2026-06-17
[dev-v5] Update `Width` handling for FluentField

### LINKED ISSUE #4938: Wrong width for FluentTextInput in v5
[FluentTextInput-4938.zip](https://github.com/user-attachments/files/29008983/FluentTextInput-4938.zip)

# 🐛 Bug Report

I have just migrated from v4 to v5 (and by that I mean Claude has migrated...)
The input fields in the application now have wrong sizes, for example this FluentTextInput with width 100%:
```
<FluentTextInput Label="Hello" Style="width:100%;"></FluentTextInput>
```

It should take up the full width, but doesn't do that:
Image

The problem is that only the outer **fluent-field** gets the width 100% and the inner **fluent-text-input** has default width:
Image

However if i use the Width parameter like so:
```
<FluentTextInput Label="Hello" Width="50%"></FluentTextInput>
```

Then the width is applied to both and the result is 25%

Image

Image

## 🤔 Expected Behavior

I would expect to just use Style or Width to control the width of input fields.

## 😯 Current Behavior

Only the outer fluent-field is controllable with the Style, the inner fluent-text-input has default size or the size set by Width.

## 💁 Possible Solution

I would prefer just using Style because that can be used to things like flex-grow also.
If the inner  **fluent-text-input** just always have width:100% then it seems to work?

## 🔦 Context

Pretty much messes up my whole application ;-)

## 🌍 Your Environment

.NET 10, Blazor WASM, Windows 11, Firefox+Edge.

dvoituron:
# [dev-v5] Update `Width` handling for FluentField

Enhance the default `Width` handling for various **Fluent Input** components to 100% for improved layout and responsiveness. Introduce new components with `Width` parameters and refactor existing components to streamline width management. 

This change aims to provide a more consistent and user-friendly interface across the application.

To fix #4938

## Example

image

image

image

## Demo

See: `/Debug/Width`

https://github.com/user-attachments/assets/5a5ab164-4a53-44fc-80d2-43f63faa0e9c

### DIFF
--- a/examples/Demo/FluentUI.Demo.Client/DebugPages/WidthParameterPage.razor
+++ b/examples/Demo/FluentUI.Demo.Client/DebugPages/WidthParameterPage.razor
@@ -52,8 +52,13 @@
         new("Selects", RenderSelect),
         new("Listboxes", RenderListbox),
         new("Comboboxes", RenderCombobox),
+        new("Calendars", RenderCalendar),
         new("Date Pickers", RenderDatePicker),
         new("Time Pickers", RenderTimePicker),
+        new("Color Picker Inputs", RenderColorPickerInput),
+        new("Radio Groups", RenderRadioGroup),
+        new("Sliders", RenderSlider),
+        new("Checkboxes", RenderCheckbox),
         new("Switches", RenderSwitch),
     ];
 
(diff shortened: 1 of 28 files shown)


MarvinKlein1508:
This might also fix #4763 

Will validate this tomorrow

MarvinKlein1508:
@dvoituron I tested this changes for #4763 and this issue is indeed fixed.

vnbaaij (approved):

PR #4960, opened 2026-06-24
Fix symbol-publish gate in build-all-lib.yml to use a branch allowlist

tlmii:
# Pull Request

## 📖 Description

Follow-up to #4910, which added internal symbol publishing to the two CI pipelines.

In `eng/pipelines/build-all-lib.yml`, the two symbol-publishing template invocations were gated on:

```
and(succeeded(), startsWith(variables['Build.SourceBranch'], 'refs/tags/v'))
```

The pipeline's declared trigger looks tag-based (`tags: include: v*`), but in practice this pipeline is always **queued manually against a release branch HEAD** (e.g. after tagging) rather than via a tag push. As a result the `refs/tags/v*` condition never evaluated true and symbols were never actually published from this pipeline.

This change switches both conditions to the branch allowlist already used successfully by `eng/pipelines/build-core-lib.yml` (`main` / `dev` / `dev-v5` / `archives/*`), OR-ing in the `refs/tags/v*` check so the declared tag-trigger path stays covered:

```
and(succeeded(), or(in(variables['Build.SourceBranch'], 'refs/heads/main', 'refs/heads/dev', 'refs/heads/dev-v5'), startsWith(variables['Build.SourceBranch'], 'refs/heads/archives/'), startsWith(variables['Build.SourceBranch'], 'refs/tags/v')))
```

The stale inline comment that claimed the gate existed to prevent manual branch queues from uploading (the opposite of the desired behavior) was updated to match.

### 🎫 Issues

Follow-up to #4910.

## 👩‍💻 Reviewer Notes

Surgical change — only the two conditions and one comment in `build-all-lib.yml` are touched. `build-core-lib.yml`, `publish-symbols-stage.yml`, and `publish-symbols-upload.yml` are unchanged. The new condition is intentionally identical to the one in `build-core-lib.yml` (plus the tag check).

## 📑 Test Plan

No code changes; CI pipeline YAML only. The pipeline is internal and cannot be exercised from this PR. The conditions were verified to match `build-core-lib.yml`'s working gate.

### General

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.8) on behalf of @tlmii</em></sub>

### DIFF
--- a/eng/pipelines/build-all-lib.yml
+++ b/eng/pipelines/build-all-lib.yml
@@ -182,11 +182,12 @@ extends:
 
         # Stage shipping binaries for symbol upload BEFORE the test step;
         # see publish-symbols-stage.yml for why this must run pre-test.
-        # Pipeline only triggers on v* tags, but condition is explicit so
-        # a manual queue against a branch does not unintentionally upload.
+        # Releases are cut by manually queueing this pipeline against a release
+        # branch HEAD, so gate on the branch allowlist (consistent with
+        # build-core-lib.yml); the tag check covers the declared trigger.
         - template: /eng/pipelines/publish-symbols-stage.yml@self
           parameters:
-            condition: and(succeeded(), startsWith(variables['Build.SourceBranch'], 'refs/tags/v'))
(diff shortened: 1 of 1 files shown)


dvoituron (approved):

vnbaaij:
Do we need to release a new v4 version to make the symbols appear? Or can it be triggered manually? 
New version is not ideal but also not impossible to do. It will probably be a while before we do another v4 release.

PR #4961, opened 2026-06-25
[dev-v5][Overflow] refactor to wrap web component

### LINKED ISSUE #4949: FluentOverflow: Add MaxRenderedItems parameter to avoid rendering all children as DOM elements
## Summary

`FluentOverflow` currently renders **all** `FluentOverflowItem` children as DOM elements, then uses JavaScript interop (MutationObserver/ResizeObserver) to detect which ones overflow. When there are hundreds or thousands of items, this forces an expensive browser reflow that can block the UI thread long enough to kill a Blazor Server SignalR connection (~10+ seconds of unresponsive layout).

## Problem

In the Aspire Dashboard, we use `FluentOverflow` to display resource URLs and metric filter tags. Some resources can have hundreds of URLs or filter dimension values. Rendering all of them as DOM elements triggers a forced reflow that's catastrophically expensive — in our case, it disconnects the SignalR connection entirely.

The component only ever **displays** a handful of items (the ones that fit in the available space), so rendering all items into the DOM just to measure which ones overflow is wasteful when the item count is large.

## Current Workaround

We manually cap the number of items passed into `ChildContent` and track a "pre-overflowed" count to include in the `+N` badge:

```razor
@{
    var maxRenderedUrls = 20;
    var renderedUrls = DisplayedUrls.Take(maxRenderedUrls).ToList();
    var preOverflowedCount = DisplayedUrls.Count - renderedUrls.Count;
}

<FluentOverflow>
    <ChildContent>
        @for (var i = 0; i < renderedUrls.Count; i++)
        {
            <FluentOverflowItem Data="renderedUrls[i]">...</FluentOverflowItem>
        }
    </ChildContent>
    <MoreButtonTemplate Context="overflow">
        <FluentButton>@($"+{overflow.ItemsOverflow.Count() + preOverflowedCount}")</FluentButton>
    </MoreButtonTemplate>
    <OverflowTemplate Context="overflow">
        @* Must manually concat pre-overflowed items into the popup *@
        @foreach (var item in overflow.ItemsOverflow.Select(i => i.Data).Concat(allItems.Skip(maxRenderedUrls)))
        {
            ...
        }
    </OverflowTemplate>
</FluentOverflow>
```

This works but requires every consumer to duplicate the same capping logic and manually merge pre-overflowed items into templates.

## Proposed Solution

Add a `MaxRenderedItems` parameter (or similar) to `FluentOverflow`:

```csharp
/// <summary>
/// Maximum number of items to render as DOM elements for overflow measurement.
/// Items beyond this limit are treated as pre-overflowed without being added to the DOM.
/// When null (default), all items are rendered (current behavior).
/// </summary>
[Parameter]
public int? MaxRenderedItems { get; set; }
```

When set, the component would:
1. Only render the first `MaxRenderedItems` items as `FluentOverflowItem` DOM elements
2. Automatically include the remaining items in `ItemsOverflow` (and thus in `MoreButtonTemplate` / `OverflowTemplate` counts and collections)
3. Preserve the existing `Data` property on overflow items so consumers can still access the underlying data in templates

This keeps backward compatibility (default is `null` = render all, same as today) while giving consumers a simple escape hatch for large collections.

## Environment

- **Package**: Microsoft.FluentUI.AspNetCore.Components 4.14.1
- **Scenario**: Blazor Server with SignalR (most sensitive to long UI thread blocks)
- **Scale**: 100-500+ items in a single `FluentOverflow`

vnbaaij:
# Pull Request

## 📖 Description

This refactor moves `FluentOverflow` to a web-component-first implementation (`<fluent-overflow>`) to reduce resize latency and remove the old back-and-forth Blazor/JS flow. It also aligns AppBar overflow behavior with the same model and refreshes Overflow docs/examples to match the new API and migration path.

The Overflow component now wraps a dedicated fluent-overflow custom element, replacing the previous Blazor-heavy back-and-forth flow with direct overflow events and on-demand state retrieval. This reduces resize latency and makes overflow behavior more responsive.

## What changed

- Migrated overflow logic to a dedicated web component implementation in scripts, with startup/event wiring.
- Reworked FluentOverflow to use direct children as managed items, removing the old item component pattern.
- Added and integrated custom overflow event payloads (including overflow count and first overflow index).
- Updated AppBar overflow handling to the same web-component path and boundary-based overflow state updates.
- Added support for capping returned overflow payload items (MaxRenderedItems / max-rendered-items) while keeping total overflow count available.
- Improved overflow calculation behavior around:
  - fixed items and ellipsis items,
  - selector-based managed vs unmanaged children,
  - custom “more” content sizing,
  - narrow-width edge cases and badge visibility behavior.
- Refined keyboard behavior and styling interactions in AppBar and Overflow where needed.

## API and migration surface

- Renamed MoreButtonTemplate to MoreTemplate.
- Added/clarified fixed item modes (fixed="fixed" and fixed="ellipsis").
- Removed legacy Overflow item model and related enum usage in favor of direct child attributes.
- Introduced a dedicated migration guide for FluentOverflow and linked it from docs.

## Docs and examples

Overflow docs were expanded and reorganized with focused examples:

- custom templates/dynamic items,
- selector-based overflow,
- MaxRenderedItems behavior,
- multiple fixed items,
- fixed mode documentation.

Examples were adjusted to better reflect real behavior and expected constraints at narrow sizes.

## Validation

- Updated relevant Overflow tests and naming.
- Existing Overflow/AppBar test suites pass.
- Demo/docs build succeeds with the updated examples and migration content.

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

Please focus review on:
- Overflow behavior with fixed items, selector-based management, and custom more-template sizing.
- AppBar overflow transitions and keyboard navigation changes.
- API rename from `MoreButtonTemplate` to `MoreTemplate`.

## 📑 Test Plan

- Ran Overflow/AppBar component tests.
- Built Demo Client and validated updated Overflow docs/examples.
- Verified behavior manually in demo pages for resize, selector mode, fixed modes, max-rendered-items, and custom template scenarios.

### General

### Component-specific

image


src/Core/Components/Overflow/FluentOverflow.razor.cs
```diff
@@ -51,34 +45,55 @@ internal FluentOverflow(LibraryConfiguration configuration, List<FluentOverflowI
...
+    /// </summary>
+    [Parameter]
+    public string? Selector { get; set; } = string.Empty;
+
     /// <summary>
     /// Gets or sets the CSS selectors of the items to include in the overflow.
+    /// Use <see cref="Selector"/> instead.
     /// </summary>
     [Parameter]
     public string? Selectors { get; set; } = string.Empty;
```
dvoituron: I don't understand why we have 2 "CSS selector" parameters. I think only one can be solve all cases.

src/Core/Components/Overflow/FluentOverflow.razor.cs
dvoituron: Can you update the `<c>` to respect [MA0154](https://github.com/meziantou/Meziantou.Analyzer/blob/main/docs/Rules/MA0154.md)

src/Core/Components/Overflow/FluentOverflow.razor
```diff
@@ -2,31 +2,38 @@
...
-            {
-                <FluentBadge Appearance="BadgeAppearance.Filled" Style="min-width: 32px; max-width:32px;" Content="@($"+{ItemsOverflow.Count()}")" />
-            }
-        </div>
+<fluent-overflow id="@Id"
+                 class="@ClassValue"
+                 style="@StyleValue"
+                 orientation="@(Orientation.ToAttributeValue())"
+                 selector="@ItemSelector"
+                 visible-on-load="@(VisibleOnLoad ? "true" : "false")"
```
dvoituron: To avoid rendering extra HTML code, I think you can set `? "true" : null`

src/Core/Components/Overflow/FluentOverflow.razor.cs
```diff
@@ -88,99 +103,141 @@ internal FluentOverflow(LibraryConfiguration configuration, List<FluentOverflowI
...
         }
 
         await InvokeAsync(StateHasChanged);
     }
-    internal async Task RegisterAsync(FluentOverflowItem item)
+
+    private async Task LoadOverflowItemsAsync()
     {
-        await InvokeAsync(() => _items.Add(item));
+        var items = await JSRuntime.InvokeAsync<OverflowItem[]>("Microsoft.FluentUI.Blazor.Components.Overflow.GetOverflowItems", [Id]);
```
dvoituron: To avoid 2 JS calls, the `GetOverflowItems` result can be a object
```
{
   Items: <GetOverflowItems>,
   Count: <GetOverflowCount>
}
```

src/Core/Components/Overflow/FluentOverflow.razor.cs
```diff
@@ -88,99 +103,141 @@ internal FluentOverflow(LibraryConfiguration configuration, List<FluentOverflowI
...
+                Index = item.Index
+            }));
     }
 
     /// <summary>
-    /// Represents an item that may be subject to overflow handling, typically used in scenarios where content or data
-    /// exceeds a predefined limit.
+    /// Represents an item that may be subject to overflow handling.
     /// </summary>
     public class OverflowItem
```
dvoituron: This class must be included in a separate file named `OverflowItem.cs`, and the namespace will be located in`Microsoft.FluentUI.AspNetCore.Components`

src/Core/Components/Overflow/FluentOverflow.razor.cs
```diff
@@ -88,99 +103,141 @@ internal FluentOverflow(LibraryConfiguration configuration, List<FluentOverflowI
...
+
+        _items.AddRange(items
+            .Where(item => item.Overflow)
+            .Select(item => new OverflowItem
+            {
+                Id = item.Id,
+                Overflow = item.Overflow,
+                Text = item.Text,
+                Fixed = item.Fixed,
+                Index = item.Index
```
dvoituron: Add a coma after the last value

src/Core/Components/Overflow/FluentOverflow.razor.cs
```diff
@@ -88,99 +103,141 @@ internal FluentOverflow(LibraryConfiguration configuration, List<FluentOverflowI
...
     }
 
     /// <summary>
-    /// Represents an item that may be subject to overflow handling, typically used in scenarios where content or data
-    /// exceeds a predefined limit.
+    /// Represents an item that may be subject to overflow handling.
     /// </summary>
     public class OverflowItem
     {
         /// <summary />
```
dvoituron: Can you add comment to explain all properties?

src/Core/Components/Overflow/FluentOverflow.razor.cs
```diff
@@ -88,99 +103,141 @@ internal FluentOverflow(LibraryConfiguration configuration, List<FluentOverflowI
...
         public string? Id { get; set; }
```
dvoituron: All of these properties must be read-only outside the component. 
Can we use `{ get; init; }`, for example?

src/Core/Components/AppBar/FluentAppBar.razor.css
```diff
@@ -18,6 +18,7 @@
     height: calc(var(--appbar-item-size) - var(--appbar-item-height-adjustment));
   }
 
+
```
dvoituron, line 21: There are no change in this CSS file

src/Core/Events/OverflowChangedItem.cs
```diff
@@ -0,0 +1,36 @@
...
+
+/// <summary>
+/// Represents one overflow item state passed by the overflowchange custom event.
+/// </summary>
+public class OverflowChangedItem
+{
+    /// <summary>
+    /// Gets or sets the item identifier.
+    /// </summary>
+    public string? Id { get; set; }
```
dvoituron, line 15: All of these properties must be read-only outside the component. 
Can we use `{ get; init; }` or `{ get; internal set; }`, for example?

src/Core/Events/OverflowChangedItem.cs
```diff
@@ -0,0 +1,36 @@
...
+
+    /// <summary>
+    /// Gets or sets the item text.
+    /// </summary>
+    public string? Text { get; set; }
+
+    /// <summary>
+    /// Gets or sets the fixed behavior.
+    /// </summary>
+    public string? Fixed { get; set; }
```
dvoituron: This property must be an enumeration containing “Fixed” and “Ellipsis”

src/Core.Scripts/src/Components/Overflow/FluentOverflow.ts
```diff
@@ -0,0 +1,552 @@
+import { StartedMode } from "../../d-ts/StartedMode";
+
+export namespace Microsoft.FluentUI.Blazor.Components.Overflow {
```
dvoituron, line 16: Could you ask Copilot to explain the code by adding some comments? :-)

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Overflow/FluentOverflow.md
```diff
@@ -21,18 +22,50 @@ will be added when the container size is too small.
...
+This example sets `MaxRenderedItems="2"` and shows the difference between `OverflowCount` (total) and `ItemsOverflow` (rendered subset).
+{{ OverflowMaxRenderedItemsExample }}
+
+## Fixed item modes
+
+The `fixed` attribute on direct children supports these modes:
+
+| Value | Behavior |
+| --- | --- |
+| `fixed="fixed"` | Item always remains visible at full size and does not move to overflow. |
```
dvoituron: I don't like the syntax `fixed=fixed` :-)

Copilot suggests the following:
- `<div pinned="static">...</div>` or `<div pinned>...</div>`
- `<div pinned="ellipsis">...</div>`

dvoituron (changes requested):

src/Core/Components/Overflow/FluentOverflow.razor.cs
vnbaaij: Yup. That was a refactoring leftover. Only `Selector` remains. Will update PR description

src/Core/Components/Overflow/FluentOverflow.razor.cs
vnbaaij: Refactored so there is only one `OverflowState` that returns everything

src/Core/Components/Overflow/FluentOverflow.razor.cs
vnbaaij: Moved to a separate file and converted to a record type as the properties are immutable (same for new `OverflowState`

src/Core/Components/Overflow/FluentOverflow.razor.cs
vnbaaij: Done. Converted to record because properties are immutable

src/Core/Components/Overflow/FluentOverflow.razor.cs
vnbaaij: It is a record now. Records are the better choice since these are data transfer objects from JSInterop.

src/Core/Components/AppBar/FluentAppBar.razor.css
vnbaaij: Undid the extra newline at the end

src/Core/Events/OverflowChangedItem.cs
vnbaaij: Done

examples/Demo/FluentUI.Demo.Client/Documentation/Components/Overflow/FluentOverflow.md
vnbaaij: I renamed the attribute to `behavior`. Enum was already named `OverflowBehavior`  (by me)

src/Core.Scripts/src/Components/Overflow/FluentOverflow.ts
vnbaaij: Done

src/Core/Components/Overflow/OverflowState.cs
```diff
@@ -0,0 +1,32 @@
+// ------------------------------------------------------------------------
+// This file is licensed to you under the MIT License.
+// ------------------------------------------------------------------------
+
+namespace Microsoft.FluentUI.AspNetCore.Components;
+
+/// <summary>
+/// Represents the current overflow state.
+/// </summary>
+public record OverflowState
```
dvoituron: This class can safely be `internal`.

dvoituron (changes requested):

dvoituron (approved):

PR #4971, opened 2026-06-26
[dev-v5] Add styles as adopted stylesheets 

### LINKED ISSUE #4834: [dev-v5] fix: theme initialization in SSR mode
# 🐛 Bug Report

The theme will "unload" in SSR mode when you navigate to another page.

## 💻 Repro or Code Sample

https://github.com/microsoft/fluentui-blazor/pull/4831

1. Run the project
2. Click on weather

## 🤔 Expected Behavior

The theme should not unload

## 😯 Current Behavior

https://github.com/user-attachments/assets/377afa4f-c94d-4217-a207-28c31e3c2941

## 🌍 Your Environment

* Browser latest Firefox, Chrome & Edge
* net 10 and latest dev-v5 branch

vnbaaij:
Fix #4834

## Summary: Prevent styles from being removed in SSR 
Every time Blazor enhanced navigation happened, the external CSS `<link>` elements were being removed and re-added to the DOM. This forced the browser to:
1.	Remove the stylesheets
2.	Re-fetch the CSS files
3.	Re-parse and apply the styles
4.	This caused a visible layout shift/"flash"

## The New Solution
Instead of using DOM `<link>` elements, the CSS is now loaded via `fetch()` and added to `document.adoptedStyleSheets` as `CSSStyleSheet` objects. This approach:
Benefits:
•	✅ No viewport flash - Adopted stylesheets persist and don't require DOM manipulation
•	✅ Styles are preserved - CSS files are cached in memory as CSSStyleSheet objects after initial load
•	✅ No unnecessary refetches - After the first load, subsequent navigations just re-add the sheet references to adoptedStyleSheets
•	✅ No layout recalculations - Since it's not DOM manipulation, there are no layout churn

## How Navigation Works Now
1.	User navigates via nav menu → enhancedload event fires
2.	reapplyStyles() is called from Startup
3.	It checks if adopted stylesheets are in document.adoptedStyleSheets
4.	If missing (which they are after Blazor SSR navigation), it re-adds them
5.	No DOM changes, no CSS refetch, no visual flash ✨

The solution is elegant because it leverages the fact that JavaScript `CSSStyleSheet` objects persist in memory - we just need to re-add their references to `adoptedStyleSheets` on navigation!

### DIFF
--- a/src/Core.Scripts/src/Startup.ts
+++ b/src/Core.Scripts/src/Startup.ts
@@ -74,10 +74,8 @@ export namespace Microsoft.FluentUI.Blazor.Startup {
     FluentUICustomEvents.TextInput(blazor);
     // [^^^ Add your other custom events before this line ^^^]
 
-    // Register enhancedload listener to apply styles on each navigation
     if (blazor.addEventListener && typeof blazor.addEventListener === 'function') {
       blazor.addEventListener('enhancedload', () => {
-        Logger.debug('enhancedload event triggered, reapplying FluentUI styles');
         FluentUIStyles.reapplyStyles();
       });
     }
--- a/src/Core.Scripts/src/Startup.ts
+++ b/src/Core.Scripts/src/Startup.ts
@@ -74,6 +74,14 @@ export namespace Microsoft.FluentUI.Blazor.Startup {
     FluentUICustomEvents.TextInput(blazor);
     // [^^^ Add your other custom events before this line ^^^]
 
+    // Register enhancedload listener to apply styles on each navigation
+    if (blazor.addEventListener && typeof blazor.addEventListener === 'function') {
+      blazor.addEventListener('enhancedload', () => {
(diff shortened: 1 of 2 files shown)


src/Core.Scripts/src/Startup.ts
```diff
@@ -74,6 +74,14 @@ export namespace Microsoft.FluentUI.Blazor.Startup {
     FluentUICustomEvents.TextInput(blazor);
     // [^^^ Add your other custom events before this line ^^^]
 
+    // Register enhancedload listener to apply styles on each navigation
+    if (blazor.addEventListener && typeof blazor.addEventListener === 'function') {
+      blazor.addEventListener('enhancedload', () => {
+        Logger.debug('enhancedload event triggered, reapplying FluentUI styles');
```
dvoituron: This log must be removed. No interest for the devs.

dvoituron (changes requested):

dvoituron (approved):

Issue #4701, opened 2026-04-14
[dev-v5] Refactor the DataGrid class name


dvoituron:
# [dev-v5] Refactor the DataGrid class name

In a component library, two dominant CSS strategies emerge: class-name prefixing (e.g., `fluent-grid`, `fluent-grid-spacing-3`, `fluent-grid-item-xs-6`) following methodologies like BEM, and root class with attribute selectors (e.g., `.fluent-grid[spacing="3"] > div[xs="6"]`). The prefix approach, popularized by BEM, ensures globally unique, self-documenting class names but quickly leads to verbose markup and an explosion of class combinations. The attribute-selector approach anchors styles to a single, namespaced root class and leverages HTML attributes that directly mirror component parameters. This produces leaner, more semantic markup: the attribute xs="6" on a grid item simultaneously serves as the styling hook and the declarative API surface. 

I think the best approach is to use a root `fluent-data-grid` class and then use attributes to be targeted by CSS selectors: this avoids overly verbose HTML and prevents the use of generic CSS classes.

Example of "generic" class to avoid.
```css
.column-header { background: red; }
.grid { background: blue; }
.col-justify-start { background: yellow; }
```

PR #4983, opened 2026-06-30
[dev-v5][DataGrid] Fix FluentDataGrid not working correctly when used in Drawer/Dialog

### LINKED ISSUE #4562: [dev-v5] fix: keyboard support for `FluentDataGrid` when being using in Drawer/Dialog
# 🐛 Bug Report

Keyboard support for `FluentDataGrid` is broken when it's being used within a Drawer/Dialog.

## 💻 Repro or Code Sample

Copy this code into `SimpleDialog.razor` in the demos:

```razor
@inherits FluentDialogInstance

<FluentDialogBody>
    <FluentTextInput Label="Name:" @bind-Value="@Name" Disabled="true" Style="max-width: 100%;" />
    <FluentTextInput Label="Age:" @bind-Value="@Person.Age" Style="max-width: 100%;" />

    <FluentPaginator State="@pagination" SummaryTemplate="@template" />

    <FluentDataGrid Items="@people" Pagination="@pagination">
        <PropertyColumn Property="@(p => p.PersonId)" Sortable="true" />
        <PropertyColumn Property="@(p => p.Name)" Sortable="true" />
        <PropertyColumn Property="@(p => p.BirthDate)" Format="yyyy-MM-dd" Sortable="true" />
    </FluentDataGrid>

    <FluentPaginator State="@pagination" />

</FluentDialogBody>
@code {
    PaginationState pagination = new PaginationState() { ItemsPerPage = 2 };

    record Person2(int PersonId, string Name, DateOnly BirthDate);

    IQueryable<Person2> people = new[]
    {
        new Person2(10895, "Jean Martin", new DateOnly(1985, 3, 16)),
        new Person2(10944, "António Langa", new DateOnly(1991, 12, 1)),
        new Person2(11203, "Julie Smith", new DateOnly(1958, 10, 10)),
        new Person2(11205, "Nur Sari", new DateOnly(1922, 4, 27)),
        new Person2(11898, "Jose Hernandez", new DateOnly(2011, 5, 3)),
        new Person2(12130, "Kenji Sato", new DateOnly(2004, 1, 9)),
    }.AsQueryable();

    private RenderFragment template = @<span />;
}
@code {
    // A simple type is not updatable
    [Parameter]

    // A class is updatable
    [Parameter]
    public PersonDetails Person { get; set; } = new();

    // Initialize the dialog
    protected override void OnInitializeDialog(DialogOptionsHeader header, DialogOptionsFooter footer)
    {
        header.Title = $"Dialog Title - {Name}";
        footer.SecondaryAction.Visible = true;
    }

    // Handle the action click
    protected override async Task OnActionClickedAsync(bool primary)
    {
        if (primary)
        {
            await DialogInstance.CloseAsync("Yes");
        }
        else
        {
            await DialogInstance.CancelAsync();
        }
    }
}

```

Then open the Drawer demo and open the drawer. Try to tab through the `FluentDataGrid`in the drawer or try to select text and copy via `CTRL+C`

## 🤔 Expected Behavior

I expect the keyboard events to work properly just like when the `FluentDataGrid` is being used on a normal page.

## 😯 Current Behavior

Almost every keyboard feature is not working anymore. You can Tab through the first two headers though.

## 🌍 Your Environment

* Browser Latest Firefox and Edge
* .NET 10 and latest dev-v5 branch

vnbaaij:
Fixes #4562

## Problem

When a `FluentDataGrid` with header menus is hosted inside a `FluentDialog`/drawer, opening a
column-header menu caused it to flash open and close instantly. Both mouse and keyboard
interaction were affected, leaving the menu unusable.

## Root cause

`FluentDialog` used the default `IHandleEvent` behavior, which calls `StateHasChanged`
automatically after every event callback. The `beforetoggle`/`toggle` and `keydown` events
that bubble up from popover content inside the dialog (such as the DataGrid header menu) were
attributed to the dialog's own handlers via Blazor's event delegation. The resulting automatic
re-render of the entire dialog subtree recreated the grid's keyed header cells, which detached
the just-opened menu popover and closed it immediately.

## Fix

`FluentDialog` now implements `IHandleEvent` to suppress the unnecessary automatic re-render
after its event callbacks, and ignores toggle events that don't target the dialog instance. This
keeps the dialog content stable when inner popovers open, so the menu stays open and is fully
operable by mouse and keyboard.


src/Core/Components/DataGrid/FluentDataGrid.razor.ts
```diff
@@ -163,7 +329,9 @@ export namespace Microsoft.FluentUI.Blazor.DataGrid {
     document.body.addEventListener('click', bodyClickHandler, { signal });
     document.body.addEventListener('mousedown', bodyClickHandler, { signal });
     document.body.addEventListener('keydown', bodyKeyDownHandler, { signal });
-    gridElement.addEventListener('keydown', keyDownHandler, { signal });
+    // Listen at document capture phase so keyboard navigation still works when focus is
+    // inside inner elements that stop propagation before table-level handlers run.
+    document.addEventListener('keydown', keyDownHandler, { signal, capture: true });
```
dvoituron: As I’ve already said, it’s always **risky to use global events**. I can’t see where they’re removed. 
What’s more, this will affect performance: every click or keydown will trigger these events

Example

https://github.com/user-attachments/assets/8f346e6b-84b1-4a89-8f89-ffe991c890e1

What would happen if the developer added another library that handles arrays `<Table>`, `<tr>`, …? 
I get the impression that this code would also be applied

src/Core.Scripts/src/Components/Dialog/FluentDialog.ts
```diff
@@ -90,4 +99,33 @@ export namespace Microsoft.FluentUI.Blazor.Components.Dialog {
...
+
+    const activeElement = getDeepActiveElement();
+    if (!activeElement || !dialog.contains(activeElement)) {
+      return false;
+    }
+
+    // Never steal keyboard handling from editable/menu-like controls in dialog content.
+    if (
+      activeElement.matches('input, textarea, select, fluent-data-grid, [contenteditable=""], [contenteditable="true"], [role="textbox"], [role="combobox"], [role="spinbutton"], [role="listbox"], [role="menu"], [role="menuitem"]') ||
+      !!activeElement.closest('fluent-menu, fluent-menu-list, fluent-menu-item, .col-header-ui')
```
dvoituron: I think it’s not safe to specify tags to be ignored. For example, this issue #4701 would affect this code. Furthermore, if the user adds other components, they won’t be able to manage this functionality. 

We need to find another standard solution or allow configuration via LibraryConfiguration.
Or, more simply, via a static, global JavaScript property, which can be modified by the developer if necessary

dvoituron (changes requested):

src/Core/Components/Dialog/FluentDialog.razor.cs
```diff
@@ -185,6 +214,15 @@ private async Task OnKeyDownHandlerAsync(Microsoft.AspNetCore.Components.Web.Key
             return;
         }
 
+        if (IsDrawer())
```
dvoituron: Why does this apply only to “Drawer” and not to “Dialog”?

dvoituron (changes requested):

src/Core/Components/Dialog/FluentDialog.razor.cs
```diff
@@ -185,6 +214,15 @@ private async Task OnKeyDownHandlerAsync(Microsoft.AspNetCore.Components.Web.Key
...
+        {
+            var shouldHandleShortcut = await JSRuntime.InvokeAsync<bool>("Microsoft.FluentUI.Blazor.Components.Dialog.ShouldHandleShortcut", Id);
+            if (!shouldHandleShortcut)
```
dvoituron: Wouldn't it be simpler to ask the developer to modify or remove the `PrimaryAction.Shortcut` or `SecondaryAction.Shortcut` if they don't want to use it?

Or we could provide a list of element IDs or a `QuerySelectorAll[]` to ignore the shortcuts when the focus is within those `HtmlElement`s. Example: `DialogOptionsFooter.ExcludeActionShortcuts`

dvoituron (changes requested):

src/Core/Components/DataGrid/FluentDataGrid.razor.ts
vnbaaij: Clear, I'll work on that

src/Core/Components/Dialog/FluentDialog.razor.cs
vnbaaij: It depends on where the focus is at. If it is in the DataGrid, then pressing esc should close a DataGrid column option menu but not close the dialog. Once focus is out of the DataGrid, the ESC kay can resume it's normal function. That is what the `ShouldHandleShortcut` does

src/Core/Components/Dialog/FluentDialog.razor.cs
vnbaaij: Need to check that

src/Core/Components/Dialog/FluentDialog.razor.cs
dvoituron: Here’s what I’m proposing: a `DialogOptionsFooter.ExcludeActionShortcuts` property that allows you to define which elements to exclude from the dialogue’s shortcuts. By default, it would contain FluentUI components such as the DataGrid or the FluentMenu, but this would allow developers to add other elements they have created. 
For example, if they add their own DataGrid, the actual PR code won’t work

In summary, `ExcludeActionShortcuts` provides a list of `HtmlElements`. If the focus is on this element (or a child element), then the shortcut is disabled

src/Core/Components/DataGrid/FluentDataGrid.razor.ts
dvoituron: I think you need to add events to the DataGrid and unsubscribe when the component is removed. The advantage is that even if there's a problem with the unsubscription process (network error, JavaScript error, etc.), once the DataGrid is removed from the DOM, all associated objects and events are also removed.

vnbaaij: It is a listener on the grid element now. It gets removed with stop()

src/Core.Scripts/src/Components/Dialog/FluentDialog.ts
vnbaaij: the whole matches/closest check is not needed anymore now that we have the correct focus checks in place

src/Core/Components/Dialog/FluentDialog.razor.cs
vnbaaij: Restriction is removed (after I tested behavior in normal Dialog)

dvoituron (approved):

PR #4989, opened 2026-07-02
[dev-v5] Make browser native validation opt-in

vnbaaij:
When using the `FluentValidationMessage` (introduced with RC4) we now also hook into the browser native validation (the Constrained Validation API). This means that when an input looses focus (_not on submit!_), it is checked against the set constraints and shows browser native UI (see image below)

With this PR we make that behavior opt-in. 

You opt-in by setting `UseNativeConstraintValidationUI=true` on the `LibraryConfiguration` when calling the `AddFluentUIComponents()` extension method in `Program.cs`:

```
// Add FluentUI services
builder.Services.AddFluentUIComponents(config  => 
        config.UseNativeConstraintValidationUI = true);
```

### Before
image

### After
image

Note that you still get the red border on the input. That is part of the web component where it is applying the `:user-invalid` state 

Unit tests to validate whether the `reportValidity` function is called, have been added.

### DIFF
--- a/src/Core/Components/Base/FluentInputBase.cs
+++ b/src/Core/Components/Base/FluentInputBase.cs
@@ -37,7 +37,7 @@ public abstract partial class FluentInputBase<TValue> : InputBase<TValue>, IFlue
         // and can be overridden by setting the component parameter in markup.
         if (configuration is not null)
         {
-            UseNativeConstraintValidationUi = configuration.UseNativeConstraintValidationUi;
+            UseNativeConstraintValidationUI = configuration.UseNativeConstraintValidationUI;
         }
     }
 
@@ -218,7 +218,7 @@ public abstract partial class FluentInputBase<TValue> : InputBase<TValue>, IFlue
     /// Gets or sets whether the control will use the native browser constraint validation UI.
     /// </summary>
     [Parameter]
-    public bool UseNativeConstraintValidationUi { get; set; }
+    public bool UseNativeConstraintValidationUI { get; set; }
 
     /// <summary />
     [System.Diagnostics.CodeAnalysis.SuppressMessage("Style", "IDE0059:Unnecessary assignment of a value", Justification = "TODO")]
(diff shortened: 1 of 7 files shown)


src/Core/Components/Base/FluentInputBase.cs
```diff
@@ -31,6 +31,14 @@ protected FluentInputBase(LibraryConfiguration configuration)
     {
         ValueExpression = () => CurrentValueOrDefault;
         configuration?.DefaultValues.ApplyDefaults(this);
+
+        // Apply the library configured default for using the native browser
+        // constraint validation UI. This value acts as the component default
+        // and can be overridden by setting the component parameter in markup.
+        if (configuration is not null)
+        {
+            UseNativeConstraintValidationUi = configuration.UseNativeConstraintValidationUi;
```
dvoituron: Typo: "UI"

tests/Core/Components/Validation/FluentInputValidationDefaultsTests.cs
```diff
@@ -0,0 +1,62 @@
...
+    [Fact]
+    public void Default_NoOptIn_ReportValidityNotCalled()
+    {
+        using var ctx = new Bunit.BunitContext();
+        ctx.JSInterop.Mode = JSRuntimeMode.Loose;
+        ctx.Services.AddSingleton<IHtmlParser>(new HtmlParser());
+        ctx.Services.AddFluentUIComponents();
+
+        // Arrange
+#pragma warning disable CS0619 // RenderComponent is obsolete in current bUnit, but safe for tests
```
dvoituron: Why not to use a razor file to avoid this pragma and to simply the code?

dvoituron (changes requested):

tests/Core/Components/Validation/FluentInputValidationDefaultsTests.cs
vnbaaij: It is not a razor file because we are not testing a specific component. We ar testing specific functionality only.

The pragma was no longer needed and is removed

dvoituron (approved):

PR #4994, opened 2026-07-03
[dev-v5] MessageBar - Add ResultTiming

### LINKED ISSUE #4992: [dev-v5] Refactor MessageBar helper methods to be non blocking
# 🙋 Feature Request

In #4965 we have refactored the helper methods for showing toast notifications to be non blocking by default. We should align this change for the messagebar helper methods as well. This will improve the experience for the developers because both helpers work in the exact same way. 

In case the developer actually wants to block the code execution he can always fallback to the general method of the service.

dvoituron:
# [dev-v5] MessageBar - Add ResultTiming

Introduce a new `ResultTiming` option for the `MessageBar` to control when the result task is completed, either when the bar is visible or closed. 
This allows for greater flexibility in using the methods, but also ensures consistency with how Toasts work

Now, these four helper methods are non-blocking. Because they use `MessageBarResultTiming.Visible`, the awaited call completes 
as soon as the bar is displayed; thus, the code following the `await` statement continues to execute immediately, without waiting for the bar to be closed: `ShowSuccessBarAsync`, `ShowWarningBarAsync`, `ShowErrorBarAsync`, `ShowInfoBarAsync`

Update documentation. 

This change improves flexibility in handling message bar results.

Fix #4992

## Unit Tests

Added

### DIFF
--- a/examples/Demo/FluentUI.Demo.Client/Documentation/Components/MessageBar/Examples/MessageBarServiceCustomized.razor
+++ b/examples/Demo/FluentUI.Demo.Client/Documentation/Components/MessageBar/Examples/MessageBarServiceCustomized.razor
@@ -27,6 +27,7 @@
             options.Intent = MessageBarIntent.Success;
             options.Title = "Hello from the NotificationService!";
             options.Lifetime = TimeSpan.FromSeconds(10);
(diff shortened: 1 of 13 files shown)


MarvinKlein1508:
LGTM

vnbaaij (approved):

Issue #4929, opened 2026-06-15
fix: [dev-5] DataGrid rendering performance is slow


Bartsz:
# 🐛 Bug Report
DataGrid takes a lot of time to apply changes after sorting or grouping (reordering columns also takes a lot of time).
For 200 rows and ~8 columns it takes ~500ms from click to show changes
 

 
## 💻 Repro or Code Sample

Steps to reproduce:
1. Open weather page
2. Sort by any column.

```C#
@page "/weather"
@using System.Diagnostics
@inject ILogger<Weather> Logger

<PageTitle>Weather</PageTitle>
 
<FluentDataGrid Id="weathergrid"
                Items="@forecasts"
                OnSortChanged="@HandleOnSortChanged" 
                GridTemplateColumns="1fr 1fr 1fr 2fr" TGridItem=WeatherForecast>
    <PropertyColumn Title="Date" Property="@(c => c!.Date)" Sortable="true" Align="DataGridCellAlignment.Start" />
    <PropertyColumn Title="Temp. (C)" Property="@(c => c!.TemperatureC)" Sortable="true" Align="DataGridCellAlignment.Center" />
    <PropertyColumn Title="Temp. (F)" Property="@(c => c!.TemperatureF)" Sortable="true" Align="DataGridCellAlignment.Center" />
    <PropertyColumn Title="Summary" Property="@(c => c!.Summary)" Sortable="true" Align="DataGridCellAlignment.End" />
</FluentDataGrid>

@code {
    private IQueryable<WeatherForecast>? forecasts;
    private Stopwatch _eventToRenderWatch = new();
    private int _renderCount;

    private void HandleOnSortChanged()
    {
        _eventToRenderWatch.Restart();
    }
    protected override void OnInitialized()
    {
        base.OnInitialized();

        forecasts = Enumerable.Range(1, 505).Select(index => new WeatherForecast
        {
            Date = DateOnly.FromDateTime(DateTime.Now.AddDays(index)),
            TemperatureC = Random.Shared.Next(-20, 55),
            Summary = (index % 5) switch
            {
                0 => "Freezing",
                1 => "Bracing",
                2 => "Chilly",
                3 => "Cool",
                4 => "Mild",
            }
        })
         .ToList()
         .AsQueryable();
    }

    protected override void OnAfterRender(bool firstRender)
    {
        _renderCount++;

        if (_eventToRenderWatch.IsRunning)
        {
            _eventToRenderWatch.Stop();
[...]

PR #5023, opened 2026-07-16
[dev-v5][DataGrid] Rendering perf improvement - non-interactive cells as plain <td>

gkizior:
# Pull Request

## 📖 Description

`FluentDataGrid` renders every cell as a `FluentDataGridCell` component. For a grid with no cell-level interaction, that component doesn't produce anything beyond a styled `<td>`, but it still pays the full per-cell cost: instantiation, parameter binding, cascading values, and rebuilding the class/style on every render.

So for those cells I render a plain `<td>` instead (which is what the QuickGrid base already does), and keep `FluentDataGridCell` around for the grids that actually need it. This also gets at the question raised in #4929 about whether the cost is QuickGrid's design or something we added: the expensive part is the per-cell component, and a read-only grid can skip it.

How it works:
- `ColumnBase.RequiresCellComponent` (new `protected internal virtual`, defaults to `false`) lets a column say its cells need the component. `SelectColumn` overrides it to `true`.
- The grid only renders `FluentDataGridCell` components when a column needs them, the grid has `OnCellClick`/`OnCellFocus` handlers, or a column is hierarchical. Otherwise it emits plain `<td>` cells. It's an all-or-nothing decision per grid so the cells stay uniform.
- Both paths pull their class/style from the same shared helpers on `FluentDataGridCell`, so the markup that comes out is identical.
- `FluentDataGridRow` now marks its `OwningRow` cascade `IsFixed="true"`. A cell never moves to a different row over its lifetime, so the reference doesn't change, and telling Blazor that lets it skip subscribing every cell to the cascade for change notifications. That's a real chunk of the per-cell overhead on grids with a lot of cells.
- A few smaller changes came along with it: `StyleBuilder.AddStyle(prop, value, bool)` no longer allocates a closure (now matches how `CssBuilder` does it), cells only wire up click/focus/keydown when something actually handles them, and the cell title/aria-label is passed as a parameter rather than splatted attributes.

This isn't a breaking change. The rendered output is the same, and the only additions to the public surface are one parameter (`FluentDataGridCell.CellTitle`) and one virtual (`ColumnBase.RequiresCellComponent`).

Numbers from a static server render of a 500-row × 8-column read-only grid (via `HtmlRenderer`, so allocations are deterministic):

| | Allocations / render |
|---|---|
| Before | ~37 MB |
| After | ~13 MB |
| QuickGrid (same data) | ~1.3 MB |

Roughly 65% fewer allocations, and render time drops accordingly.

### 🎫 Issues

Related to #4929. It covers the per-cell rendering cost discussed there; the cost of the sort/data pipeline itself is a separate thing and not part of this PR.

## 👩‍💻 Reviewer Notes

Plain `<td>` is only used when the `FluentDataGridCell` component wouldn't add anything observable. The decision lives in `FluentDataGrid.UseComponentCells()`/`CellNeedsComponent()`, and both paths build their class/style from the shared `FluentDataGridCell.BuildClass`/`BuildStyle`, so the markup is identical either way.

The existing tests cover the behavior, but if it's easier to eyeball: a read-only grid still sorts and paginates the same, a `SelectColumn` grid still selects, `OnCellClick`/`OnCellFocus` still fire, and keyboard navigation still works.

## 📑 Test Plan

- The full `Components.Tests` suite passes, including `SelectColumn`, hierarchical select, keyboard navigation, and the DataGrid snapshot/rendering tests (those confirm the plain `<td>` markup matches what the component produced before).
- Manually checked pagination, sorting, and rich (templated) cell content in a sample app on both the plain-cell and component-cell paths.
- Perf measured with the static-render benchmark above.

### General

### Component-specific

## ⏭ Next Steps

- The same idea could be applied to rows (plain `<tr>`), but that removes the `FluentDataGridRow` component that tests and consumers reach through `FindComponent<FluentDataGridRow>`, so I left it out here.
- The sort operation itself (re-materializing the data) is a separate cost from cell rendering and would be worth a follow-up look.

### DIFF
--- a/src/Core/Components/DataGrid/FluentDataGrid.razor
+++ b/src/Core/Components/DataGrid/FluentDataGrid.razor
@@ -116,46 +116,52 @@
 
                 string? tooltip = col.Tooltip ? @col.RawCellContent(item) : null;
 
-                @if (_useComponentCells)
+                @if (CellNeedsComponent(col))
                 {
                     <FluentDataGridCell GridColumn=@(colIndex + 1) Class="@ColumnJustifyClass(col)" Style="@col.Style" @key="@col" TGridItem="TGridItem" Item="@item" CellTitle="@tooltip">
-                        @if (col.HierarchicalToggle && item is IHierarchicalGridItem hierarchicalItem)
-                        {
-                            <span class="hierarchical-toggle-container" style="margin-inline-start: calc(var(--spacingHorizontalL) * @hierarchicalItem.Depth);">
-                                <span @onclick:stopPropagation="true" class="hierarchical-expander @(hierarchicalItem.IsCollapsed ? "hierarchical-collapsed" : "hierarchical-expanded")">
-                                    @if (hierarchicalItem.HasChildren)
-                                    {
-                                        <FluentButton aria-label="@Localizer[Localization.LanguageResource.DataGrid_ToggleNesting]" Appearance="ButtonAppearance.Subtle" Size="ButtonSize.Small" Class="hierarchical-expand-button" OnClick="@(() => ToggleExpandedAsync(item))">
-                                            <FluentIcon Width="12" Value="@(new CoreIcons.Regular.Size20.ChevronRight())" Color="Color.Default" />
-                                        </FluentButton>
-                                    }
-                                </span>
-                                @((RenderFragment)(__builder => col.CellContent(__builder, item)))
-                            </span>
-                        }
-                        else
-                        {
-                            @((RenderFragment)(__builder => col.CellContent(__builder, item)))
-                        }
+                        @((RenderFragment)(__builder => RenderCellInner(__builder, col, item)))
                     </FluentDataGridCell>
                 }
                 else
                 {
-                    // Fast path: simple, non-interactive column -> plain <td> instead of the
+                    // Columns that don't need cell-level interaction render as a plain <td> instead of the
                     // FluentDataGridCell component. Class/style come from the same shared builders the
                     // component uses, so the emitted markup is identical.
                     <td col-index="@(colIndex + 1)"
+                        @key="@col"
                         class="@PlainCellClass(col, rowClass)"
                         style="@PlainCellStyle(col, colIndex + 1, rowStyle)"
                         role="gridcell"
                         tabindex="@(col.DisableCellFocus ? null : "0")"
                         title="@tooltip"
                         aria-label="@tooltip">
-                        @((RenderFragment)(__builder => col.CellContent(__builder, item)))
+                        @((RenderFragment)(__builder => RenderCellInner(__builder, col, item)))
                     </td>
                 }
             }
         </FluentDataGridRow>
+    }
+
+    private void RenderCellInner(RenderTreeBuilder __builder, ColumnBase<TGridItem> col, TGridItem item)
+    {
+        @if (col.HierarchicalToggle && item is IHierarchicalGridItem hierarchicalItem)
+        {
+            <span class="hierarchical-toggle-container" style="margin-inline-start: calc(var(--spacingHorizontalL) * @hierarchicalItem.Depth);">
+                <span @onclick:stopPropagation="true" class="hierarchical-expander @(hierarchicalItem.IsCollapsed ? "hierarchical-collapsed" : "hierarchical-expanded")">
+                    @if (hierarchicalItem.HasChildren)
+                    {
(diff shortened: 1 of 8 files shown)


src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -1580,6 +1582,45 @@ private string AriaSortValue(ColumnBase<TGridItem> column)
...
+    private bool CellNeedsComponent(ColumnBase<TGridItem> column)
+        => column.RequiresCellComponent
+        || column.HierarchicalToggle
+        || OnCellClick.HasDelegate
+        || OnCellFocus.HasDelegate;
+
+    /// <summary>
+    /// Returns whether the grid renders its cells as <see cref="FluentDataGridCell{TGridItem}"/> components.
+    /// When no column needs the component, cells are rendered as plain <c>&lt;td&gt;</c> elements instead.
+    /// The decision is all-or-nothing so cells stay uniform within a grid.
```
vnbaaij: Why would we want/need the all-or-nothing approach? in the end we just get rendered td's. If we only use the `FluentDataGridCell` on columns that actually need it, we can still get a performance benefit by not using it for the others, right?

src/Core/Components/DataGrid/FluentDataGrid.razor
```diff
@@ -116,26 +116,44 @@
 
                 string? tooltip = col.Tooltip ? @col.RawCellContent(item) : null;
 
-                <FluentDataGridCell GridColumn=@(colIndex + 1) Class="@ColumnJustifyClass(col)" Style="@col.Style" @key="@col" TGridItem="TGridItem" Item="@item" title="@tooltip" aria-label="@tooltip" tabindex="@(col.DisableCellFocus ? null : 0)">
-                    @if (col.HierarchicalToggle && item is IHierarchicalGridItem hierarchicalItem)
```
vnbaaij, line 120: I think we can have this hierarchical check in the plain cell variant as well. The output from this block is put _inside_ the td and is if no consequence on something that Blazor needs to track on the td, right?

In other words, I don't think the `column.HierarchicalToggle` needs to be considered in the `UseComponentCells` method

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -1580,6 +1582,45 @@ private string AriaSortValue(ColumnBase<TGridItem> column)
...
+            }
+        }
+
+        return false;
+    }
+
+    /// <summary>
+    /// Builds the class for a plain <c>&lt;td&gt;</c> cell, using the same builder as <see cref="FluentDataGridCell{TGridItem}"/>.
+    /// </summary>
+    private string? PlainCellClass(ColumnBase<TGridItem> column, string? rowClass)
```
vnbaaij, line 1595: We still have #4701 on the todo list to work more with attributes instead of class names. We need to make sure that we can still do that even after we make the change in this function

vnbaaij (changes requested):

vnbaaij:
I like this! Made a few comments, but good stuff!

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
gkizior: resolved

src/Core/Components/DataGrid/FluentDataGrid.razor
gkizior: resolved

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
gkizior: @vnbaaij, I pushed a first pass at a seam in the [latest commit](https://github.com/microsoft/fluentui-blazor/pull/5023/changes/905af6f87738a8a1eb3db2e8c45d26c328b4388f), but I want to check the direction since I'm not sure how #4701 will land.

What I did: a new `ColumnBase.CellStyleAttributes` (virtual, null today) that both cell variants render, so the class -> attribute move would happen in one place. It's null/no-op right now, so no markup or allocation change.

That said, I'm wary it's premature since I don't know what #4701 will actually change. Is this roughly what you're after, or would you rather I drop it and leave the attribute work entirely to #4701?

vnbaaij: I have not looked at what your new code is doing, but I _do_ know that we don't want to do it in this PR... 

We want to (or at least try to 😉) keep our PRs focused on 1 thing and not make them too large. The styling with attributes can stay in the PR for addressing #4701. So please revert the things you did for that in this PR.

gkizior: ok, reverted.

vnbaaij (approved):

PR #5038, opened 2026-07-21
[dev-v5][Input] Make sure label is created at the right location

### LINKED ISSUE #4980: [dev-v5] Input components produce no accessible name — aria-labelledby/label target the roleless <fluent-*> host, leaving the shadow-DOM <input> unnamed
### Version

- `Microsoft.FluentUI.AspNetCore.Components` **5.0.0-rc.4-26180.1**
- .NET 10, Blazor Server (interactive)

### Is this a regression?

Yes — relative to v4 (see notes below).

### Describe the bug

In v5 the input components don't expose an **accessible name** on the actual focusable control. With a simple labelled input:

```razor
<FluentTextInput Label="Email" @bind-Value="_email" />
```

the rendered markup associates the label with the **host** custom element, not the real input:

```html
<label id="x-label" slot="label" for="x">Email</label>
<fluent-text-input id="x" aria-labelledby="x-label" ...>
  #shadow-root
    <input id="control" part="control" ...>   <!-- the real, focusable control -->
</fluent-text-input>
```

- `<label for="x">` and `aria-labelledby="x-label"` target `<fluent-text-input id="x">`, which has **no ARIA role**, so the accessible-name computation ignores them.
- The real `<input id="control">` lives in the shadow DOM with **no** `aria-label` / `aria-labelledby` and no associated `<label>`.

Result: the control's **computed accessible name is empty**, so screen readers announce the field with no name. This is a WCAG 4.1.2 / 1.3.1 failure.

### Evidence (computed accessibility tree, not just a linter)

Chromium's computed accessibility tree (via Playwright `AriaSnapshot`) for a labelled `FluentTextInput` / `FluentSelect`:

```
- text: Email
- textbox            ← no accessible name
- combobox           ← no accessible name (FluentSelect)
```

A correctly-named control would read `- textbox "Email"`.

axe-core reports, on the same page:
- `label` (**critical**) on the shadow `<input id="control">` — "Form element has explicit `<label>` that is hidden".
- `aria-prohibited-attr` (**serious**) on the host — "`aria-labelledby` cannot be used on a `fluent-text-input` with no valid role attribute".

**Confirmed via the AX tree:** `FluentTextInput` and `FluentSelect` (rendered as `fluent-dropdown`) produce a nameless control.
**Likely affected (same `aria-labelledby`-on-roleless-host markup; axe flags `aria-prohibited-attr` on each, but I did not individually inspect their computed name):** `FluentTextArea`, `FluentCheckbox`, `FluentSwitch`, `FluentRadio`.

### Workaround attempted

Setting `AriaLabel="Email"` does **not** fix it — the `aria-label` lands on the same roleless `<fluent-text-input>` host, the computed name is still empty, and axe then flags **both** `aria-label` and `aria-labelledby` as prohibited.

### Notes / prior art

- #3259 reported the v4 incarnation of this (label points at the host, not the input) and was closed *not planned*. However, on v4 the equivalent page was axe-clean for `label` in our suite, whereas v5 RC4 is not — so for our usage this is a **regression**.

### Steps to reproduce

1. Render `<FluentTextInput Label="Email" @bind-Value="x" />` in an interactive Blazor Server app on v5 RC4.
2. Inspect the computed accessibility tree (DevTools Accessibility pane, or Playwright `AriaSnapshot`).
3. Observe the textbox node has no accessible name; run axe-core and observe `label` (critical) + `aria-prohibited-attr` (serious).

### Expected behavior

The focusable control should carry the accessible name — e.g. the inner `<input>` should reference the label (`aria-labelledby` on the control, or a `<label for>` targeting the control's id), so screen readers announce the field name.

vnbaaij:
Fix #4980 

We need to re-evaluate how to render the label on input components. We are deviating from how it is done in the standard web components and that's causing a11y issues.

This is only a first stab at solving this. It is not complete and unit tests have not been created/altered.

We can use https://stackblitz.com/edit/typescript-juzjvyeg?file=index.html for seeing how it works on the WC side

**This must be released before go-live**

### DIFF
--- a/src/Core/Components/Field/FluentField.razor.cs
+++ b/src/Core/Components/Field/FluentField.razor.cs
@@ -223,9 +223,25 @@ public partial class FluentField : FluentComponentBase, IFluentField
     }
 
     private bool HasLabel
-        => !string.IsNullOrWhiteSpace(Parameters.Label)
-        || Parameters.LabelTemplate is not null
-        || Parameters.HasLabelInfo;
+    {
+        get
+        {
+            var hasLabelContent = !string.IsNullOrWhiteSpace(Parameters.Label)
+                || Parameters.LabelTemplate is not null
+                || Parameters.HasLabelInfo;
+
+            // Only show label if the InputComponent is FluentCheckbox or FluentSwitch
+            if (InputComponent is not null)
+            {
+                return IsCheckboxOrSwitch() && hasLabelContent;
+            }
+
+            return hasLabelContent;
+        }
+    }
+
+    private bool IsCheckboxOrSwitch()
+        => InputComponent?.GetType() is { Name: "FluentCheckbox" or "FluentSwitch" };
 
     private bool HasMessage
         => !string.IsNullOrWhiteSpace(Parameters.Message)
--- a/src/Core/Components/NumberInput/FluentNumberInput.razor
+++ b/src/Core/Components/NumberInput/FluentNumberInput.razor
@@ -1,4 +1,4 @@
-@namespace Microsoft.FluentUI.AspNetCore.Components
+﻿@namespace Microsoft.FluentUI.AspNetCore.Components
 @using Microsoft.FluentUI.AspNetCore.Components.Extensions
 @typeparam TValue
 @inherits FluentInputImmediateBase<TValue>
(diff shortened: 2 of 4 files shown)


dvoituron:
I would like to investigate more about this issue

Issue #5002, opened 2026-07-09
FluentDataGrid column width popup stays open after tabbing out


adamint:
## Repro environment

Observed downstream in the Aspire Dashboard with `Microsoft.FluentUI.AspNetCore.Components` 4.14.1. I also checked the current FluentUI Blazor `dev` branch locally at commit `4cf1c1139` and the same root-cause code appears to still be present.

Proof artifacts are from a browser repro in Aspire's Resources grid using `FluentDataGrid HeaderCellAsButtonWithMenu="true"`, `ResizableColumns="true"`, and `ResizeType="DataGridResizeType.Discrete"`.

Full focus log:
https://raw.githubusercontent.com/adamint/aspire/refs/heads/adamint/fluentui-focus-proof-17926/proof/fluentui-popover-focus/focus-log.jsonl

## Steps to reproduce

1. Render a resizable `FluentDataGrid` with header buttons/menus enabled.
2. Open a column header menu.
3. Choose the `Column width` menu item to open the resize UI.
4. Press `Shift+Tab` from the first resize control.
5. Reopen the resize UI and press `Tab` from the last resize control.

## Expected behavior

When keyboard focus leaves the column width popup before the first control or after the last control, the popup should close and focus should be restored/moved in a predictable way around the invoking header/menu trigger.

This is especially important because the UI is visually popover-like and uses autofocus. After tabbing out, the stale column width popup should not remain open next to a header that no longer owns focus.

## Actual behavior

The column width UI remains open when focus leaves it with `Shift+Tab` or `Tab`.

When pressing `Shift+Tab` from the first control, focus moves back to the original `Name` header button, but the column width popup stays open:

![column width popup remains open after Shift+Tab](https://raw.githubusercontent.com/adamint/aspire/refs/heads/adamint/fluentui-focus-proof-17926/proof/fluentui-popover-focus/09-column-resize-after-shift-tab.png)

When pressing `Tab` from the last control, focus moves to the next `State` header button, but the column width popup remains open under the previous column:

![column width popup remains open after Tab](https://raw.githubusercontent.com/adamint/aspire/refs/heads/adamint/fluentui-focus-proof-17926/proof/fluentui-popover-focus/11-column-resize-after-tab.png)

Relevant active-element snapshots from the log:

```text
07-column-resize-open-autofocus: active fluent-button, text Shrink column width
09-column-resize-after-shift-tab: active fluent-button#..., text Name
[...]

PR #5041, opened 2026-07-22
Fix anchored region focus across shadow DOM and inline popups

### LINKED ISSUE #5001: FluentOverflow MoreButtonTemplate popover loses focus return on Tab/Shift+Tab
## Repro environment

Observed downstream in the Aspire Dashboard with `Microsoft.FluentUI.AspNetCore.Components` 4.14.1. I also checked the current FluentUI Blazor `dev` branch locally at commit `4cf1c1139` and the same root-cause code appears to still be present.

Proof artifacts are from a browser repro in Aspire's Resources grid using `FluentOverflow` + `MoreButtonTemplate` + `FluentPopover`.

Full focus log:
https://raw.githubusercontent.com/adamint/aspire/refs/heads/adamint/fluentui-focus-proof-17926/proof/fluentui-popover-focus/focus-log.jsonl

## Repro shape

The downstream component uses this pattern:

```razor
<FluentOverflow>
    ...
    <MoreButtonTemplate Context="overflow">
        <FluentButton OnClick="() => _popoverVisible = !_popoverVisible">
            +N
        </FluentButton>
    </MoreButtonTemplate>
    <OverflowTemplate Context="overflow">
        <FluentPopover AnchorId="@overflow.IdMoreButton"
                       @bind-Open="_popoverVisible"
                       AutoFocus="false">
            ... focusable links ...
        </FluentPopover>
    </OverflowTemplate>
</FluentOverflow>
```

## Steps to reproduce

1. Render a `FluentOverflow` with a `MoreButtonTemplate` containing a focusable `FluentButton`.
2. Render a `FluentPopover` in `OverflowTemplate` and anchor it to `overflow.IdMoreButton`.
3. Open the popover from the More button.
4. Put focus on the first focusable element in the popover and press `Shift+Tab`.
5. Reopen the popover, put focus on the last focusable element in the popover, and press `Tab`.

## Expected behavior

When tabbing out before the first popover item, focus should return to the overflow More button/template trigger and the popover should close.

When tabbing out after the last popover item, focus should move relative to the actual overflow More button/template trigger. It should not wrap to an unrelated first focusable element in the document, and focus should not be left inside the popover.

## Actual behavior

`Shift+Tab` from the first popover link leaves focus on that same link instead of returning focus to the More button/template trigger.

Before `Shift+Tab`:

focus on first popover link before Shift+Tab

After `Shift+Tab`:

focus remains on first popover link after Shift+Tab

`Tab` from the last popover link jumps to the first focusable element on the page, the Aspire logo, instead of moving relative to the More button/template trigger.

Before `Tab`:

focus on last popover link before Tab

After `Tab`:

focus jumps to page logo after Tab

Relevant active-element snapshots from the log:

```text
01-overflow-more-focused: active fluent-button, text +20
03-overflow-after-shift-tab: active a, text about:blank#overflow-url-1
05-overflow-after-tab: active fluent-anchor, text Aspire
```

## Likely root cause

`FluentOverflow` exposes `IdMoreButton` on a wrapper `div`, not on the actual focusable More button rendered by `MoreButtonTemplate`:

```razor
<div id="@IdMoreButton" class="fluent-overflow-more" ...>
    @if (MoreButtonTemplate != null)
    {
        @MoreButtonTemplate(this)
    }
```

`FluentPopover` passes `AnchorId` to `initializeKeyboardNavigation`:

```csharp
await _anchoredRegionModule.InvokeVoidAsync("initializeKeyboardNavigation", AnchorId, Id, _dotNetHelper, closeKeyCodes);
```

The anchored-region keyboard helper assumes the anchor is focusable and is present in the focusable-elements list:

```js
anchorElement.focus();
...
new FocusableElement(anchorElement.getRootNode()).findNextFocusableElement(startFrom)?.focus();
```

For `FluentOverflow` + `MoreButtonTemplate`, the anchor is the non-focusable wrapper `div#...-more`, while the actual focusable trigger is a descendant `fluent-button` from the template. That causes two problems:

- `anchorElement.focus()` is a no-op for the wrapper `div`, so `Shift+Tab` cannot restore focus to the trigger.
- `findNextFocusableElement(startFrom)` receives a non-focusable wrapper `div`. Because that wrapper is not in `getFocusableElements()`, `currentIndex` becomes `-1`, and the helper wraps to index `0`, which is the first focusable element in the document.

## Possible fix direction

The fix should probably be in FluentUI rather than downstream code. A couple of non-workaround options:

- Let `FluentOverflow` put `IdMoreButton` on the actual focusable More button/template trigger, or provide a way for `MoreButtonTemplate` to bind the anchor id to its focusable trigger.
- Make `initializeKeyboardNavigation` resolve a focusable descendant when `AnchorId` points at a non-focusable composite anchor.
- Make `findNextFocusableElement` handle `currentIndex === -1` without wrapping to the first focusable element in the whole root.

### LINKED ISSUE #5027: Anchored popup Tab navigation wraps to the first page control for Fluent shadow-DOM anchors
## Repro environment

Observed downstream in the Aspire Dashboard with `Microsoft.FluentUI.AspNetCore.Components` 4.14.1. Current Fluent UI `main` still has the same implementation in `FluentAnchoredRegion.razor.js`.

This is a remaining case after #4127 / #4561. Downstream issue and browser tests: https://github.com/microsoft/aspire/issues/17469

Related composite-anchor case: #5001

## Steps to reproduce

1. Anchor a `FluentMenu` or `FluentPopover` to a focusable Fluent web component such as `FluentButton`.
2. Open the popup with the keyboard.
3. Press `Tab` from a menu item, or from the last focusable element in a popover.
4. Inspect the newly focused control.

## Expected behavior

The popup closes and focus moves to the next logical page control after the Fluent anchor.

## Actual behavior

Focus can wrap to the first focusable control in the document. In Aspire's Resources page, the tester observed focus reset from View options/filter popups; a Playwright repro timed out waiting for the next logical control because focus moved elsewhere.

The same downstream scenarios pass when document-order lookup starts from the stable Fluent host and active-element checks resolve the deep shadow-root active element.

## Root cause

For forward Tab, `initializeKeyboardNavigation` intentionally replaces a Fluent anchor host with its first shadow child:

```js
if (anchorElement.tagName.startsWith("FLUENT-") && anchorElement.shadowRoot?.children.length > 0) {
    startFrom = anchorElement.shadowRoot.children[0];
} else {
    startFrom = anchorElement;
}
new FocusableElement(anchorElement.getRootNode()).findNextFocusableElement(startFrom)?.focus();
```

But `FocusableElement(anchorElement.getRootNode()).getFocusableElements()` queries the light DOM with `querySelectorAll("*")`; it does not include that inner shadow child in the root's returned list. `findNextFocusableElement` therefore gets `currentIndex === -1` and computes index `0`, wrapping to the first focusable control on the page:

```js
const currentIndex = filteredElements.indexOf(current);
const nextIndex = (currentIndex + 1) % filteredElements.length;
```

There is a related boundary-detection problem for popup controls inside Fluent shadow roots: `document.activeElement` is the Fluent host while `getFocusableElements()` can contain an inner shadow control, so `activeIndex` can also be `-1`.

## Suggested fix direction

- Keep the stable Fluent host as the document-order starting point instead of replacing it with a shadow child.
- Resolve the deep active element when comparing popup focus boundaries.
- Treat `currentIndex === -1` as an error/lookup case rather than modulo-wrapping to index `0`.
- Add browser coverage with a `FluentButton` anchor for forward Tab and a popup control whose active element is inside a shadow root.

adamint:
# Pull Request

## 📖 Description

Anchored region keyboard navigation currently treats the light DOM focus list as the complete tab order. This causes focus to wrap to the first page control when a menu anchor or popup control is represented through a Fluent component shadow DOM.

This updates focus discovery and active-element mapping to follow the composed tree, including nested open shadow roots and composite anchors.

It also excludes the still-open popup from that traversal. `CloseAsync` is invoked *after* `focus()` and resolves asynchronously, so the popup is still in the DOM while the next element is being located. When a popup is rendered inline and sits between its anchor and the following control — `FluentPopover`, and `FluentMenu` with `DrawMenuWithoutService` — the next focusable element in document order after the anchor *is* the popup's own content, so `Tab` from the last item moved focus back into the popup it had just closed, which then closed underneath it and dropped focus to `<body>`.

### 🎫 Issues

Fixes #5027
Fixes #5001

Related: microsoft/aspire#17469

## 👩‍💻 Reviewer Notes

The fix is isolated to `FluentAnchoredRegion.razor.js`. The only other change is a new Popover demo example, which gives the inline-popup case a permanent home in the demo site.

Two notes for re-review:

- The popup-exclusion commit landed after the first approval, so this needs another look.
- `findNextPageElementAfterAnchor` deliberately does **not** reuse the modulo wrap in `findNextFocusableElement`. Wrapping to index 0 is precisely the symptom reported in microsoft/aspire#17469, so when the anchor is the last control on the page this returns `null` and focus stays put.

This PR now also covers #5001, which was previously addressed by #5003. #5003 has been reduced to the `FluentDataGrid` fix (#5002) so the two no longer overlap.

## 📑 Test Plan

Completed validation:

- `npm run build`
- `dotnet test tests/Core/Microsoft.FluentUI.AspNetCore.Components.Tests.csproj --configuration Release --no-restore --filter FullyQualifiedName~AnchoredRegion --verbosity minimal` (36 passed)
- Focused AnchoredRegion plus Menu test run (37 passed)
- `dotnet build Microsoft.FluentUI.sln --configuration Debug`
- `node --check src/Core/Components/AnchoredRegion/FluentAnchoredRegion.razor.js`

### Automated coverage

None accompanies this change, and it is worth being explicit about why. The logic is entirely in JS; `FluentAnchoredRegionTests` runs with `JSRuntimeMode.Loose`, so bUnit never executes the module and a test there would pass identically with and without the fix. The repository has no JS test runner.

### Browser evidence

I drove this module directly in Chromium from a temporary harness. The DOM places the popup **between** the anchor and the following control, and the anchor is a custom element with an open shadow root and `delegatesFocus: true`, mirroring `fluent-button`:

```html
<button id="before">before</button>
<my-anchor id="anchor"></my-anchor>          <!-- open shadow root, delegatesFocus -->
<div id="popup"><button id="p1"></button><button id="p2"></button></div>
<button id="after">after</button>
```

| Case | Before | After |
| --- | --- | --- |
| Popover: `Tab` from last popup item | `p1` (back into the popup) | `after` |
| Menu (`tabExitsAlways`): `Tab` from any item | `p1` | `after` |
| Anchor is last control on page | `p1` | stays put, no wrap |
| Popover: `Shift+Tab` from first item | anchor's shadow button | anchor's shadow button |
| Popover: `Escape` | anchor's shadow button | anchor's shadow button |
| `CloseAsync` still requested | yes | yes |

One trap worth flagging for anyone reproducing this: assert against the *composed* active element, not `document.activeElement`. Because the anchor uses `delegatesFocus`, `document.activeElement` reports the host element while `anchor.shadowRoot.activeElement` reports the real focused control — which makes a correct result look like a failure.

```js
let element = document.activeElement;
while (element?.shadowRoot?.activeElement) element = element.shadowRoot.activeElement;
element;
```

### Manual repro — regression check (#5027)

Run the demo server with the **Run Demo Server** VS Code launch profile, or:

```bash
dotnet run --project examples/Demo/Server/FluentUI.Demo.Server.csproj --framework net9.0 --launch-profile FluentUI.Demo.Server
```

1. Open `http://localhost:5062/datagrid-menu-header`.
2. `Tab` to the `PersonId` column-header menu button and press `Enter` to open it.
3. Press `Tab` from a menu item. The menu should close and focus should advance to the next logical control instead of wrapping to the first page control.
4. Reopen and verify `Shift+Tab` and `Escape` close it and return focus to the column-header button.
5. Repeat with `Name` and `BirthDate`.

Note this page renders through `FluentMenuProvider`, which places the menu at the end of the layout — far from the anchor in document order. It confirms the shadow-DOM traversal but the popup filter is a no-op here, so it cannot exercise the second fix.

### Manual repro — inline popup (#5001)

This PR adds a **Keyboard navigation** example to `/Popover` where the popover is declared between its anchor and the following control, matching the harness above. No existing demo page covered that arrangement: on `/Popover` both `FluentPopover` elements are declared after both buttons, so the next control after callout 1's anchor is callout 2's button, and `/Overflow` has no `MoreButtonTemplate` example.

1. Open `http://localhost:5062/Popover` and scroll to **Keyboard navigation**.
2. `Tab` to **Open Callout** and press `Enter`.
3. `Tab` through **Name** and **Save**, then press `Tab` once more.
4. Expected: the callout closes and focus lands on **Next control**.
5. Before this change focus moved back into the closing callout instead, ending on `<body>` once `CloseAsync` completed.
6. Also confirm `Shift+Tab` from **Name** and `Escape` both return focus to **Open Callout**.

### General

### Component-specific

- [ ] I have added Unit Tests for my new component
- [x] I have validated the Unit Tests for an existing component

## ⏭ Next Steps

None.


vnbaaij (approved):

adamint:
@vnbaaij will this fix be included in the next release?

adamint:
Heads up @vnbaaij — I pushed a commit after your approval, so this needs another look before merge. Auto-merge is currently off.

While double-checking this against the cases in microsoft/aspire#17469 I found one the PR didn't cover. `CloseAsync` is invoked *after* `focus()` and resolves asynchronously, so the popup is still in the DOM while we're locating the next element. When the popup is rendered inline next to its anchor — `FluentPopover`, and `FluentMenu` with `DrawMenuWithoutService` — the next focusable element in document order after the anchor *is* the popup's own content. So `Tab` from the last item moved focus back into the popup, which then closed underneath it and dropped focus to `<body>`.

The fix filters the popup's composed subtree out of the candidate list before locating the anchor:

```js
const candidates = focusable.getFocusableElements()
    .filter(element => !containsComposedElement(popupElement, element));
```

Measured in Chromium against this module, with the popup placed between the anchor and the following control:

| Case | Before | After |
| --- | --- | --- |
| Popover: `Tab` from last popup item | `p1` (back into the popup) | `after` |
| Menu (`tabExitsAlways`): `Tab` from any item | `p1` | `after` |
| Anchor is last control on page | `p1` | stays put, no wrap |
| `Shift+Tab` from first item | anchor's shadow button | anchor's shadow button |
| `Escape` | anchor's shadow button | anchor's shadow button |

Why the original manual testing missed it: I validated on `datagrid-menu-header`, which renders through `FluentMenuProvider` at the end of the layout. There the anchor and popup are nowhere near each other in document order, so the filter is a no-op and the old code looked correct. It only breaks for inline popups.

Two things worth your judgement:

- `findNextPageElementAfterAnchor` intentionally does not reuse the modulo wrap in `findNextFocusableElement`. Wrapping to index 0 is exactly the reported symptom, so when the anchor is the last control the function returns `null` and focus stays where it is.
- This now also closes #5001. I've reduced #5003 to just the `FluentDataGrid` fix (#5002) so the two PRs no longer touch the same file.

No test accompanies this. The logic is entirely in JS, and `FluentAnchoredRegionTests` runs with `JSRuntimeMode.Loose`, so bUnit never executes the module — a test there would pass identically with or without the fix. The only suite that could catch this was the Playwright one removed earlier in this PR, and re-adding it isn't in scope here.

adamint:
Added a **Keyboard navigation** example to the Popover demo page so the inline-popup case has a permanent repro instead of a paste-in snippet. The popover is declared between its anchor and the following control, which is the arrangement none of the existing examples covered.

Verified against the running demo server on `/Popover`, reading the *composed* active element at each step:

| Action | Composed focus | Callout |
| --- | --- | --- |
| Open, then `Tab` | `input` (Name) | open |
| `Tab` | Save button | open |
| `Tab` | **Next control** | closed |
| `Shift+Tab` from Name | anchor | closed |
| `Escape` | anchor | closed |

Before the fix the third `Tab` re-entered the closing callout and focus ended on `<body>`.

The PR is now three files: the one-file fix plus the demo example and its registration.

dvoituron (approved):

vnbaaij (approved):

PR #5044, opened 2026-07-27
[dev-v5] [Charts] Add FluentAreaChart

vnbaaij:
Just need to add more examples...

Adds the `FluentAreaChart` web component wrapper.

image

## Unit tests

image


examples/Demo/FluentUI.Demo/Program.cs
```diff
@@ -27,6 +28,7 @@
     // config.DefaultValues.ForAny<FluentAutocomplete<object, object>>().Set(p => p.Width, "100%");
     // config.DefaultValues.ForAny<FluentAutocomplete<object, object>>().Set(p => p.Multiple, false);
 
+    config.DefaultValues.ForAny<FluentChartBase>().Set(p => p.RoundedCorners, true);
```
dvoituron: Probably this line needs the PR #4999. Correct?

dvoituron (changes requested):

examples/Demo/FluentUI.Demo/Program.cs
vnbaaij: Yes, that was left over from a test. Removed with next commit

dvoituron (approved):

PR #5185, opened 2026-08-21
[dev-v5] Refactor FluentTooltip 

dvoituron:
# [dev-v5] Refactor FluentTooltip 

- In the default service mode, the issue appears to be caused by the link between a tooltip and its anchored element. FluentTooltipProvider associates them using Blazor’s auto-generated element ID. When content re-renders, Blazor may reuse that ID for a different element, while the tooltip remains linked to the previous registration.
As a result, the tooltip may become associated with the wrong anchor or lose its anchor entirely. 

- Refactor `FluentTooltip` and `FluentTooltipProvider` to utilize **default values for delay and positioning**. 
- Add tests to ensure the correct application of these default configurations. 

### Before

https://github.com/user-attachments/assets/3c46e69f-39a4-4bdd-95df-990bc77401c6

### After

https://github.com/user-attachments/assets/1b6dfbbd-6b73-4f9e-9ba2-09cbc22dfa20

```razor
<FluentButton OnClick="@RerenderButtonAsync">Re-render button</FluentButton>

@if (_isLoading)
{
    <FluentButton @key="_renderVersion" Loading="true" Tooltip="Loading...">Loading...</FluentButton>
}
else
{
    <FluentButton @key="_renderVersion" Tooltip="Delete">Delete</FluentButton>
}

@code
{
    private bool _isLoading;
    private int _renderVersion;

    private async Task RerenderButtonAsync()
    {
        _isLoading = true;
        _renderVersion++;

        await Task.Delay(500);

        _isLoading = false;
        _renderVersion++;
    }
}
```

## Unit Tests

Added

### DIFF
--- a/src/Core/Components/Tooltip/Services/LibraryTooltipOptions.cs
+++ b/src/Core/Components/Tooltip/Services/LibraryTooltipOptions.cs
@@ -29,7 +29,7 @@ public class LibraryTooltipOptions
 
     /// <summary>
     /// Gets or sets number of milliseconds to delay the tooltip from showing/hiding on hover.
-    /// Default is 250ms.
+    /// The default value is `null`. Internally the component uses 250ms when no value is provided.
     /// </summary>
     public int? Delay { get; set; }
 }
--- a/src/Core/Components/Tooltip/Services/LibraryTooltipOptions.cs
+++ b/src/Core/Components/Tooltip/Services/LibraryTooltipOptions.cs
@@ -21,4 +21,15 @@ public class LibraryTooltipOptions
     /// If set to true, add the FluentTooltipProvider component at end of the MainLayout.razor page.
     /// </summary>
     public bool UseServiceProvider { get; set; } = true;
+
+    /// <summary>
+    /// Gets or sets the default tooltip positioning.
+    /// </summary>
+    public Positioning? Positioning { get; set; }
+
+    /// <summary>
+    /// Gets or sets number of milliseconds to delay the tooltip from showing/hiding on hover.
+    /// Default is 250ms.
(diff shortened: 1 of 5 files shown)


src/Core/Components/Tooltip/Services/LibraryTooltipOptions.cs
```diff
@@ -21,4 +21,15 @@ internal LibraryTooltipOptions()
...
+    /// <summary>
+    /// Gets or sets the default tooltip positioning.
+    /// </summary>
+    public Positioning? Positioning { get; set; }
+
+    /// <summary>
+    /// Gets or sets number of milliseconds to delay the tooltip from showing/hiding on hover.
+    /// Default is 250ms.
+    /// </summary>
+    public int? Delay { get; set; }
```
vnbaaij, line 34: Comment says default is 250ms but type is nullable int so actual parameter default is null.
I think we use different wording elsewhere to indicate that we use the web component default value

vnbaaij (changes requested):

src/Core/Components/Tooltip/Services/LibraryTooltipOptions.cs
dvoituron: That is the exact same comment in the FluentTooltip component. How can we give this info using another way (without referencing the web component)?

dvoituron: Updated to 
> The default value is `null`. Internally the component uses 250ms when no value is provided.

vnbaaij (approved):

PR #5197, opened 2026-08-24
fix: [Inputs] Immediate mode hides an external Value assignment made while the field has focus

### LINKED ISSUE #5196: fix: Immediate mode hides an external Value assignment made while the field has focus
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

In `Immediate` mode, the rendered `value` of an input is frozen for the **whole period the field has focus**. As a result, any value the consuming app assigns from the outside while the field has focus never reaches the browser: the bound property is correct, but the native input keeps showing the old text until the field loses focus, and then the new value appears all at once.

When the app deliberately **returns focus to the field** — a very common keyboard-navigation pattern: open a lookup/selection dialog from the field, pick a row, close the dialog and focus the field it belongs to — the value is never shown at all.

The freeze comes from `FluentInputImmediateManager.GetImmediateValueAsString` (before #5181 it lived in `FluentInputImmediateBase`):

```csharp
internal string? GetImmediateValueAsString(string? currentValueAsString) => !_input.Immediate
    ? currentValueAsString
    : _hasFocus
        ? _frozenValueAsString ??= _pendingImmediateText ?? currentValueAsString   // <-- frozen
        : _pendingImmediateText ?? currentValueAsString;
```

I understand the freeze is deliberate and load-bearing: it stops a re-render from resetting the native input to a stale string while the user is still typing (the browser may hold keystrokes the server hasn't received yet), and it works by keeping the rendered string constant so Blazor's diff never touches the DOM. The problem is only its **scope**: it protects against the user's own typing, but it also blocks assignments that have nothing to do with typing. An external assignment is newer than anything the user typed, so it should win.

Note that `FluentInputText_NotImmediate_WhileFocused_ValueAttributeReflectsExternalUpdates` already asserts this exact behaviour for the non-immediate case ("the freeze must not apply - the DOM reflects the update right away, not after blur"). The same expectation is not met in `Immediate` mode.

### Expected Behavior

An external assignment to `Value` while the field has focus should reach the browser, in `Immediate` mode too — while still not clobbering text the user is actively typing.

### Steps To Reproduce

1. Render a `FluentTextInput` with `Immediate="true"` and a bound `Value`.
2. Give the input focus.
3. Without typing anything, assign a new `Value` from the app (e.g. from a dialog result, a validation rule, or another component).
4. The native input keeps showing the old value. It only updates after the field loses focus.

Minimal bUnit repro (fails on `dev-v5`):

```csharp
var cut = Render(@<FluentTextInput Value="init" Immediate="true" ImmediateDelay="0" />);
var textInput = cut.FindComponent<FluentTextInput>();
cut.Find("fluent-text-input").TriggerEvent("onfocusin", new FocusEventArgs());

textInput.Render(parameters => parameters.Add(p => p.Value, "external"));

// expected "external", actual "init"
Assert.Equal("external", cut.Find("fluent-text-input").GetAttribute("value"));
```

### Exceptions (if any)

None — no exception, no console error. The bound property holds the right value; only the DOM does not reflect it.

### .NET Version

10.0

### Anything else?

Affects every component deriving from `FluentInputImmediateBase` when `Immediate` is enabled — `FluentTextInput`, `FluentTextArea`, `FluentNumberField`, `FluentCombobox` (since #5181), etc.

I have a fix and will open a PR: scope the freeze to the user's **typing** instead of to focus — hold it while the confirmed value keeps following what typing produced, and lift it as soon as the confirmed value diverges from that trajectory (which means the consumer assigned it).

One approach I want to explicitly rule out, in case it looks tempting: keying the freeze on `_pendingImmediateText is not null` ("freeze only while a keystroke is in flight"). That flag only covers the server-side processing window, so it leaves the client-side debounce — and the gap between one confirmed keystroke and the next — unprotected, which is exactly what the freeze exists for. It fixes the dialog case and breaks typing.

christian-sig:
Fixes #5196

## The problem

In `Immediate` mode the rendered `value` was frozen for the **whole period the field has focus**, so any value the consuming app assigned from the outside while the field had focus never reached the browser: it stayed invisible until the field lost focus and then appeared all at once. When the consumer deliberately returns focus to the field — closing a lookup/selection dialog and focusing the field it belongs to, a common keyboard-navigation pattern — the value was never shown at all, even though the bound property was correct.

## Why the freeze exists, and what actually needed changing

The freeze is load-bearing and I kept it. It stops a re-render from resetting the native input to a stale string while the user is still typing (the browser may hold keystrokes the server hasn't received yet), and it works by keeping the rendered string **constant** so Blazor's diff never touches the DOM.

What was wrong was its **scope**. That protection is about the user's own typing, not about arbitrary assignments: an external assignment is newer than anything the user typed and has to win. `FluentInputText_NotImmediate_WhileFocused_ValueAttributeReflectsExternalUpdates` already asserts exactly this for the non-immediate path ("the freeze must not apply - the DOM reflects the update right away, not after blur"); this PR makes `Immediate` mode meet the same expectation.

## The change

The freeze is now scoped to typing instead of to focus. It holds while the confirmed value keeps following what typing produced (`_expectedConfirmedValue`, re-baselined after each confirmed keystroke via `_typingConfirmed`), and lifts as soon as the confirmed value diverges from that trajectory — a divergence means the consumer assigned the value.

`_typingConfirmed` deliberately does **not** predict the new value from the event args: the component's own parse/format round trip (numeric, date, masked input) can legitimately change the string, so the next read re-baselines against whatever the confirmed value actually became.

Also replaced the `??=` freeze latch with an explicit `_isFrozen` flag, so freezing on a `null`/empty value sticks instead of re-evaluating on every read.

## Rejected alternative (worth recording)

Keying the freeze on `_pendingImmediateText is not null` — "freeze only while a keystroke is in flight" — looks like the obvious one-liner and is wrong. That flag is only set while a keystroke is being processed server-side, so it leaves the **client-side debounce window**, and the gap between one confirmed keystroke and the next, unprotected. That is precisely what the freeze is for (see technical note 3 in the file). It fixes the dialog case and reintroduces the typing corruption the note describes. I left a comment in the code so it does not get "simplified" into that later.

## Tests

Three added to `FluentTextInputTests`:

- `FluentInputText_Immediate_WhileFocused_ValueAttributeReflectsExternalUpdates` — external assignment with focus and no prior typing.
- `FluentInputText_Immediate_ExternalUpdateAfterTyping_WhileStillFocused_Reflected` — external assignment after the user typed, still focused.
- `FluentInputText_Immediate_RepeatedTyping_KeepsValueAttributeFrozen` — several confirmed keystrokes in a row keep the rendered value frozen, i.e. the typing protection is intact.

The existing `FluentInputText_Immediate_WhileFocused_ValueAttributeStaysFrozen` and `FluentInputText_Immediate_OnFocusOut_UnfreezesAndResyncsValueAttribute` still pass unchanged. Full Core suite: 3865/3865.

## Verified in a real app

Blazor Server app, `FluentTextInput` with `Immediate="true"`: picking a row in a lookup dialog now shows the value immediately with focus returned to the field (attribute, property and shadow `input` all updated). Counter-check on typing: while typing in another field the rendered attribute stays constant and the input keeps exactly what was typed — no reset.

### DIFF
--- a/src/Core/Components/Base/FluentInputImmediateManager.cs
+++ b/src/Core/Components/Base/FluentInputImmediateManager.cs
@@ -83,12 +83,25 @@ namespace Microsoft.FluentUI.AspNetCore.Components;
          still typing, it can reset the field to a stale, shorter string mid-keystroke, scrambling
          what the user typed - confirmed via manual and scripted repro on a throttled/high-latency
          connection.
-       - Fix: the rendered value is FROZEN while the field has focus (_hasFocus /
-         _frozenValueAsString) - Blazor never touches the native `value` while the user might
-         still be typing, and resyncs to the confirmed value only once the field loses focus
-         (FocusOutHandlerAsync). This is a hard guarantee, not a timing heuristic: a fixed-duration
-         "wait a bit before rendering" timer was considered and rejected, since it can always be
-         defeated by high-enough or variable network latency.
+       - Fix: while the field has focus the rendered value is FROZEN (_hasFocus / _isFrozen /
+         _frozenValueAsString) - it is kept CONSTANT, so Blazor's diff never touches the native
+         `value` while the user might still be typing, and it resyncs to the confirmed value once
+         the field loses focus (FocusOutHandlerAsync). This is a hard guarantee, not a timing
+         heuristic: a fixed-duration "wait a bit before rendering" timer was considered and
+         rejected, since it can always be defeated by high-enough or variable network latency.
+       - The freeze is scoped to the user's own typing, not to the whole focused period. It holds
+         while the confirmed value keeps following what typing produced (_expectedConfirmedValue,
+         re-baselined on each confirmed keystroke via _typingConfirmed), and lifts as soon as the
+         confirmed value diverges from that: a divergence means the CONSUMER assigned the value
+         from the outside (a lookup dialog selection, a validation rule rewriting the field, a
+         value pushed by another component), which is newer than anything the user typed and has
+         to reach the browser. Freezing on focus alone made such an assignment invisible until the
+         field lost focus - and permanently invisible for a consumer that returns focus to the
+         field after closing a dialog, a common keyboard-navigation pattern.
+       - This cannot be keyed on _pendingImmediateText instead ("freeze only while a keystroke is
+         in flight"): that flag only covers the server-side processing window, leaving the
+         client-side debounce - and the gap between one confirmed keystroke and the next -
+         unprotected, which is exactly what step 3 exists to protect.
 
     4. Render suppression (ShouldRender)
        - Re-renders are also skipped, more generally, while a keystroke is known but not yet
@@ -107,7 +120,20 @@ internal sealed class FluentInputImmediateManager
     private readonly SemaphoreSlim _immediateGate = new(1, 1);
     private string? _pendingImmediateText;
     private bool _hasFocus;
+    private bool _isFrozen;
     private string? _frozenValueAsString;
+
+    // Confirmed value the freeze was taken against, kept up to date with every value the user's own
+    // typing confirms. While the confirmed value follows this trajectory there is nothing new to show
+    // and the freeze holds; the moment it diverges, the consumer assigned the value from the outside
+    // and that assignment has to reach the browser (see GetImmediateValueAsString).
+    private string? _expectedConfirmedValue;
+
+    // Set when this field's own typing has just confirmed a value: the next read re-baselines
+    // _expectedConfirmedValue to whatever the confirmed value became, instead of reading the move as
+    // an external assignment. Deliberately not predicted from the event value - the component's own
+    // parse/format round trip (numeric, date, masked input) can legitimately change the string.
+    private bool _typingConfirmed;
 
     public FluentInputImmediateManager(IFluentInputImmediate input)
     {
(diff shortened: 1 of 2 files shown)


src/Core/Components/Base/FluentInputImmediateManager.cs
```diff
@@ -107,8 +120,21 @@ internal sealed class FluentInputImmediateManager
...
+    // typing confirms. While the confirmed value follows this trajectory there is nothing new to show
+    // and the freeze holds; the moment it diverges, the consumer assigned the value from the outside
+    // and that assignment has to reach the browser (see GetImmediateValueAsString).
+    private string? _expectedConfirmedValue;
+
+    // Set when this field's own typing has just confirmed a value: the next read re-baselines
+    // _expectedConfirmedValue to whatever the confirmed value became, instead of reading the move as
+    // an external assignment. Deliberately not predicted from the event value - the component's own
+    // parse/format round trip (numeric, date, masked input) can legitimately change the string.
+    private bool _typingConfirmed;
```
dvoituron, line 136: I need to understand exactly what the problem is (in what situations it occurs) and whether this solution can cover all possible scenarios. 

The description of the issue seems clear, but I need a little time to work on it.

dvoituron (changes requested):

dvoituron:
Thank you for the detailed investigation and for identifying the issue. I agree with the main conclusions: the focus-based freeze is necessary to protect pending browser input, but an explicit external `Value` update must still be displayed while the component is focused.

After reviewing the implementation, I would prefer a simpler approach that detects external updates in `SetParametersAsync`, where Blazor already provides the information we need: See PR #5198

My main concern with the current PR is that `_isFrozen`, `_expectedConfirmedValue`, and `_typingConfirmed` form an implicit state machine inside the rendering path. This makes the behavior harder to understand and introduces ordering-sensitive assumptions.

In particular, `_typingConfirmed` means “the next observed value comes from this input”. That may not always hold if the parent normalizes or replaces the value from `ValueChanged` before the next render. Such a value could be re-baselined as typing instead of being recognized as an external assignment. It is also set from a finally block, including when `changeHandlerAsync` throws, even though no successful confirmation may have occurred.

Detecting the external assignment directly in `SetParametersAsync` avoids these inferences and keeps the existing focus and latency protection unchanged.

So I’d prefer to close this PR. I’ve still included exactly the same unit tests, though.

PR #5234, opened 2026-09-04
[dev-v4] Update package versions 

dvoituron:
# [dev-v4] Update package versions 

Update the package versions to the latest features and fixes. 
This change does not introduce breaking changes.

## Roslyn packages intentionally not updated
`Microsoft.CodeAnalysis.CSharp` and `Microsoft.CodeAnalysis.Analyzers` must remain at version `5.6.0`. 
Updating them to `5.9.0` makes the Design Token source generator reference a newer compiler than the one included with the current .NET SDK (10.0.303).

Consequently, the compiler cannot load the source generator, AddDesignTokens is not generated, and the Core project fails to build. These packages can only be updated once the repository uses an SDK that includes a compatible Roslyn compiler.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -23,7 +23,7 @@
     <PackageVersion Include="xunit.runner.visualstudio" Version="3.1.5" />
     <PackageVersion Include="coverlet.msbuild" Version="10.0.1" />
     <PackageVersion Include="coverlet.collector" Version="10.0.1" />
-    <PackageVersion Include="AngleSharp.CSS" Version="1.0.2" />
+    <PackageVersion Include="AngleSharp.CSS" Version="1.0.0-beta.154" />
     <!-- Shared dependencies -->
     <PackageVersion Include="Markdig.Signed" Version="1.3.2" />
     <PackageVersion Include="Microsoft.CodeAnalysis.Analyzers" Version="5.6.0" />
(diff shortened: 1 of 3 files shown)


Directory.Packages.props
```diff
@@ -18,12 +18,12 @@
...
     <!-- Test dependencies -->
     <PackageVersion Include="bunit" Version="1.40.0" />
-    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="18.8.1" />
+    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="18.9.0" />
     <PackageVersion Include="xunit" Version="2.9.3" />
     <PackageVersion Include="xunit.runner.visualstudio" Version="3.1.5" />
     <PackageVersion Include="coverlet.msbuild" Version="10.0.1" />
     <PackageVersion Include="coverlet.collector" Version="10.0.1" />
-    <PackageVersion Include="AngleSharp.CSS" Version="1.0.0-beta.154" />
+    <PackageVersion Include="AngleSharp.CSS" Version="1.0.2" />
```
vnbaaij: This leads to the bUnit/AngleSharp error

vnbaaij (changes requested):

Directory.Packages.props
dvoituron: Rollbqcked

vnbaaij (approved):

PR #5239, opened 2026-09-06
[dev-v5] fix: guard JavaScript module imports against component disposal

JamesNK:
# Pull Request

## Description

Handle component disposal while a JavaScript module import is pending, and clean up modules that arrive after the component has already been disposed.

The motivating report was an unhandled exception in the Aspire dashboard's Blazor Server circuit:

```text
fail: Aspire.Hosting.Dashboard.ThirdParty.Microsoft.AspNetCore.Components.Server.Circuits.CircuitHost[111]
      Unhandled exception in circuit.
      System.ObjectDisposedException: Cannot access a disposed object.
      Object name: 'Microsoft.JSInterop.Implementation.JSObjectReference'.
         at Microsoft.JSInterop.Implementation.JSObjectReference.InvokeAsync[TValue](String identifier, Object[] args)
         at Microsoft.FluentUI.AspNetCore.Components.FluentDataGrid`1.OnAfterRenderAsync(Boolean firstRender)
         at Microsoft.AspNetCore.Components.RenderTree.Renderer.GetErrorHandledTask(Task taskToHandle, ComponentState owningComponentState)
```

Blazor components can be disposed while an asynchronous lifecycle method is awaiting work. The existing import-then-initialize paths continued with JavaScript initialization without checking whether disposal occurred during the import. If disposal ran before the imported module reference was available, it could see a null reference and leave the subsequently imported module uncleaned.

### Changes

- Add `IsDisposed` to `IFluentComponentBase`, implemented by both `FluentComponentBase` and `FluentInputBase` and set before base disposal cleanup begins. The options classes implementing this interface explicitly return false because they have no disposal lifecycle.
- Pass the owning `IFluentComponentBase` into `FluentJSModule` and put public `TryImportJavaScriptModuleAsync` on that wrapper. It imports the module and returns false when the owning component has started disposal, otherwise true.
- Coordinate module disposal ownership inside `FluentJSModule`. Both component base classes claim an available module before awaiting component-specific cleanup. The late-import helper disposes a module only if component disposal did not already claim it. This preserves cleanup order and avoids repeated module disposal through these paths.
- Make the raw `ImportJavaScriptModuleAsync` method internal. Update XML documentation to direct callers to the guarded helper and require checking its boolean result before using the reference.
- Update all ten component import sites across DataGrid, Calendar, Grid, InputFile, Nav, NavCategory, PullToRefresh, Tooltip, and TreeView to return early on false before invoking JavaScript or creating initialization references. Calendar now shares the implementation through FluentJSModule despite belonging to the input-component hierarchy.
- Add ten regression cases covering active and late imports, component/input disposal state, cleanup ordering, repeated import/disposal, and rendered Grid/Calendar components skipping initialization after disposal.

### Issues

No linked issue. Reported while using FluentDataGrid in the Aspire dashboard.

## Reviewer Notes

This change is deliberately scoped to the import/disposal boundary. It does not add blanket exception handling, and it does not address disposal between later awaited JavaScript calls after a successful import. The supplied stack trace does not identify which invocation in OnAfterRenderAsync failed.

The delayed-import and cleanup tests use controlled TaskCompletionSource instances, not timing delays. The rendered component regressions explicitly await the lifecycle task; their module fake rejects any initialization invocation. The disposal-ownership regressions were verified to fail before the ownership fix and pass afterward. No API visibility assertion was added.

### API Compatibility

This dev-v5 change restricts the previously public raw import method to internal access and changes the FluentJSModule constructor to require its owning component. Custom IFluentComponentBase implementations must provide IsDisposed. External module users should supply an owner and call TryImportJavaScriptModuleAsync, checking for true before invoking JS. These are source/API compatibility changes for consumers of those extensibility types.

## Test Plan

- `dotnet test tests/Core/Components.Tests.csproj --configuration Debug`: **3,912 passed, 1 skipped, 0 failed** after the refactor. The skipped test requires a Release build with Core targeting .NET 9.
- Shared lifecycle tests: **22 passed**, including input cleanup and delayed Calendar initialization.
- `git diff --check`: passed.
- Editor diagnostics for the changed files: clean.
- Initial `dotnet build Microsoft.FluentUI-v5.slnx --configuration Debug`: **blocked during restore** by NU1603 in the unchanged Demo.Client and WasmStandalone projects. Microsoft.NET.Sdk.WebAssembly.Pack version 11.0.0-preview.4.26203.108 was unavailable; NuGet resolved 11.0.0-preview.4.26230.115 instead, and warnings are treated as errors. No dependency changes are included in this PR. The solution build was not rerun for the refactor.

### General

- [x] I have documented the guarded helper and cleanup ownership in XML documentation/code comments, with API compatibility notes above.
- [x] I have read the contributing documentation and followed the project conventions.

### Component-specific

- [x] I have modified existing components.
- [x] I have validated the Core unit tests.

## Next Steps

Handling circuit disconnection and disposal during subsequent interop calls remains separate follow-up work.


src/Core/Components/Base/FluentComponentBase.cs
```diff
@@ -102,6 +107,25 @@ protected FluentComponentBase(LibraryConfiguration configuration)
...
+    /// <param name="file">The path of the JavaScript module to import.</param>
+    /// <returns><see langword="true"/> if the component is still active; otherwise, <see langword="false"/>.</returns>
+    protected async Task<bool> TryImportJavaScriptModuleAsync(string file)
+    {
+        await JSModule.ImportJavaScriptModuleAsync(file);
+        if (IsDisposed)
+        {
+            // The module may have been null when DisposeAsync ran while the import was pending.
+            // Dispose it now so the completed import doesn't leak the module reference.
+            await JSModule.DisposeAsync();
```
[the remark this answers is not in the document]
JamesNK: Addressed in daec3c3b5. Module disposal now has one owner: base `DisposeAsync` captures an already-imported module and claims its disposal before awaiting component-specific cleanup. The import helper returns false without disposing a module claimed by that cleanup. If base disposal saw no module, the late-import helper claims and disposes it instead. Repeated base disposal calls are also ignored.

Added three deterministic regression cases: import during paused component cleanup, repeated import/disposal after normal cleanup, and repeated import/disposal after late-import cleanup. All three failed before the fix (early disposal during cleanup, or three DisposeAsync calls instead of one) and now pass. The original delayed-import tests continue to pass. Full Core suite: 3,909 passed, 1 Release-only test skipped.

This preserves the requested import guard without adding exception suppression or changing FluentJSModule's public behavior.

src/Core/Components/Base/FluentComponentBase.cs
```diff
@@ -102,6 +108,30 @@ protected FluentComponentBase(LibraryConfiguration configuration)
     [Parameter(CaptureUnmatchedValues = true)]
     public virtual IReadOnlyDictionary<string, object>? AdditionalAttributes { get; set; }
 
+    /// <summary>
+    /// Imports the JavaScript module and disposes it if the component has been disposed.
+    /// </summary>
+    /// <param name="file">The path of the JavaScript module to import.</param>
+    /// <returns><see langword="true"/> if the component is still active; otherwise, <see langword="false"/>.</returns>
+    protected async Task<bool> TryImportJavaScriptModuleAsync(string file)
```
dvoituron: Good idea this "Try" method.

- Could you use it in the `FluentCalendar.razor.cs` file to standardize the code?
- Could you change the method `public async Task<IJSObjectReference> ImportJavaScriptModuleAsync(string file)` to `internal` in order to recommend using ONLY the “Try” method?
- Could you update the XML document by replacing `ImportJavaScriptModuleAsync` with `TryImportJavaScriptModuleAsync`?

dvoituron (changes requested):

src/Core/Components/Base/FluentComponentBase.cs
dvoituron: I just saw your comment "FluentCalendar remains unchanged because it inherits from the separate FluentInputBase hierarchy, not FluentComponentBase."

That's right… so forget my first comment, but maybe we could update the “internal” and the documentation.

JamesNK: Addressed in d5abf4cb0. The raw `ImportJavaScriptModuleAsync` method is now internal, and the XML documentation points to the guarded method and explicitly requires checking its boolean result before using the module.

I also moved `TryImportJavaScriptModuleAsync` and the disposal-ownership flag into `FluentJSModule`, which now receives its owning `IFluentComponentBase`. Both `FluentComponentBase` and `FluentInputBase` implement the new `IsDisposed` interface property, so Calendar can now use the same guarded import without duplicating the helper. Existing callers use `JSModule.TryImportJavaScriptModuleAsync`.

Added input disposal/cleanup-ordering tests and a rendered Calendar delayed-import regression. Full Core suite: 3,912 passed, 1 Release-only test skipped. No API visibility assertion was added.

MarvinKlein1508:
TODO: Once this got merged I need to update the community components to use the new `TryImportJavaScriptModuleAsync` method.

vnbaaij (approved):

JamesNK:
You could consider making IsDisposed an internal DIM property on IFluentComponent. However that will block people from accurately creating their own types that implement IFluentComponent (they can still inherit from FluentComponentBase).

dvoituron (approved):

PR #5261, opened 2026-09-10
perf: generate enum attribute value lookups

### LINKED ISSUE #5260: Improve ToAttributeValue performance by avoiding repeated enum reflection
## Summary

`EnumExtensions.ToAttributeValue()` resolves `[Description]` by performing reflection on every call:

```csharp
var fieldInfo = value.GetType().GetField(value.ToString());
var attribute = fieldInfo?.GetCustomAttribute<DescriptionAttribute>();
```

PR #5259 narrows the previous `GetMember` lookup to `GetField`, fixing the broad Native AOT analysis path. However, repeatedly looking up `FieldInfo` and constructing attribute instances remains much slower and more allocation-heavy than cached or generated mappings. This matters because `ToAttributeValue()` is used while rendering components.

Related: #136 previously proposed replacing `ToAttributeValue()` with `ToString().ToLowerInvariant()`. That issue was closed because attribute-backed values must be preserved and there was no conclusive performance data. This issue retains `[Description]` behavior and provides measurements.

## Benchmark

Focused microbenchmark on .NET 10 preview, Windows x64, Release configuration. Native AOT was published for `win-x64`.

- Five measured passes of 2,000,000 lookups after warmup
- Every implementation accepts `Enum`, matching the current public API
- Inputs alternate between two concrete enum types
- Inputs are pre-boxed to isolate lookup cost
- `GetMember` and `GetField` include `ToString()`, metadata lookup, `GetCustomAttribute<DescriptionAttribute>()`, and fallback handling
- The dictionary is one read-only-after-initialization `Dictionary<Enum, string>` containing values from multiple enum types

Median results:

| Strategy | Optimized JIT | Native AOT | Allocation per lookup (JIT / AOT) |
|---|---:|---:|---:|
| `GetMember(Enum)` + attribute | ~714 ns | ~2,560 ns | 288 B / 1,160 B |
| `GetField(Enum)` + attribute | ~619 ns | ~2,340 ns | 256 B / 1,128 B |
| `Dictionary<Enum, string>` | ~15 ns | ~25 ns | 0 B / 0 B |
| `switch (Enum)` | ~2.9 ns | ~3.1 ns | 0 B / 0 B |

The heterogeneous dictionary is approximately 40x faster than `GetField` under JIT and 95x faster under Native AOT. The switch is approximately 210x faster under JIT and 750x faster under Native AOT.

Passing a concrete enum directly to an `Enum` parameter adds the same boxing cost to every strategy and was excluded from these measurements.

### 1. Cache values in a dictionary

Use a cache keyed by `Enum`, which naturally supports multiple enum types because enum equality includes the concrete enum type.

A lazy `ConcurrentDictionary<Enum, string>` could reflect only on first access to each value and avoid enumerating all values. Subsequent calls would use the cache. This is the smallest behavioral change, but first access still uses reflection, the cache retains boxed enum keys, and dynamically populated caches need careful AOT/trimming validation.

A pre-generated `Dictionary<Enum, string>` would remove runtime reflection and allocations entirely while retaining the existing `Enum` API.

### 2. Add internal switch-based mappings

Keep the public methods accepting `Enum`, then dispatch internally by concrete enum type and value:

```csharp
private static string GetAttributeValue(Enum value) => value switch
{
    Color.Primary => "var(--colorBrandForeground1)",
    Orientation.Horizontal => "horizontal",
    // ...
};
```

This is the fastest option and requires no reflection or runtime cache. The main drawbacks are maintenance, generated-code size, and ensuring fallback behavior remains consistent for unknown and flags values.

### 3. Generate mappings at compile time

Use a Roslyn source generator to scan Fluent UI enum fields and emit either switch-based mappings or a heterogeneous dictionary. This keeps `[Description]` as the source of truth while eliminating runtime reflection and manual duplication.

Generation should happen at compile time rather than initializing the cache with `Enum.GetValues(Type)`. Microsoft documents `Enum.GetValues(Type)` as `RequiresDynamicCode`; it can fail under Native AOT because creating an array of a runtime-determined enum type might not be possible. `Enum.GetValues<TEnum>()` is AOT-safe, but requires each enum type to be statically known, which a generator could provide.

A generated mapping statically contains every supported value, so entries cannot disappear because enum fields or attribute metadata were trimmed. The tradeoff is that the generated mapping intentionally keeps its strings and entries even when individual enum values are unused.

### 4. Generate per-enum registration or lookup methods

Generate one mapping method/table per known enum and an internal dispatcher from `Enum`. This isolates generated code by enum type, can make the output easier to inspect, and could allow future strongly typed internal call sites. It still preserves the existing public `Enum` API.

### 5. Cache per-type metadata

Cache a mapping per runtime enum `Type` instead of per individual value. This amortizes reflection across all subsequent values of a type, but runtime initialization must avoid `Enum.GetValues(Type)` and must correctly preserve public fields and their attributes under trimming. It is more complex than lazy per-value caching and remains reflection-based.

## AOT considerations

- Do not use `Enum.GetValues(Type)` to initialize a runtime cache; it has `RequiresDynamicCode`.
- `Enum.GetValues<TEnum>()` is suitable only when generated/static code knows `TEnum`.
- `Enum.GetValuesAsUnderlyingType(Type)` avoids dynamic enum-array creation but returns underlying numeric values and does not solve attribute lookup.
- Suppressing trimming warnings does not preserve metadata by itself.
- A generated dictionary or switch avoids these concerns completely because runtime attribute metadata is not needed.

## Suggested direction

Prototype a source-generated mapping while retaining the current public `ToAttributeValue(Enum?)` API. Compare generated switch and generated dictionary output size, startup cost, and lookup throughput across the complete Fluent UI enum set. If source generation is too much machinery, a lazy `ConcurrentDictionary<Enum, string>` is a smaller interim improvement.

JamesNK:
# Pull Request

## 📖 Description

Avoid repeated reflection and boxing in known enum-to-attribute conversions by generating concrete, internal `GetDescription` and `ToAttributeValue` overloads backed by switch expressions.

Generation is explicitly opt-in through `[EnumAttributeValues(typeof(...))]` registrations on partial context classes, similar to a serialization context. Core registers 81 enum types and Charts registers five, including Core's `Orientation`. There is no automatic enum or assembly scanning, generic dispatcher, or runtime cache.

Existing typed calls select the generated overloads, including nullable, `isNull`, and `returnEmptyAsNull` variants. Charts imports the generated namespace in C# and Razor. The public `Enum`-based APIs remain unchanged for dynamic and unregistered callers.

Add diagnostic `FLUENTGEN001` to report concrete `ToAttributeValue` calls that still select the reflection API, including calls in Razor-generated code. The generator is referenced only at build time. Documentation explains registration and overload selection.

### 🎫 Issues

Fixes #5260.

Related: #5259 addresses broad enum reflection separately; this PR does not include that change.

## 👩‍💻 Reviewer Notes

- Review the explicit Core/Charts registrations and concrete overload selection rather than expecting a rename at each call site.
- Generated mappings preserve description literals, invariant lowercase fallback, empty strings, undefined values, nullable values, and sentinel handling. Aliases use the AOT-compatible generic `Enum.GetName` to preserve the runtime's name choice.
- Tests compare every registered enum value against the existing reflection implementation and exercise explicit opt-in, imported types, duplicate registrations, constants, string escaping, and the registration diagnostic.
- Generated methods are internal; this does not replace the public reflection API or change `GetDisplay`/`IsObsolete`.

## 📑 Test Plan

- `dotnet test tests/Core/Components.Tests.csproj --configuration Debug`: 4,340 passed; one existing Release-only test skipped.
- `dotnet test tests/Charts/Components.Charts.Tests.csproj --configuration Debug`: 402 passed.
- Debug Core and Charts project builds passed with the registration analyzer enabled.
- Disposable `net10.0`/`win-x64` Native AOT consumer published without warnings. It verified actual `Icon.WithColor` behavior, concrete generated overloads, null/sentinel/undefined values, and a registered enum field used only through its numeric value. One million lookup iterations allocated zero bytes after warmup.
- `git diff --cached --check` passed.
- Full Debug solution build was attempted but blocked during restore by unrelated `NU1603` warnings-as-errors in `FluentUI.Demo.Client` and `FluentUI.Samples.WasmStandalone`: `Microsoft.NET.Sdk.WebAssembly.Pack` version `11.0.0-preview.4.26203.108` was unavailable and `11.0.0-preview.4.26230.115` was resolved instead. The full solution build therefore remains unverified.

### General

- [x] I have read the [CONTRIBUTING](https://github.com/microsoft/fluentui-blazor/blob/master/docs/contributing.md) documentation and followed the standards for this project.

## ⏭ Next Steps

None required for this change.


src/Core.Generators/EnumAttributeValueAnalyzer.cs
```diff
@@ -0,0 +1,52 @@
...
+        context.RegisterOperationAction(static context =>
+        {
+            var invocation = (IInvocationOperation)context.Operation;
+            if (invocation.TargetMethod.Name != "ToAttributeValue" ||
+                invocation.TargetMethod.ContainingType.ToDisplayString() != "Microsoft.FluentUI.AspNetCore.Components.Extensions.EnumExtensions")
+            {
+                return;
+            }
+
+            var receiver = invocation.Arguments[0].Value;
```
JamesNK: [P2] Select the receiver by parameter ordinal, not argument position

Roslyn keeps these arguments in source order, so `Arguments[0]` need not be the `value` parameter. I reproduced both directions with the current analyzer:

```csharp
// Color value: compiles, but the required FLUENTGEN001 is missed.
EnumExtensions.ToAttributeValue(returnEmptyAsNull: true, value: value);

// System.Enum value: compiles, but incorrectly reports FLUENTGEN001 for Color.
EnumExtensions.ToAttributeValue(isNull: Color.Default, value: value);
```

The first call retains boxing/reflection unnoticed. The second rejects an intentionally dynamic receiver in a warnings-as-errors build, even though registering Color cannot change that receiver. Find the argument whose `Parameter.Ordinal == 0`, return if absent, and add tests for reordered named arguments with both concrete and dynamic receivers.

src/Core.Generators/EnumAttributeValueGenerator.cs
```diff
@@ -0,0 +1,113 @@
...
+            var suffix = field.ConstantValue is ulong ? "UL" : "L";
+            source.Append("            (").Append(typeName).Append(")(").Append(constant).Append(suffix).Append(") => ");
+            if (group.Count() == 1)
+            {
+                source.Append(SymbolDisplay.FormatLiteral(GetDescription(field), true)).AppendLine(",");
+            }
+            else
+            {
+                source.AppendLine("global::System.Enum.GetName(value) switch");
+                source.AppendLine("            {");
```
JamesNK: [P2] Preserve ToString name selection for aliased Flags values

`Enum.GetName(value)` and the existing `value.ToString()` lookup can choose different names for a `[Flags]` enum with aliases. A runtime probe using the current generator reproduced this:

```csharp
[Flags]
public enum Selected
{
    [Description("first-token")] First = 1,
    [Description("alias-token")] Alias = 1,
    Second = 2
}
```

For `(Selected)1`, generated `GetDescription` returns `first-token`, while `EnumExtensions.GetDescription` returns `alias-token`. Removing `[Flags]` makes the two agree, which explains why the existing non-flags compilation-only alias test misses this.

No mismatch was found among the enums currently registered in Core, so this is a gap in the generator's advertised alias compatibility rather than a current component regression. Preserve the legacy flags name-selection behavior in this branch and add an execution/parity test, not just a compilation test.

src/Core.Generators/EnumAttributeValueGenerator.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        var contexts = context.SyntaxProvider.ForAttributeWithMetadataName(
+            "Microsoft.FluentUI.AspNetCore.Components.Generators.EnumAttributeValuesAttribute",
+            static (node, _) => node is ClassDeclarationSyntax,
+            static (syntax, _) => syntax);
+
+        context.RegisterSourceOutput(contexts, static (production, input) =>
+        {
+            var contextType = (INamedTypeSymbol)input.TargetSymbol;
+            var types = input.Attributes.Select(attribute => attribute.ConstructorArguments[0].Value)
```
JamesNK: [P3] Skip malformed registrations instead of aborting generation

`ForAttributeWithMetadataName` also passes attribute applications with invalid constructor arguments to this transform. With a valid `[EnumAttributeValues(typeof(Selected))]` followed by an unfinished `[EnumAttributeValues]`, this index throws `IndexOutOfRangeException`. The current driver reports `CS8785`, and the valid context has zero generated `GetDescription` methods instead of one.

This is easy to encounter while editing the shared registration list and turns one local syntax error into missing generated methods across the project. Check the constructor-argument shape before reading its value, skip malformed registrations so the compiler reports their own errors, and add a driver test that keeps valid registrations working alongside an invalid one.

JamesNK (commented):
Reviewed a7a248d11a4f5b7f73f8b00d34c0616c5b26f3af. Three reproduced findings are inline:

- Fix receiver identification in FLUENTGEN001: reordered named arguments both hide concrete reflection calls and incorrectly reject dynamic Enum calls.
- Correct the aliased Flags lookup: Enum.GetName and the legacy ToString path can select different aliases and therefore different descriptions. This affects the generator's promised compatibility; no mismatch was found in the currently registered Core enums.
- Guard malformed registration attributes so an unfinished entry does not abort otherwise valid generation. This is an IDE robustness improvement.

I would address the first two before merging. The registration guard is lower priority.

The normal production path checks out: fresh Core and Charts Release builds passed; Icon.WithColor calls the generated overload, allocated zero bytes over 200,000 calls after warmup, and worked in a fresh net10.0/win-x64 Native AOT publish. All 81 registered Core enums matched the reflection implementation for declared values, with extra undefined/Flags values checked. The inspected Core package did not contain the generator or acquire a generator runtime dependency.

Separate Roslyn probes reproduced the named-argument and malformed-registration findings, and a generated assembly reproduced the Flags alias mismatch. Those probes also disproved the panel's proposed long.MinValue compilation failure and incomplete-invocation analyzer crash, so those are not findings.

Remaining verification gaps: minimum build-SDK compatibility for the Roslyn 5.6 dependency was not exercised; the full solution build remains blocked by the previously documented preview WebAssembly.Pack restore mismatch. The IDE incremental-generation concern raised by the panel was not measured and is not treated as a confirmed performance regression.

Full-tier lanes: architecture and .NET/Roslyn domain review used GPT-5.6 Sol; edge/regression and real-path proof used Claude Opus 5. All four completed independently. The agent tool does not expose reasoning-effort configuration, so effort levels could not be selected or verified.

No PR implementation files were changed during review.

src/Core.Generators/EnumAttributeValueGenerator.cs
JamesNK: Addressed in 79aa809be with an explicitly documented limitation rather than runtime alias lookup. Generated overloads use the first declared member for each numeric value and never call ToString or Enum.GetName. Conflicting alias descriptions can differ from the reflection implementation; this is documented beside the selection and in the generator README. Tests cover that deterministic rule for flags and non-flags enums, plus runtime parity for all four flags overloads where alias descriptions agree, including named/unnamed combinations, nullable values, and sentinel handling.

src/Core.Generators/EnumAttributeValueAnalyzer.cs
JamesNK: Fixed in 79aa809be. The analyzer now selects the argument with Parameter.Ordinal == 0 and returns if it is absent. Added regression cases for reordered named arguments with concrete, nullable, and dynamic Enum receivers. The full Core suite passed (4,372 tests, one existing skip).

src/Core.Generators/EnumAttributeValueGenerator.cs
JamesNK: Fixed in 79aa809be. Generation now checks the constructor-argument count and enum symbol before indexing, skipping malformed registrations while retaining valid ones. Added driver tests for an unfinished [EnumAttributeValues] and an unresolved typeof(Missing); only the expected compiler error remains, with no generator exception and valid overloads preserved. The full Core suite passed.

src/Core.Generators/EnumAttributeValueGenerator.cs
```diff
@@ -0,0 +1,113 @@
...
+    private static string GetDescription(IFieldSymbol field)
+    {
+        var attribute = field.GetAttributes().FirstOrDefault(attribute => attribute.AttributeClass?.ToDisplayString() == "System.ComponentModel.DescriptionAttribute");
+        if (attribute is { ConstructorArguments.Length: 0 })
+        {
+            return string.Empty;
+        }
+
+        return attribute?.ConstructorArguments.FirstOrDefault().Value as string ?? field.Name.ToLowerInvariant();
+    }
```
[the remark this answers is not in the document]
JamesNK: This is a false positive. The parameterless DescriptionAttribute constructor delegates to string.Empty, but DescriptionAttribute(string description) stores its argument unchanged. Its Description property therefore returns null for [Description(null)], not an empty string.

Source: https://github.com/dotnet/runtime/blob/v8.0.0/src/libraries/System.ComponentModel.Primitives/src/System/ComponentModel/DescriptionAttribute.cs

EnumExtensions.GetDescription uses attribute?.Description ?? lowercaseName, so [Description(null)] Value correctly produces "value", matching the generator. [Description] and [Description("")] produce "" instead. I verified this against the built Core library and locally extended the existing description test to execute generated and handwritten conversions, including returnEmptyAsNull. All five cases passed. That test strengthening is local and not yet committed; no production change is needed. Resolving as a false positive.

tests/Core/Extensions/EnumAttributeValueGeneratorTests.cs
```diff
@@ -0,0 +1,132 @@
...
+        Assert.Single(output.SyntaxTrees.Skip(1));
+        Assert.DoesNotContain(output.SyntaxTrees, tree => tree.GetText(TestContext.Current.CancellationToken).ToString().Contains("ToAttributeValue"));
+    }
+
+    [Theory]
+    [InlineData("[System.ComponentModel.Description] Value", "")]
+    [InlineData("[System.ComponentModel.Description(null)] Value", "value")]
+    [InlineData("[System.ComponentModel.Description(\"\")] Value", "")]
+    [InlineData("[System.ComponentModel.Description(\"quoted\\\"\\nvalue\")] Value", "quoted\"\nvalue")]
+    [InlineData("NoAttribute", "noattribute")]
```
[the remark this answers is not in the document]
JamesNK: The "value" expectation is correct and is being retained. [Description(null)] preserves null in DescriptionAttribute.Description, so the handwritten EnumExtensions method falls back to the lowercase field name. Only the parameterless constructor and an explicit empty string yield "".

Verified against the .NET implementation: https://github.com/dotnet/runtime/blob/v8.0.0/src/libraries/System.ComponentModel.Primitives/src/System/ComponentModel/DescriptionAttribute.cs

Locally extended this theory to emit and load the generated assembly, execute both generated and handwritten ToAttributeValue, and compare normal and returnEmptyAsNull results for null, empty, parameterless, escaped, and absent descriptions. All five cases passed. The test-only improvement is not yet committed, but the existing implementation and expectation already agree. Resolving this duplicate report as a false positive.

src/Core.Generators/EnumAttributeValueAnalyzer.cs
```diff
@@ -0,0 +1,58 @@
...
+
+namespace Microsoft.FluentUI.AspNetCore.Components.Generators;
+
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public sealed class EnumAttributeValueAnalyzer : DiagnosticAnalyzer
+{
+    private static readonly DiagnosticDescriptor MissingRegistration = new(
+        "FLUENTGEN001",
+        "Register enum attribute values for generation",
+        "Register '{0}' with EnumAttributeValuesAttribute and import its generated extensions to avoid reflection and boxing",
```
dvoituron: Can you add the name of the class containing these attributes? Easier to find where is it.

```csharp
with [EnumAttributeValues] attribute (example, in GeneratedEnumExtensions.cs) and import its generated extensions to avoid reflection and boxing
```

src/Core.Generators/AnalyzerReleases.Unshipped.md
```diff
@@ -0,0 +1,7 @@
+; Unshipped analyzer release
+
+### New Rules
+
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+FLUENTGEN001 | Performance | Warning | Register concrete enum attribute conversions for generation.
```
dvoituron: We've already used an “acronym” for FluentUI-Blazor: `fuib` (to distinguish it from FluentUI WebComponents or React).
Can we use the code `FUIBGEN001` (including in EnumAttributeValueAnalyzer.cs, README.md and EnumAttributeValueGeneratorTests.cs files)?

.github/workflows/build-core-lib.yml
```diff
@@ -166,7 +166,7 @@ jobs:
          title: 'Unit Tests Code Coverage'
          assemblyfilters: '-Microsoft.FluentUI.AspNetCore.Components.Tests.Tools;-McpServer.Tests'
          classfilters: '-Microsoft.FluentUI.AspNetCore.Components.DesignTokens.*'
-         filefilters: '-*RegexGenerator.g.cs'
+         filefilters: '-*RegexGenerator.g.cs;-*GeneratedEnumExtensions.g.cs'
```
dvoituron, line 169: Can you update the script `_StartCodeCoverage.ps1` with the same rule?

dvoituron (changes requested):

dvoituron (approved):

Issue #5151, opened 2026-08-15
fix: [DataGrid] Virtualized grid gets permanently stuck on "No data to show" once TotalItemCount is ever 0 (rc.5 regression)


christian-sig:
# 🐛 Bug Report

A `FluentDataGrid` with `Virtualize="true"` + `ItemsProvider` gets **permanently stuck showing "No data to show."** if the provider ever reports `TotalItemCount = 0` — even after data becomes available and `RefreshDataAsync()` is called.

- **Broken in:** `5.0.0-rc.5-26219.1` (also reproducible reading current `dev-v5` HEAD sources)
- **Works in:** `5.0.0-rc.4-26180.1` (same app code)
- Regression introduced by #4972 ("Improve Virtualize handling and data loading").

(I searched open and closed issues — no existing report for this.)

## 💻 Repro or Code Sample

1. `dotnet new install Microsoft.FluentUI.AspNetCore.Templates` and `dotnet new fluentblazor -n Rc5Repro` (interactive server), then upgrade the package:
   `dotnet add Rc5Repro package Microsoft.FluentUI.AspNetCore.Components --version 5.0.0-rc.5-26219.1`
2. Add the page below as `Components/Pages/Repro.razor` (no other changes needed).
3. Run, navigate to `/repro`, click **Load data**.

```razor
@page "/repro"
@rendermode InteractiveServer

<FluentButton OnClick="@LoadAsync">Load data</FluentButton>

<FluentDataGrid @ref="_grid" TGridItem="int"
                ItemsProvider="@Provider"
                Virtualize="true" ItemSize="32"
                Style="height: 300px;">
    <TemplateColumn TGridItem="int" Title="Value">@context</TemplateColumn>
</FluentDataGrid>

@code {
    private FluentDataGrid<int>? _grid;
    private List<int> _items = [];   // empty at first — data not loaded yet

    private ValueTask<GridItemsProviderResult<int>> Provider(GridItemsProviderRequest<int> req)
    {
        var page = _items.Skip(req.StartIndex).Take(req.Count ?? _items.Count).ToList();
        return ValueTask.FromResult(GridItemsProviderResult.From(page, _items.Count));
    }

    private async Task LoadAsync()
    {
        _items = Enumerable.Range(1, 1000).ToList();
        await _grid!.RefreshDataAsync();
    }
}
```

[...]

PR #5269, opened 2026-09-15
perf: optimize DataGrid debounce and cancellable delays

JamesNK:
# Pull Request

## Description

Improve virtualized DataGrid responsiveness and avoid exception overhead when cancelling routine delays.

- Keep the cancellable 20 ms debounce for rapid viewport requests, but skip it on the first request and after at least 100 ms since the previous request arrived.
- Skip one debounce after an explicit virtualized refresh, including when the virtualizer is remounted after an empty result, and after an accepted provider result changes the total item count.
- Track the last accepted virtualized total separately from the shared grid context so count changes are detected for ItemsProvider, IQueryable, and async query executor sources. Cancelled results do not update the accepted total.
- Use ConfigureAwaitOptions.SuppressThrowing for cancellable delays in DataGrid, the Debounce utility, and message-bar auto-dismiss. Preserve the captured context where needed and retain cancellation checks before subsequent work.
- Add deterministic debounce/count/cancellation regression tests and replace DataGrid test reflection with compile-time-checked internal access or assertions on rendered behavior.

No public API changes. The broader rendering, selection, and query-caching optimizations are not part of this PR.

### Issues

Related background: #4676 (initial-load performance) and #5151 (virtualizer recovery after an empty result). This PR does not claim to close either issue.

## Reviewer Notes

- The burst window uses monotonic request-arrival timestamps, including cancelled requests, rather than provider completion times.
- Review one-shot bypass behavior for explicit refreshes and accepted total-count changes, including IQueryable and async query executor sources.
- Suppression applies only to the known cancellable delays, not provider/database operations or arbitrary callbacks.
- Suggested browser smoke test: scroll rapidly, pause for more than 100 ms, resume scrolling, then explicitly refresh after changing the total count and after transitioning from empty to non-empty data. Browser smoke testing was not performed for this PR.
- No quantitative performance benchmark is included.

## Test Plan

- `dotnet build Microsoft.FluentUI-v5.slnx --configuration Debug`: passed.
- `dotnet test tests/Core/Components.Tests.csproj --configuration Debug`: 4,391 passed, 1 skipped, 0 failed. The skipped MaxItemCount test requires a Release build with Core targeting .NET 9.
- Regression coverage includes the 99/100 ms boundary, repeated requests extending a burst, count changes across all three data-source paths, cancellation before provider invocation, cancelled results preserving accepted counts, and empty-state remount refreshes.
- Rendering tests verify that successful loads omit error content and that refreshing after a handled loading error restores rows, without reflection.

### General

- [ ] I have updated the project documentation to reflect my changes. Not required for these internal performance changes.
- [x] I have read the [CONTRIBUTING](https://github.com/microsoft/fluentui-blazor/blob/dev-v5/docs/contributing.md) documentation and followed the project standards.

### Component-specific

- [x] I have modified an existing component.
- [x] I have validated the unit tests for an existing component.

## Next Steps

Evaluate broader DataGrid performance improvements in a separate PR, with measurements: cell metadata caching, selection indexing, reduced column-interaction setup/layout work, count caching with explicit freshness semantics, and avoiding unused row event handlers.


src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -1518,27 +1529,30 @@ private async Task RefreshDataCoreAsync()
...
     [SuppressMessage("Design", "MA0051:Method is too long", Justification = "Not going to do artificial optimization because of some random arbitrary determined line count number")]
-    private async ValueTask<ItemsProviderResult<(int, TGridItem)>> ProvideVirtualizedItemsAsync(ItemsProviderRequest request)
+    internal async ValueTask<ItemsProviderResult<(int, TGridItem)>> ProvideVirtualizedItemsAsync(ItemsProviderRequest request)
     {
         _lastRefreshedPaginationState = Pagination;
-        // Debounce the requests (except on first call). This eliminates a lot of redundant queries at the cost of slight lag after interactions.
-        if (_isFirstVirtualizeProviderCall)
+        // Debounce rapid requests from scrolling, but do not delay the first request after an idle period,
+        // an explicit refresh, or a provider result that changed the total item count.
+        if (ShouldDebounceVirtualizeProviderRequest(Stopwatch.GetTimestamp()))
```
JamesNK: A request whose token is already cancelled consumes `_skipNextVirtualizeProviderDelay` here, then returns at the guard below without delaying or invoking the provider. That means a refresh/count-change bypass can be spent on work that never starts, and the next usable request in the same burst still pays the 20 ms delay. Could we check cancellation before consuming the one-shot flag, while still recording cancelled arrivals if that timestamp behavior is intentional?

tests/Core/Components/DataGrid/FluentDataGridTests.razor
```diff
@@ -794,6 +803,125 @@
...
+    [InlineData(99, true)]
+    [InlineData(100, false)]
+    [InlineData(101, false)]
+    public void ShouldDebounceVirtualizeProviderRequest_ElapsedInterval_UsesBurstBoundary(int elapsedMilliseconds, bool expectedDelay)
+    {
+        var grid = new FluentDataGrid<Customer>(new LibraryConfiguration());
+        var timestamp = System.Diagnostics.Stopwatch.Frequency;
+
+        Assert.False(grid.ShouldDebounceVirtualizeProviderRequest(timestamp));
+        Assert.Equal(expectedDelay, grid.ShouldDebounceVirtualizeProviderRequest(timestamp + elapsedMilliseconds * System.Diagnostics.Stopwatch.Frequency / 1000));
```
JamesNK: This rounds the requested duration down to whole stopwatch ticks. On a platform where `Stopwatch.Frequency` is not divisible by 1,000, the 100 ms case becomes slightly less than 100 ms and can incorrectly debounce. Please construct the offset with ceiling, for example `(long)Math.Ceiling(elapsedMilliseconds * Stopwatch.Frequency / 1000d)`, so the boundary case is portable.

JamesNK (commented):
Two issues to address: an already-cancelled virtualizer request can consume the one-shot delay bypass without doing work, and the exact 100 ms test boundary is platform-dependent because its stopwatch tick conversion truncates.

Focused validation passed 22 targeted DataGrid/Debounce tests plus 3 MessageBar lifecycle tests. The supplied demo page only renders a non-virtualized one-row grid, so it could not prove the changed rapid-scroll, count-change, or empty-to-non-empty virtualized paths in a browser. The existing bot warning about very long MessageBar lifetimes is not introduced by this change: `Task.Delay` validates that argument synchronously before `ConfigureAwait` can suppress anything.

tests/Core/Components/DataGrid/FluentDataGridTests.razor
JamesNK: Fixed in `a41ec57a3`. The test now uses `Math.Ceiling` when converting milliseconds to stopwatch ticks, so the 100 ms case cannot be rounded below the boundary. The four boundary cases and the full DataGrid test file pass.

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
JamesNK: Fixed in `1b2425102`. Already-cancelled requests still update the arrival timestamp, but now return before consuming `_skipNextVirtualizeProviderDelay`. The regression test verifies the provider is not called and the bypass remains available for the next usable request.

dvoituron (approved):

vnbaaij (approved):

Issue #5065, opened 2026-08-01
FluentDataGrid: EnableColumnReordering runs for every grid and throws on a detached element, killing the Blazor Server circuit


MustafaLattouf:
## Summary

On `5.0.0-rc.4-26180.1`, `FluentDataGrid` calls
`Microsoft.FluentUI.Blazor.DataGrid.EnableColumnReordering` for **every** grid, even when
`ReorderableColumns` is left at its default `false`. That JS function dereferences the grid element
without a null check, so when the captured `ElementReference` no longer resolves to a node in the
DOM it throws:

```
Microsoft.JSInterop.JSException: Cannot read properties of null (reading 'id')
TypeError: Cannot read properties of null (reading 'id')
    at Object.EnableColumnReordering (.../Components/DataGrid/FluentDataGrid.razor.js:371:44)
    at b.processJSCall (blazor.web.js:1:4448)
   at Microsoft.JSInterop.JSRuntime.InvokeAsync[TValue](Int64 targetInstanceId, String identifier, JSCallType callType, Object[] args)
   at Microsoft.FluentUI.AspNetCore.Components.FluentDataGrid`1.OnAfterRenderAsync(Boolean firstRender)
   at Microsoft.AspNetCore.Components.RenderTree.Renderer.GetErrorHandledTask(Task taskToHandle, ComponentState owningComponentState)
```

On Blazor Server this escapes `OnAfterRenderAsync` as an unhandled exception and **terminates the
circuit** — the page keeps its rendered pixels and silently stops responding to every click, which is
a hard failure mode for a bug in a feature the app never opted into.

### 1. The interop is not gated on `ReorderableColumns`

In `FluentDataGrid.razor.cs`, `FinishCollectingColumns()` sets the resizing flag conditionally but the
reordering flag unconditionally:

```csharp
if (ResizableColumns) { _checkColumnResizing = true; }   // gated
_checkColumnReordering = true;                            // not gated
```

so `OnAfterRenderAsync` runs the reordering path for grids that will never reorder:

```csharp
if (_checkColumnReordering && _gridReference is not null && JSModule.Imported && _selfReference is not null)
{
    _checkColumnReordering = false;
    await JSModule.ObjectReference.InvokeVoidAsync(
        "Microsoft.FluentUI.Blazor.DataGrid.EnableColumnReordering", _gridReference, _selfReference);
}
```

Expected: `_checkColumnReordering` is set only when `ReorderableColumns` is `true`, matching
[...]

PR #5270, opened 2026-09-15
[dev-v5] Perf: avoid redundant DataGrid column interaction initialization

JamesNK:
# Pull Request

## Description

Avoid rebuilding DataGrid column interactions on ordinary renders and data refreshes.

Previously, FinishCollectingColumns scheduled resize and reorder initialization on every grid render. With those features enabled, each refresh aborted listeners, measured headers, removed and recreated resize handles, and registered new listeners.

This change:

- Tracks column identity/order, keys, pinning, widths, and relevant grid options so unchanged renders skip interaction setup.
- Retains setup for structural/configuration changes and performs cleanup when interactions are disabled.
- Uses a coalesced ResizeObserver callback to update existing resize-handle geometry when table/header dimensions change, independently of listener initialization.
- Removes duplicate resize initialization from Initialize and refreshes cached column widths when a pointer resize starts.
- Retains the JavaScript controller returned by Initialize and stops/disposes it through the component disposal hook. Browser testing found that otherwise the new observer retained a detached table and headers after unmounting.

No public .NET API changes. Data fetching and row rendering are unchanged. This is independent of the debounce changes in #5269.

### Issues

Related background: #5065 (unnecessary column-reordering initialization). No issue is closed by this PR.

## Reviewer Notes

- Review invalidation when columns are added, removed, replaced, reordered, pinned, or resized through configuration, and when header visibility or interaction options change.
- Review listener/observer teardown on disable, reconfiguration, and component disposal, including disposal during a pointer resize.
- A Playwright run with 30 real Blazor refreshes at 100 ms intervals retained the same resize handles and registered no additional interaction listeners. Resize and reorder drags held across refreshes completed successfully.
- No millisecond speedup is claimed; browser checks measured avoided setup work and verified behavior.

## Test Plan

- `dotnet build Microsoft.FluentUI-v5.slnx --configuration Debug`: passed.
- `dotnet test tests/Core/Components.Tests.csproj --configuration Debug`: 4,396 passed, 1 skipped, 0 failed. The existing MaxItemCount test is Release-only with Core targeting .NET 9.
- Focused DataGrid TypeScript strict typecheck: passed.
- Added 15 bUnit regression cases covering unchanged renders/data refreshes, configuration changes, dynamic columns, idempotent ordering, disable/re-enable, and controller disposal exactly once. Tests do not use reflection.

Playwright Chromium validation used the existing demos and a temporary Blazor fixture, which was removed after testing:

- 10 Hz refreshes, including resize and reorder drags spanning refreshes.
- Grid and Table layouts in LTR and RTL, including width reset.
- Growing and shrinking row counts, header-only resize handles, and feature toggles.
- Header hide/show, column removal/re-addition/replacement, programmatic order changes, and drag-and-drop.
- Virtualized scrolling to the final rows and refreshing at the scrolled position.
- Repeated mount/unmount cycles, observer cleanup, and unmount during an active resize.

No further regressions or page errors were observed in these Chromium checks after fixing disposal. Firefox and WebKit were not tested. The temporary browser fixture is not part of the PR; the bUnit tests are retained.

### General

- [ ] I have updated the project documentation to reflect my changes. Not required for this internal optimization.
- [x] I have read the [CONTRIBUTING](https://github.com/microsoft/fluentui-blazor/blob/dev-v5/docs/contributing.md) documentation and followed the project standards.

### Component-specific

- [x] I have modified an existing component.
- [x] I have validated the unit tests for an existing component.

## Next Steps

Other DataGrid performance opportunities, including cell metadata caching, selection indexing, and count caching, remain separate follow-up work.


src/Core/Components/DataGrid/FluentDataGrid.razor.ts
```diff
@@ -573,6 +588,9 @@ export namespace Microsoft.FluentUI.Blazor.DataGrid {
         }
 
         const isGrid = gridElement.getAttribute('display-mode') === 'grid';
+        columns.forEach(column => {
```
JamesNK: Refresh the cached widths for menu resizing as well as pointer resizing. `ResizeColumnDiscrete` and `ResizeColumnExact` still serialize `grid.columns[].size` for the untouched columns. Once ordinary renders stop reinitializing interactions, those sizes can be stale after a responsive layout change.

Chromium reproduction using the current module: initialize a `1fr 1fr` grid at 500px (`[250, 250]`), grow it to 600px (`[300, 300]`), then increase column 1 by 10px or set it to 310px. Both commands produce `[310, 250]`, shrinking column 2 by 50px. Calling `EnableColumnResizing` before the command, as an ordinary render did before this PR, gives the correct `[310, 300]`.

Please share this width-refresh logic with the discrete/exact paths and cover both commands after a container resize followed by an unchanged render. Refreshing only in `pointerdown` leaves the accessible resize controls with different behavior.

JamesNK (commented):
Reviewed 20a036a5f. Two issues should be addressed before merge:

- Discrete/exact resize commands can shrink untouched columns after a responsive layout change because only pointer resizing refreshes cached widths. The new inline comment includes a Chromium reproduction and a comparison with the previous reinitialization behavior.
- Both new disable functions need null-element guards. This is already covered by the existing review thread; I confirmed that both functions throw in Chromium and will add the reproduction there rather than create a duplicate finding.

Fresh browser proof also passed sorting without listener reinitialization, handle geometry updates, pointer resizing with pinned offsets, and drag-and-drop reordering. The browser-only two-column probe measured `[300, 300] -> [310, 250]` for both menu resize commands, versus `[310, 300]` when reinitialized first. Temporary browser elements were removed.

Review coverage: independent architecture, lifecycle/edge-case, domain, and browser-proof lanes, plus a focused completion pass over discrete/exact resizing and CSS. GPT-5.6 Sol handled architecture, browser proof, and the domain follow-up; Claude Opus 5 handled lifecycle and the initial domain pass. The agent tool does not expose reasoning-effort control.

The existing initialization/disposal race remains a follow-up concern: the new cleanup test covers completed initialization, not disposal while Initialize is pending. I have not classified the inherited leak as a new regression. Firefox/WebKit and delayed interop completion were not exercised in this review. No source changes were made.

src/Core/Components/DataGrid/FluentDataGrid.razor.ts
```diff
@@ -407,11 +405,21 @@ export namespace Microsoft.FluentUI.Blazor.DataGrid {
     }
   }
 
-  export function EnableColumnResizing(gridElement: HTMLElement, resizeColumnOnAllRows: boolean = true, signal?: AbortSignal) {
+  export function DisableColumnResizing(gridElement: HTMLElement) {
+    const grid = grids.find(grid => grid.id === gridElement.id);
```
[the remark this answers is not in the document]
JamesNK: Confirmed in Chromium against 20a036a5f: both `DisableColumnResizing(null)` and `DisableColumnReordering(null)` throw `TypeError: Cannot read properties of null (reading 'id')`. The enable functions already return for missing elements. Please apply that same guard to both new disable functions; the C# non-null ElementReference check cannot guarantee that the DOM node still exists when the call reaches the browser.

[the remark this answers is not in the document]
JamesNK: Fixed in c9b2b36d771841963dedad243ecff27ba9c3559f. Both disable functions now return early with `if (!gridElement)`, consistent with the enable/initialize paths. Verified null and undefined calls leave a live grid's handles and draggable headers intact in Chromium; the final simplified guards also passed direct falsy-input checks and strict TypeScript compilation.

src/Core/Components/DataGrid/FluentDataGrid.razor.ts
JamesNK: Fixed in c9b2b36d771841963dedad243ecff27ba9c3559f. Pointer, discrete, and exact resizing now share `refreshColumnWidths`, so menu commands measure current widths before modifying the selected column.

Verified both commands against the retained Blazor fixture after growing its container from 500px to 600px and calling RefreshDataAsync: `[300, 300]` now becomes `[310, 300]`, with the original handles retained. Minimum-width clamping, reset, and pointer resizing also passed in Chromium. The optional automated integration test class was removed as requested; the manual fixture remains in the commit.

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -78,6 +79,8 @@ public int GetHashCode((int, TGridItem) item)
     private bool _checkColumnHeaderUiPosition;
     private bool _checkColumnResizing;
     private bool _checkColumnReordering;
+    private (bool Resizing, bool Reordering, bool AllRows, DataGridGeneratedHeaderType? Header, DataGridDisplayMode DisplayMode, DataGridRowSize RowSize, bool MultiLine, string? Template)? _columnInteractionOptions;
```
dvoituron: This tuple stores eight related values and is kept as component state, so a dedicated readonly record struct makes the intent clearer and the code easier to maintai

```
private readonly record struct ColumnInteractionOptions(
    bool Resizing,
    bool Reordering,
    bool AllRows,
    DataGridGeneratedHeaderType? Header,
    DataGridDisplayMode DisplayMode,
    DataGridRowSize RowSize,
    bool MultiLine,
    string? Template);
```

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
```diff
@@ -726,6 +719,53 @@ protected override async Task OnAfterRenderAsync(bool firstRender)
...
+    /// <inheritdoc />
+    protected override async ValueTask DisposeAsync(IJSObjectReference jsModule)
+    {
+        if (_gridController is not null)
+        {
+            var controller = _gridController;
+            _gridController = null;
+            try
+            {
+                await controller.InvokeVoidAsync("stop");
```
dvoituron, line 742: Can you add a comment to explain that `stop` is a method of the object returned by `Initialize`. `Initialise` creates an `AbortController`, registers events with its signal, and then returns an object containing a `stop` function.

dvoituron (changes requested):

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
JamesNK: Addressed in 2d481cdcc29eeaa5236d36e958bec8588923515a. Added a comment explaining that `Initialize` returns the controller and its `stop` method aborts the registered event listeners and observers. The disposal test continues to verify that `stop` is invoked exactly once; all 15 focused column-interaction tests pass.

src/Core/Components/DataGrid/FluentDataGrid.razor.cs
JamesNK: Addressed in 2d481cdcc29eeaa5236d36e958bec8588923515a. Replaced the eight-value tuple with a private readonly `ColumnInteractionOptions` record struct and construct it explicitly in `UpdateColumnInteractionState`; value equality preserves the existing invalidation behavior. All 15 `FluentDataGrid_ColumnInteractions` tests pass on the latest `dev-v5`.

JamesNK:
Done.

Please test this works with a bunch of data grids. I smoke tested against one but perhaps there are bugs hidden in different scenarios.

JamesNK:
Wait a second on merging. I want to ensure dispose doesn't leak exceptions on disconnect.

JamesNK:
It's fine as is. The base class handles catching errors.

dvoituron (approved):

vnbaaij (approved):

PR #5301, opened 2026-09-18
fix: marshal dialog provider Add/Update onto sync context

### LINKED ISSUE #5300: fix: FluentDialogProvider ShowDialogAsync races: Task.Run mutates References off the sync context
## What happens

On Blazor Server, opening a dialog immediately after another dialog closes (close-then-open sequence) intermittently throws `ArgumentOutOfRangeException` from the provider, or a `NullReferenceException` on the next render / dead circuit (frozen UI).

## Root cause

`FluentDialogProvider.ShowDialogAsync` wraps `References.Add` in `Task.Run` (`src/Core/Components/Dialog/FluentDialogProvider.razor.cs`). `References` is a plain unsynchronized `Collection<IDialogReference>`, so the pool-thread `Add` races the UI-thread render enumeration (`RenderDialogs` foreach) and the `Remove` on close. Present in every v4 tag checked (4.9.0, 4.11.8, 4.12.0, 4.14.4).

## Repro sketch

Loop on Blazor Server: open a dialog, close it via OK, immediately open the next one. Failure rate is a few percent per cycle; rapid OK-then-open makes it frequent.

## Suggested fix

Do the `Add` inline on the caller's context, exactly like the sync `ShowDialog` path already does (it has no `Task.Run`). There is no blocking work inside the delegate that justifies leaving the sync context. Same for `UpdateDialogAsync`. I will open a PR with this change.

shanewas:
Fixes #5300.

`ShowDialogAsync` / `UpdateDialogAsync` wrapped the `References` mutation in `Task.Run`, racing the UI-thread render enumeration and the close-path `Remove` whenever a dialog opened right after another closed. On Blazor Server this surfaces as `ArgumentOutOfRangeException`, a render `NullReferenceException`, or a dead circuit, at a low single-digit percent per cycle.

This runs both bodies through `InvokeAsync` instead, matching the sync `ShowDialog` path already in this file, so the `Add` / `Update` always lands on the renderer's sync context. No behavior change besides removing the race window. `ComponentBase.InvokeAsync` has no value-returning overload, so results are captured via locals.

Verified: `Microsoft.FluentUI.AspNetCore.Components` builds with 0 errors (net8.0 / net9.0 / net10.0).


shanewas:

vnbaaij (changes requested):
Hi, Thanks for your contribution but 2 things:

- We are moving into maintenance mode for v4. Final release of v5 is around the corner. That doesn't mean we can't merge this in but might take a bit longer for this to show up in a package
- You must target the `archive-v4`branch for this change (since yesterday, was `dev` before)

shanewas:
Retargeted to archive-v4. Rebased onto 518038a (same 1-file hunk, commit re-signed). Local SDK predates the pinned Roslyn so the full build can't link here; verified zero new errors against pristine base. CI should give the full gate.

vnbaaij (approved):

